PR #168, opened 2022-03-28
refactor: increase nb of children, for delete, add zip public endpoints

### LINKED ISSUE #167: Increase allowed number of children in an item

pyphilia:
This PR updates dependencies, increase the limit for deletion and creation, register the public endpoint for export/import zip.

close #167


.devcontainer/localstack/init.sh
```diff
@@ -17,6 +17,19 @@ corsConfig='{
...
+            ],
+            "AllowedMethods": [
+                "HEAD",
+                "GET",
+            ],
+            "AllowedOrigins": [
+                "*"
+            ],
+            "ExposeHeaders": []
         }
```
spaenleh, line 33: What is the difference between this and the above ? Out of curiosity.

test/membership.test.ts
```diff
@@ -833,39 +833,9 @@ describe('Membership routes tests', () => {
     });
     it(`Cannot delete more than ${MAX_ITEM_MEMBERSHIPS_FOR_DELETE} memberships`, async () => {
       const item = getDummyItem();
-      const memberships = [
+      const memberships = Array.from({ length: MAX_ITEM_MEMBERSHIPS_FOR_DELETE + 1 }, () =>
```
spaenleh, line 836: Yay loops !

spaenleh (approved):
Looks good ! Just a comment to understand 😄

.devcontainer/localstack/init.sh
pyphilia: The rule above allow only origins `null` to manipulate the files (in our case the backend). 
The second rule allow any origin to get the files, important for apps to fetch files.
Note: you can only access files given a signed url, given by the backend.

spaenleh: OK ! 🔥

PR #174, opened 2022-04-04
166/AddPasswordLoginMechanism

alvrba:


db-schema.sql
```diff
@@ -74,5 +75,6 @@ EXECUTE PROCEDURE trigger_set_timestamp();
 
 
 -- Graasp member
-INSERT INTO "member" ("id", "name", "email")
-VALUES ('12345678-1234-1234-1234-123456789012', 'Graasp', 'graasp@graasp.org');
+INSERT INTO "member" ("id", "name", "email", "password")
+VALUES ('12345678-1234-1234-1234-123456789012', 'Graasp', 'graasp@graasp.org', '$2b$10$4ixKKHAC.oSuO6D6/g35KOw3OPOkkOlfWk55y/XuQzFRvaXejzdXe');
```
pyphilia, line 79: Not sure we want this user to have a password. I think this user is used for public endpoints, so we don't want it to be accessible.

src/plugins/auth/auth.ts
```diff
@@ -205,6 +216,48 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
       .catch((err) => log.warn(err, `mailer failed. link: ${link}`));
   }
 
+  async function verifyCredentials(
+    member: Member,
+    body: { email: string; password: string },
+    challenge?: string,
+  ) {
+    /* let verified is used to store the output of bcrypt.compare()
```
pyphilia, line 224: Maybe `the verified variable` instead of `let verified`

src/plugins/auth/auth.ts
```diff
@@ -205,6 +216,48 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+    member: Member,
+    body: { email: string; password: string },
+    challenge?: string,
+  ) {
+    /* let verified is used to store the output of bcrypt.compare() 
+    bcrypt.compare() allows to compare the provided password with a stored hash. 
+    It deduces the salt from the hash and is able to then hash the provided password correctly for comparison
+    if they match, verified is true 
+    if they do not match, verified is false
+    if the user does not has a stored hash, verified is null
```
pyphilia, line 229: `have`

src/plugins/auth/auth.ts
```diff
@@ -205,6 +216,48 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+    */
+    let verified;
+    if (member.password === null) {
+      verified = null;
+    } else {
+      verified = bcrypt
+        .compare(body.password, member.password)
+        .then(async (res: boolean) => {
+          if (res) {
```
pyphilia, line 238: For simplification, I would have 
```
if(!res) {
  return false
}
```
It will avoid to scope the following logic.

src/plugins/auth/auth.ts
```diff
@@ -262,6 +316,32 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
       },
     );
 
+    // loginpassword
+    fastify.post<{ Body: { email: string; password: string } }>(
+      '/loginpassword',
```
pyphilia, line 321: `login-password`

src/plugins/auth/auth.ts
```diff
@@ -262,6 +316,32 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
       },
     );
 
+    // loginpassword
```
pyphilia, line 319: `login with password`

src/plugins/auth/auth.ts
```diff
@@ -262,6 +316,32 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+            reply.status(StatusCodes.NOT_ACCEPTABLE).send(ReasonPhrases.NOT_ACCEPTABLE);
+          } else if (link) {
+            reply.status(StatusCodes.SEE_OTHER).send({ link });
+          } else {
+            reply.status(StatusCodes.UNAUTHORIZED).send(ReasonPhrases.UNAUTHORIZED);
+          }
+        } else {
+          const { email } = body;
+          log.warn(`Login attempt with non-existent email '${email}'`);
+          reply.status(StatusCodes.NOT_FOUND).send(ReasonPhrases.NOT_FOUND);
```
pyphilia, line 340: I prefer returning the result as soon as possible, so we avoid encapsulating ifs, especially if you mis errors and success cases. Could you refactor it and take advantage of `return reply...`?

src/plugins/auth/auth.ts
```diff
@@ -382,6 +463,30 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
         },
       );
 
+      fastify.post<{
+        Body: { email: string; challenge: string; password: string };
+      }>('/loginpassword', { schema: mPasswordLogin }, async ({ body, log }, reply) => {
```
pyphilia, line 468: `login-password`

src/plugins/auth/auth.ts
```diff
@@ -382,6 +463,30 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+            reply.status(StatusCodes.OK).send({ t: token });
+          } else {
+            reply.status(StatusCodes.UNAUTHORIZED).send(ReasonPhrases.UNAUTHORIZED);
+          }
+        } else {
+          const { email } = body;
+          log.warn(`Login attempt with non-existent email '${email}'`);
+          reply.status(StatusCodes.NOT_FOUND).send(ReasonPhrases.NOT_FOUND);
+        }
+      });
```
pyphilia, line 488: Same as above.

src/plugins/auth/auth.ts
```diff
@@ -205,6 +216,48 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+              expiresIn: `${LOGIN_TOKEN_EXPIRATION_IN_MINUTES}m`,
+            });
+            if (challenge) {
+              // token for graasp mobile app
+              return token;
+            } else {
+              // link for graasp web
+              const linkPath = '/auth';
+              const link = `${PROTOCOL}://${EMAIL_LINKS_HOST}${linkPath}?t=${token}`;
+              return link;
```
pyphilia, line 250: Maybe it makes sense to factor out the challenge check? I don't really like the fact this function either return a token, or a link. I think it can be dealt in the respective endpoint (that is, build the link if needed only).

src/plugins/auth/schemas.ts
```diff
@@ -54,6 +54,18 @@ const login = {
     additionalProperties: false,
   },
 };
+const passswordLogin = {
+  body: {
+    type: 'object',
+    required: ['email', 'password'],
+    properties: {
+      email: { type: 'string', format: 'email' },
+      //TODO: review
```
pyphilia, line 63: Still todo?

src/plugins/auth/auth.ts
```diff
@@ -230,6 +283,7 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
           const task = memberTaskManager.createCreateTask(GRAASP_ACTOR, {
             ...body,
             extra: { lang },
+            password: null,
```
pyphilia, line 286: Is it necessary to specify the password as null? Cannot it be optional?

src/plugins/auth/auth.ts
```diff
@@ -351,6 +431,7 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
               name,
               email,
               extra: { lang },
+              password: null,
```
pyphilia, line 434: Same as above.

test/auth.test.ts
```diff
@@ -31,7 +32,7 @@ describe('Auth routes tests', () => {
 
       expect(mockSendRegisterEmail).toHaveBeenCalled();
       expect(mockCreate).toHaveBeenCalledWith(
-        { email, name, extra: { lang: DEFAULT_LANG } },
+        { email, name, extra: { lang: DEFAULT_LANG }, password },
```
pyphilia, line 35: If the create had a password optional, you wouldn't have needed to add the password here.

test/auth.test.ts
```diff
@@ -55,7 +57,10 @@ describe('Auth routes tests', () => {
         expect.anything(),
         lang,
       );
-      expect(mockCreate).toHaveBeenCalledWith({ email, name, extra: { lang } }, expect.anything());
+      expect(mockCreate).toHaveBeenCalledWith(
+        { email, name, extra: { lang }, password },
```
pyphilia, line 61: Same as above.

src/plugins/auth/auth.ts
```diff
@@ -205,6 +216,48 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+            const token = await promisifiedJwtSign({ sub: member.id, challenge }, JWT_SECRET, {
+              expiresIn: `${LOGIN_TOKEN_EXPIRATION_IN_MINUTES}m`,
+            });
+            if (challenge) {
+              // token for graasp mobile app
+              return token;
+            } else {
+              // link for graasp web
+              const linkPath = '/auth';
+              const link = `${PROTOCOL}://${EMAIL_LINKS_HOST}${linkPath}?t=${token}`;
```
pyphilia, line 249: I think you discuss it with JC, but why isn't redirection possible?

test/auth.test.ts
```diff
@@ -170,6 +175,84 @@ describe('Auth routes tests', () => {
...
+        method: HTTP_METHODS.POST,
+        url: '/loginpassword',
+        payload: { email: member.email, password: clearPassword },
+      });
+      expect(response.statusCode).toEqual(StatusCodes.NOT_ACCEPTABLE);
+      expect(response.statusMessage).toEqual(ReasonPhrases.NOT_ACCEPTABLE);
+      app.close();
+    });
+
+    it('Sign In does send not found error for non-existing email', async () => {
```
pyphilia, line 224: remove `does`

test/auth.test.ts
```diff
@@ -170,6 +175,84 @@ describe('Auth routes tests', () => {
...
+      const email = 'some@email.com';
+      const password = '1234';
+      const app = await build();
+      const response = await app.inject({
+        method: HTTP_METHODS.POST,
+        url: '/loginpassword',
+        payload: { email, password },
+      });
+
+      //expect(mockSendLoginEmail).not.toHaveBeenCalled();
```
pyphilia, line 234: Remove?

test/auth.test.ts
```diff
@@ -368,6 +453,83 @@ describe('Auth routes tests', () => {
...
+          method: HTTP_METHODS.POST,
+          url: '/m/loginpassword',
+          payload: { email: member.email, challenge, password: clearWrongPassword },
+        });
+        expect(response.statusCode).toEqual(StatusCodes.UNAUTHORIZED);
+        expect(response.statusMessage).toEqual(ReasonPhrases.UNAUTHORIZED);
+        app.close();
+      });
+
+      it('Sign In does send not acceptable error when member does not have password', async () => {
```
pyphilia, line 487: Remove `does`

test/auth.test.ts
```diff
@@ -368,6 +453,83 @@ describe('Auth routes tests', () => {
...
+          method: HTTP_METHODS.POST,
+          url: '/m/loginpassword',
+          payload: { email: member.email, challenge, password: clearPassword },
+        });
+        expect(response.statusCode).toEqual(StatusCodes.NOT_ACCEPTABLE);
+        expect(response.statusMessage).toEqual(ReasonPhrases.NOT_ACCEPTABLE);
+        app.close();
+      });
+
+      it('Sign In does send not found error for non-existing email', async () => {
```
pyphilia, line 502: Remove `does`

pyphilia (changes requested):
Nice addition! I think it needs some refactor, we can also discuss this afternoon 💮

src/plugins/auth/auth.ts
victorggonzalez: I refactored this. The verifyCredentials function always returns a boolean. If the password is verified then it generates the token+link for the web app and only the token for the mobile app

src/plugins/auth/auth.ts
victorggonzalez: I didn't find a way to re-request to an endpoint inside the backend. Thought it was easier to return the link and then trigger the request to /auth from the frontend

src/plugins/auth/auth.ts
```diff
@@ -205,6 +216,25 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+    if they do not match, verified is false
+    */
+    const verified = bcrypt
+      .compare(body.password, member.password)
+      .then(async (res: boolean) => {
+        if (!res) {
+          return false;
+        }
+        return true;
+      })
```
pyphilia, line 233: You can simplify with `return res`

src/plugins/auth/schemas.ts
```diff
@@ -54,6 +54,17 @@ const login = {
     additionalProperties: false,
   },
 };
+const passswordLogin = {
```
pyphilia, line 57: `passwordLogin` -> three s?

pyphilia (changes requested):

db-schema.sql
```diff
@@ -76,3 +77,6 @@ EXECUTE PROCEDURE trigger_set_timestamp();
 -- Graasp member
 INSERT INTO "member" ("id", "name", "email")
 VALUES ('12345678-1234-1234-1234-123456789012', 'Graasp', 'graasp@graasp.org');
+INSERT INTO "member" ("id", "name", "email", "password")
+VALUES ('12345678-1234-1234-1234-123456789000', 'Graasp Admin', 'admin@graasp.org', '$2b$10$4ixKKHAC.oSuO6D6/g35KOw3OPOkkOlfWk55y/XuQzFRvaXejzdXe');
+
```
pyphilia, line 82: Remove!

src/plugins/auth/auth.ts
```diff
@@ -205,6 +216,22 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+  async function verifyCredentials(member: Member, body: { email: string; password: string }) {
+    /* the verified variable is used to store the output of bcrypt.compare() 
+    bcrypt.compare() allows to compare the provided password with a stored hash. 
+    It deduces the salt from the hash and is able to then hash the provided password correctly for comparison
+    if they match, verified is true 
+    if they do not match, verified is false
+    */
+    const verified = bcrypt
+      .compare(body.password, member.password)
+      .then(async (res: boolean) => {
```
pyphilia, line 228: You don't need async -> `.then(res => res)`

src/plugins/auth/auth.ts
```diff
@@ -262,6 +289,39 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+          return reply.status(StatusCodes.NOT_FOUND).send(ReasonPhrases.NOT_FOUND);
+        }
+        if (member.password === null) {
+          log.warn('Login attempt with non-existent password');
+          return reply.status(StatusCodes.NOT_ACCEPTABLE).send(ReasonPhrases.NOT_ACCEPTABLE);
+        }
+        const verified = await verifyCredentials(member, body);
+        if (!verified) {
+          reply.status(StatusCodes.UNAUTHORIZED).send(ReasonPhrases.UNAUTHORIZED);
+        } else {
```
pyphilia, line 313: So we don't factor out?

pyphilia (changes requested):
Here are my last comments, once it's solved you can merge!

db-schema.sql
victorggonzalez: done

src/plugins/auth/auth.ts
victorggonzalez: Thought this is what you meant in previous comments. Here we try to avoid encapsulating ifs and use the `return reply`
First check error cases (check member and password exists, as it is done in other functions above) and `return`.
Then verify credential and check success cases. 
We can try to refactor again if you consider it necessary.

pyphilia (approved):

pyphilia:
@alvrba @victorggonzalez can we merge this? ^^

PR #176, opened 2022-04-11
fix: add plugin options

louisewang1:
Add the options for validation plugin

### DIFF
--- a/src/services/items/service-api.ts
+++ b/src/services/items/service-api.ts
@@ -47,6 +47,7 @@ import {
   HIDDEN_TAG_ID,
   SAVE_ACTIONS,
   CLIENT_HOSTS,
+  IMAGE_CLASSIFIER_API,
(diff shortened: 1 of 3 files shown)


src/services/items/service-api.ts
```diff
@@ -209,7 +209,15 @@ const plugin: FastifyPluginAsync = async (fastify) => {
 
         fastify.register(graaspCategoryPlugin);
 
-        fastify.register(graaspValidationPlugin);
+        fastify.register(graaspValidationPlugin, {
+          // replace this with real api from prod
+          classifierApi: 'http://172.17.0.1:8080/sync',
```
pyphilia, line 214: Get this value from env.development, otherwise we won't be able to dynamically set it on aws.

pyphilia (changes requested):

pyphilia (approved):

PR #189, opened 2022-04-29
185/copy order

### LINKED ISSUE #171: Copy folder should persist children order
That's because the order is defined within the folder definition with the children ids (eg: [ item2Id, item4Id, item1Id ] ). When copying, this array still exists in the copied folder, but of course the ids don't match. We cannot update this array on copy in the backend because of the transaction order:
1. you have to copy the parent before the children  -> you don't know the id of the children to add in the parent
2. you cannot update the parent afterwards -> that would be a very complicated process anyway, and the transaction does not allow you to update something not yet added

From this I guess we have to find another way to store the order. Let me know what you think

### LINKED ISSUE #185: Copying an item copy its deleted content
Since copy is a basic operation, there's no prehook removing the deleted items.

Refactoring the recycle bin with tags might help: tags are copied alongside the item. However it won't work for explorer where the copy remove all the tags.

### LINKED ISSUE #188: Update fastify plugins
Fastify changed their packages name.

### LINKED ISSUE #190: Sort remaining children by date if they don't belong to childrenOrder
Applying such change should improve the frontend as the non-ordered elements tend to be randomly displayed. To investigate. 

```
const sort = order => (a,b)=> {
  if(order.indexOf(a.id)>=0 && order.indexOf(b.id)>=0) {
  return order.indexOf(a.id) - order.indexOf(b.id) 
  }
  
  if(order.indexOf(a.id)>=0) {
  	return -1
  }
  
  if(order.indexOf(b.id)>=0) {
  	return 1
  }
  
  return a.createdAt - b.createdAt 
  
}
```

pyphilia:
This PR updates the fastify plugins, set up sentry and fix the order of items on copy .

close #171  
close #188 
close #190


package.json
```diff
@@ -17,7 +17,7 @@
     "serve": "NODE_PATH=./dist node dist/server.js",
     "watch-node": "NODE_PATH=./dist nodemon dist/server.js",
     "watch": "concurrently -k -p \"[{name}]\" -n \"TypeScript,Node\" -c \"cyan.bold,green.bold\" \"yarn watch-ts\" \"yarn watch-node\"",
-    "test": "DISABLE_LOGS=true jest --forceExit --coverage --verbose",
+    "test": "DISABLE_LOGS=true jest test/item.test.ts --forceExit --coverage --verbose",
```
spaenleh, line 20: Modify this before merging

spaenleh (approved):

pyphilia:
I've added a little change in this PR: order children by createdAt if they're not part of the order set.
The copy action takes advantage of this to add a complete order in the copied elements (which was not the case before, so the items were still randomly distributed since they had the same createdAt value).

spaenleh (approved):

spaenleh (approved):
👍

PR #205, opened 2022-06-16
156 h5p plugin

### LINKED ISSUE #156: Register h5p plugin

dialexo:
Closes #156 
Register graasp-plugin-h5p

TODO: merge dependencies first, then update package.json with non-branch URLs


spaenleh (approved):
Looks good to me 🏂

.devcontainer/docker-compose.yml
```diff
@@ -19,10 +19,7 @@ services:
 
     volumes:
       - ..:/workspace:cached
-      - ${TMPDIR}:/tmp/graasp-file-item-storage
-        # On Linux (and other platforms) env var TMPDIR does not exist:
-        # replace "${TMPDIR}" with permanent existing path e.g.
-        # "/tmp/graasp-file-item-storage"
+      - ../tmp:/tmp/graasp-file-item-storage
```
pyphilia, line 22: Huuum should we revert the changes?

pyphilia (approved):
Looks good! We can sync to deploy on dev 😄

.devcontainer/docker-compose.yml
dialexo: I don't think so actually, that was my question: why are we arbitrarily binding the file item storage to some OS-specific local temp storage? I believe it's more convenient to have it bound to a locally visible `tmp` folder which is simply ignored from git and other tools; it also makes debugging nicer since the files can be seen from the editor directly. WDYT? @pyphilia @spaenleh @juancarlosfarah

PR #214, opened 2022-07-04
feat: register publish plugin

### LINKED ISSUE #204: register plugin-publish

louisewang1:
close #204


src/services/items/service-api.ts
```diff
@@ -204,6 +205,13 @@ const plugin: FastifyPluginAsync = async (fastify) => {
 
         fastify.register(graaspItemTags);
 
+        await fastify.register(graaspItemPublishPlugin, {
+          publishedTagId: PUBLISHED_TAG_ID,
+          publicTagId: PUBLIC_TAG_ID,
+          graaspActor: GRAASP_ACTOR,
+          explorerHostApi: CLIENT_HOSTS[2].hostname,
```
pyphilia, line 212: Let's not mention `explorer` as we always change the name lol Let's just go for hostname (it's not the api).
For the value, let's not depend on the index, you can use `.find`

pyphilia (changes requested):
Thanks for the PR, I've left concerning comments, but everything else looks good 🐞

PR #218, opened 2022-07-06
refactor: fix item's name regex, add permission filter to shared, use…

### LINKED ISSUE #208: Shared Items: return only items with membership write/admin
This might be really useful for the recycle bin, or showing in which items you can copy in.

### LINKED ISSUE #209: Use ENV variables for Host Map
A host map is defined, but won't work dynamically.
https://github.com/graasp/graasp/blob/00e2a0df03ece3a8d0774f98a6a323b2d8e3f707/src/util/config.ts#L39

### LINKED ISSUE #212: Change item name schema
Should change to `^\S+([ +\S⁄])*$` In order to allow 2 followed spaces.
`cannot start with a white-space character, allows multiple white spaces, prevent break lines`.

pyphilia:
This PR fixes:

- close #212 
- close #209 
- close #208


package.json
```diff
@@ -17,7 +17,7 @@
     "serve": "NODE_PATH=./dist node dist/server.js",
     "watch-node": "NODE_PATH=./dist nodemon dist/server.js",
     "watch": "concurrently -k -p \"[{name}]\" -n \"TypeScript,Node\" -c \"cyan.bold,green.bold\" \"yarn watch-ts\" \"yarn watch-node\"",
-    "test": "DISABLE_LOGS=true jest --forceExit --coverage --verbose",
+    "test": "DISABLE_LOGS=true jest test/item.test.ts --forceExit --coverage --verbose",
```
spaenleh, line 20: Remove then ready to merge

src/services/items/fluent-schema.ts
```diff
@@ -42,7 +42,7 @@ const item = S.object()
 // type 'base' (empty extra {})
 const baseItemCreate = S.object()
   .additionalProperties(false)
-  .prop('name', S.string().minLength(1).pattern('^\\S+( \\S+)*$'))
+  .prop('name', S.string().minLength(1).pattern('^[ \\S]+$'))
```
spaenleh, line 45: I think this is not the regex suggested in the related issue: `^\S+([ +\S⁄])*$`

The current one allows a single space as a name.

test/item.test.ts
```diff
@@ -433,7 +433,7 @@ describe('Item routes tests', () => {
       app.close();
     });
   });
-  describe('GET /items/shared-with', () => {
+  describe.only('GET /items/shared-with', () => {
```
spaenleh, line 436: Remove only when ready to merge

spaenleh (changes requested):
Good work !

Not sure if the regex for the item name fulfills the intent, it seems to be different from the one you suggest in the issue.

PR #219, opened 2022-07-06
feat: update member password

victorggonzalez:
This PR contains necessary functions to perform the password update operation


victorggonzalez:
I have included the updatePassword procedure inside auth. As we discussed, this way we have a dedicated endpoint to perform this operation

src/plugins/auth/auth.ts
```diff
@@ -329,6 +332,61 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+      // if the member does not have a password set: return true
+      return true;
+    }
+
+    async function encryptPassword(password: string) {
+      /* encrypted: stores the output of bcrypt.hash().
+      bcrypt.hash() creates the salt and hashes the password. 
+      A new hash is created each time the function is run, regardless of the password being the same. 
+      */
+      const saltRounds = 10;
```
pyphilia, line 359: Factor out in constants/settings.

src/plugins/auth/auth.ts
```diff
@@ -329,6 +332,61 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+    // update member password
+    fastify.patch<{ Params: IdParam }>(
+      '/members/:id/update-password',
+      { schema: updatePassword },
+      async ({ params: { id }, body, log }) => {
+        const member = await mS.get(id, db.pool);
+        // verify that input current password is the same as the stored one
+        const verified = await verifyCurrentPassword(member, body['currentPassword']);
+        if (verified) {
+          delete body['currentPassword'];
```
pyphilia, line 376: You don't need to do this right?

src/plugins/auth/auth.ts
```diff
@@ -329,6 +332,61 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+          // auto-generate a salt and a hash
+          const hash = await encryptPassword(body['password']);
+          const tasks = memberTaskManager.createUpdateTaskSequence(member, id, {
+            password: hash,
+          });
+          return runner.runSingleSequence(tasks, log);
+        } else {
+          // throw error if password verification fails
+          throw new InvalidPassword();
+        }
```
pyphilia, line 386: To simplify, I would do `if(!verified) throw ...` so you don't have to use a else clause.

src/services/members/service-api.ts
```diff
@@ -115,8 +148,26 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+            } else {
+              // throw error if password verification fails
+              throw new InvalidPassword();
+            }
+          } else {
+            const tasks = taskManager.createUpdateTaskSequence(member, id, body);
+            return runner.runSingleSequence(tasks, log);
+          }
         },
       );
```
pyphilia, line 173: Remove all changes made in this file, since it is used in auth now.

src/util/graasp-error.ts
```diff
@@ -316,6 +316,20 @@ export class OrphanSession extends BaseGraaspError {
...
 }
 
+export class InvalidPassword extends BaseGraaspError {
+  constructor(data?: unknown) {
+    // this status code is custom for the browser to know it needs to refresh its token
+    super(
+      {
+        code: 'GERR025',
+        statusCode: StatusCodes.UNAUTHORIZED,
+        message: 'Current password is invalid. Please try again.',
```
pyphilia, line 326: Could you create a branch in translations as well? Or an issue? So we can later translate the errors.

pyphilia (changes requested):
Thanks for the change, I've added my comments, let me know if you have questions 🚥

src/plugins/auth/auth.ts
```diff
@@ -329,6 +332,61 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+      return encrypted;
+    }
+
+    // update member password
+    fastify.patch<{ Params: IdParam }>(
+      '/members/:id/update-password',
+      { schema: updatePassword },
+      async ({ params: { id }, body, log }) => {
+        const member = await mS.get(id, db.pool);
+        // verify that input current password is the same as the stored one
```
pyphilia, line 373: Actually you never check that actor is the member. But actually we should go for the following endpoint `/members/update-password` and assume it's for the current member. We will have another endpoint for admin where we can precise the member id.

pyphilia (changes requested):

src/plugins/auth/auth.ts
```diff
@@ -329,6 +364,30 @@ const plugin: FastifyPluginAsync<AuthPluginOptions> = async (fastify, options) =
...
+          }
+          throw new InvalidPassword();
+        }
+        // auto-generate a salt and a hash
+        const hash = await encryptPassword(body['password']);
+        const tasks = memberTaskManager.createUpdateTaskSequence(member, member.id, {
+          password: hash,
+        });
+        return runner.runSingleSequence(tasks, log);
+        // }
```
pyphilia, line 387: Remove?

pyphilia (changes requested):
It looks good! I've added some more comments but it looks good to merge! 🛍️

PR #222, opened 2022-07-08
feat: not return password value

victorggonzalez:


pyphilia (changes requested):
Thanks for the PR! 
But actually, from what we decided for the password mechanism, we don't need the password field (even as a boolean). So I think the best solution to not return the password is to add a schema to this endpoint. Let me know if it's not clear^^

victorggonzalez:
Done! Let me know what you think of this

src/services/members/schemas.ts
```diff
@@ -35,6 +47,14 @@ export default {
   },
 };
 
+// schema for getting current member
+const getCurrent = {
+  // params: { $ref: 'http://graasp.org/#/definitions/idParam' },
```
pyphilia, line 52: Remove.

src/services/members/schemas.ts
```diff
@@ -13,6 +13,18 @@ export default {
...
     },
 
+    currentMember: {
+      type: 'object',
+      properties: {
+        id: { type: 'string' },
+        name: { type: 'string' },
+        email: { type: 'string' },
+        type: { type: 'string' },
+        extra: { type: 'object', additionalProperties: true },
```
pyphilia, line 23: Maybe add createdAt and updatedAt?
Change to `additionalProperties: false`?

pyphilia (changes requested):
Looks good! Do you think you could update the test for /members/current? (member.test.ts) We would like to test that password does not exist in the response, but the other fields do.

pyphilia (approved):

Issue #223, opened 2022-07-11
Use redis password


spaenleh:
Add info on how to secure redis instance.

Remove port forwarding for redis db.

dialexo:
There seems to be a configuration issue on Linux (maybe?). When the devcontainer boots alongside its services defined in `.devcontainer/docker-compose.yml`, it seems that somehow the Redis instance does not pick up the password defined in `.devcontainer/.env`. Which is weird since the official docs say that

> docker compose will automatically pick up a file called .env in the folder containing the docker-compose.yml, but you can also create one in another location.
> https://code.visualstudio.com/remote/advancedcontainers/environment-variables

However the output of `docker logs graasp-redis` on my machine:

```
*** FATAL CONFIG FILE ERROR (Redis 7.0.0) ***
Reading the configuration file, at line 2
>>> 'requirepass'
wrong number of arguments
```

Not sure if it's caused by some weird way VSCode works on Linux, but it seems to work on Docker Desktop for MacOS (my other dev machine works just fine... and I don't see what else is different) :thinking:

pyphilia:
Since you're the only one with a linux machine, I reassign this issue to you @dialexo , feel free to tackle it whenever you have free time.

PR #230, opened 2022-07-15
refactor: include redis port forwarding

victorggonzalez:
Whenever you want to run the core locally outside of the container, it is necessary to have this for the app to recognize the redis port. It is a frequent issue with the m1 chip.
It is not harmful to have this forwarding as all of the rest of the services have them.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -58,6 +58,8 @@ services:
     # with the value of the variable set.
     command: redis-server --requirepass $REDIS_PASSWORD
     restart: unless-stopped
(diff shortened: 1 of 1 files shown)


pyphilia:
I don't know much about it, but I'll let @spaenleh and @dialexo give their reviews.

spaenleh (approved):
Thank you for the PR 📡 @pyphilia, we discussed this issue with @victorggonzalez and since the redis instance is now protected by a password by default, forwarding the port is not an issue. I removed it in #223 because i thought it was not useful. I had missed Victor's use-case.

PR #254, opened 2022-08-25
249-allow-reply-of-signed-url-of-file

alvrba:


alvrba:
It includes:
- Files refactor (return URL instead of reply) in local and S3 services, with documentation with the changes.
- Cache policy in the /download endpoint, that matches the expiration of the signed S3 file URL
- Integration for the new Graasp PDF Viewer

pyphilia (approved):

PR #280, opened 2022-11-24
feat: add isResizable to schema

### LINKED ISSUE #279: Add `isResizable` in schema
Related to https://github.com/graasp/graasp-sdk/pull/61 and https://github.com/graasp/graasp-builder/issues/500

spaenleh:
This PR adds the `isResizable` prop to the schema.

closes #279

### DIFF
--- a/src/services/items/fluent-schema.ts
+++ b/src/services/items/fluent-schema.ts
@@ -18,6 +18,7 @@ const settings = S.object()
   .prop('isPinned', S.boolean())
   .prop('tags', S.array())
   .prop('showChatbox', S.boolean())
+  .prop('isResizable', S.boolean())
   .prop('hasThumbnail', S.boolean())
   .prop('ccLicenseAdaption', S.string())
   .prop('displayCoEditors', S.boolean());


src/services/items/fluent-schema.ts
```diff
@@ -18,6 +18,7 @@ const settings = S.object()
   .prop('isPinned', S.boolean())
   .prop('tags', S.array())
   .prop('showChatbox', S.boolean())
+  .prop('isResizable', S.boolean())
```
dialexo, line 21: Now that I think about it, does it make sense that this is set in settings that can be applied to all items? Do we have a way to restrict the type to apps only?

dialexo (approved):

src/services/items/fluent-schema.ts
spaenleh: In the builder, the setting will not be shown when the item type is not an app. In the player, the setting is only used when the item type is an app. But yes you could send a request to add this setting to any item type. I did not see a mechanism to prevent some settings based on the item type.

PR #291, opened 2022-12-29
Isolate transactions

### LINKED ISSUE #287: Fix transactions
- The logic of transaction is separated from the logic of task execution, but it's impossible to abstract transactions at the data access layer because they can only be defined from the requirements of the business logic
- It is not possible to forward an open transaction to a new task execution call. This means that all `runX(tasks(s))` calls create new transactions, even if they were supposed to be dependent. E.g. if the body of a task or any pre/post hook handler for this task executes taskRunner.runX, the transaction will wait (and stall the connection), while new transactions are created

I'm not sure we can just skip transactions everywhere. This is because tasks themselves may have handlers that are dependencies for the task, but they may execute out of order. So e.g. permission check may happen after read, or deleting for instance

What we can (and should probably) do is make sure that task bodies and pre / post handlers NEVER create tasks / invoke the task runner. This will always result in creating nested transactions. We should instead make sure that task bodies and pre / post handlers ALWAYS invoke the corresponding dbservice, and to pass the handler currently being used in the context (and NEVER the global db pool).

The transaction may hence take more time (not even sure, since they are already waiting for additional transactions anyway), but the number of connections will be reduced by the number of additional nested calls

#282 will provide a more standardized approach to transaction handling.

dialexo:
This PR removes superfluous DB calls and connections by ensuring that transactions do not establish additional connections

- Reset the pool size to a sane default
- Update dependencies which erroneously bloat transactions

Related:
- https://github.com/graasp/graasp-apps/pull/67
- https://github.com/graasp/graasp-plugin-chatbox/pull/47
- https://github.com/graasp/graasp-plugin-file-item/pull/39
- https://github.com/graasp/graasp-plugin-h5p/pull/11
- https://github.com/graasp/graasp-plugin-hidden-items/pull/8
- https://github.com/graasp/graasp-plugin-recycle-bin/pull/41
- https://github.com/graasp/graasp-plugin-thumbnails/pull/28
- https://github.com/graasp/graasp-plugin-invitations/pull/7

Closes #287

### DIFF
--- a/src/util/config.ts
+++ b/src/util/config.ts
@@ -113,8 +113,12 @@ export const REFRESH_TOKEN_EXPIRATION_IN_MINUTES =
 
 /**
  * Maximum connections to the DB for slonik
+ * https://wiki.postgresql.org/wiki/Number_Of_Database_Connections
+ * should be around ((core_count * 2) + effective_spindle_count), to be tuned wrt effective workload
  */
-export const MAXIMUM_POOL_SIZE = 100;
+export const MAXIMUM_POOL_SIZE = process.env.MAXIMUM_POOL_SIZE
+  ? parseInt(process.env.MAXIMUM_POOL_SIZE)
+  : 20;
 
 // Graasp mailer config
 export const MAILER_CONFIG_SMTP_HOST = process.env.MAILER_CONFIG_SMTP_HOST;
--- a/yarn.lock
+++ b/yarn.lock
@@ -5224,7 +5224,7 @@ __metadata:
 
 "graasp-plugin-invitations@github:graasp/graasp-plugin-invitations":
   version: 0.1.0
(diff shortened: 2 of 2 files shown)


dialexo:
### TODO

Still missing fix from:
- [ ] https://github.com/graasp/graasp-plugin-invitations/pull/7

Behaviours to test (from the list of modified plugins):

- graasp-apps
	- [ ] when an app item is copied, if an app setting has `data.type` has provided file item type, the corresponding file should also be copied
-  graasp-plugin-chatbox
	- [ ]
-  graasp-plugin-file-item
	- [ ]
-  graasp-plugin-h5p
	- [ ]
-  graasp-plugin-hidden-items
	- [ ]
-  graasp-plugin-recycle-bin
	- [ ]
-  graasp-plugin-thumbnails
	- [ ]

spaenleh (approved):
LGTM ! 🥑

pyphilia:
I think it crashed for the first point (copy app setting file).

PR #303, opened 2023-01-16
feat: add registration stack trace for task hook handlers

### LINKED ISSUE #298: Add stack trace in hooks
Currently, the hooks system is completely opaque once the listeners are registered. It is impossible to determine the caller source when a hook fails, because the error will originate from the task runner and the stack trace is hence reduced to the path from the server entry point to the corresponding task runner line. Since hooks are registered externally, the register call site is completely lost.

Proposal: when a hook is registered, save the trace of the registration point inside the task runner. When a given hook fails, log this information alongside additional context information.

Example implementation:

Plugins and callers typically register their handlers like this:

```ts
taskRunner.registerTaskPostHookHandler(taskName, (params) => { ... })
```

Currently, the task runner system implements something like the following:

```ts
class TaskRunner {
    handler?: () => {};

    register(fn: () => {}) {
        this.handler = fn;
    }

    execute() {
        try {
            this.handler?.();
        } catch (error) {
            if (error instanceof Error) console.log(error.stack)
        }
    }

    scheduleExecute() {
        // later on some event, trigger execute
        setTimeout(() => this.execute(), 1000);
    }
}

const taskRunner = new TaskRunner();

export { taskRunner };
```

This will result in the following task trace if the registered hook fails:

```ts
import { taskRunner } from "./TaskRunner";

const doFail = () => {
    throw new Error();
};

taskRunner.register(doFail);
taskRunner.scheduleExecute();
```

```
Error
    at TaskRunner.doFail [as handler] (<path>/test.js:5:11)
    at TaskRunner.execute (<path>/TaskRunner.js:13:73)
    at Timeout._onTimeout (<path>/TaskRunner.js:23:47)
    at listOnTimeout (node:internal/timers:557:17)
    at processTimers (node:internal/timers:500:7)
```

As expected, the registration call site information is lost, and we have no idea what plugin / server behaviour caused the server to fail.

We can however quite simply save the call site stack trace by adding the following workaround:

```ts
class TaskRunner {
  handler?: () => {};
  callSite?: string;

  register(fn: () => {}) {
    this.handler = fn;
    // obtain current stack trace by artificially generating an error on the stack
    try {
      throw new Error();
    } catch (error) {
      if (error instanceof Error) this.callSite = error.stack; // <-- get the value immediately
    }
  }

  execute() {
    try {
      this.handler?.();
    } catch (error) {
      if (error instanceof Error) console.log(error.stack);
      console.log(this.callSite);
    }
  }

  scheduleExecute() {
    // later on some event, trigger execute
    setTimeout(() => this.execute(), 1000);
  }
}

const taskRunner = new TaskRunner();

export { taskRunner };
```

Using the same register code as before, this results in the following stack trace:

```ts
Error
    at TaskRunner.doFail [as handler] (<path>/test.js:5:11)
    at TaskRunner.execute (<path>/TaskRunner.js:21:73)
    at Timeout._onTimeout (<path>/TaskRunner.js:32:47)
    at listOnTimeout (node:internal/timers:557:17)
    at processTimers (node:internal/timers:500:7)
Error
    at TaskRunner.register (<path>/TaskRunner.js:11:19)
    at Object.<anonymous> (<path>/test.js:7:25)        <-- call site is here
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:17:47
```

dialexo:
Closes #298: this PR will allow us to receive more debugging information when a task hook handler fails and crashes the transaction; the registration call stack will be displayed.

### DIFF
--- a/src/services/global-task-runner.ts
+++ b/src/services/global-task-runner.ts
@@ -22,6 +22,20 @@ export class GlobalTaskRunner implements TaskRunner<Actor> {
   protected logger: FastifyLoggerInstance;
   private getIdIfExists = ({ id }: { id?: string }) => id;
 
+  private tasksHooks = new Map<
+    string,
+    {
+      pre?: {
+        handlers: { fn: PreHookHandlerType<unknown>; registrationCallStack: string }[];
+        wrapped: PreHookHandlerType<unknown>;
+      };
+      post?: {
+        handlers: { fn: PreHookHandlerType<unknown>; registrationCallStack: string }[];
+        wrapped: PostHookHandlerType<unknown>;
+      };
+    }
+  >();
+
   constructor(database: Database, logger: FastifyLoggerInstance) {
     this.databasePool = database.pool;
     this.logger = logger;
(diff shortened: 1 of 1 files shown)


dialexo:
Superseded by #302

PR #305, opened 2023-01-18
fix: return shared sibling items

### LINKED ISSUE #304: Two shared sibling items only return one

pyphilia:
close #304 

I couldn't think of another way of filtering the items through sql. 
The filtering function could be optimized (I'm thinking of sorting by path, then one loop comparing the shortest path and current path, filter out if the current path is a children of the shortest path).


spaenleh (approved):
I am not sure what the previous query was doing exactly, but I guess the current query makes sens !

PR #308, opened 2023-01-23
307/ChangeGraaspMobileAppNameLink

alvrba:


alvrba:
Change the graasp to the graasp-mobile-builder app name in the link we send to open the app when logging in with the email magic link

pyphilia (approved):
LGTM 👍

PR #327, opened 2023-02-07
fix(metrics): use router path for sentry

dialexo:
The current code for sentry handles each separate request as a transaction group. However, this is not very useful since for instance each item id generates its own set of metrics, so we cannot aggregate e.g. the average time to get an item. This PR fixes this issue by using the `routerPath` property, which is the URL used by the client request before substituting request parameters (such as item id). Now, instead of having statistics grouped like this:

- GET /item/item-id-A
- GET /item/item-id-B

They will now appear under the unified group

-  GET /item/:itemId


spaenleh (approved):
Nice !

PR #336, opened 2023-02-23
fix: bump @sentry dependencies

### LINKED ISSUE #335: Types issues with Sentry build
Deployments currently fail https://github.com/graasp/graasp/actions/runs/4244487093/jobs/7378675099#step:6:5990

The type of the Sentry integrations array somehow does not match, even though both definition and usage come from the SDK. New versions are available anyway, so bumping all the versions (matching) may fix the issue.

dialexo:
Closes #335

Bumps all @sentry dependencies to matching versions


pyphilia (approved):
Let's merge!! 🦆

PR #374, opened 2023-04-19
feat: add etherpad zip import / export

### LINKED ISSUE #373: Add etherpad import / export

dialexo:
Closes #373

Also includes:
chore: add script to populate db from readme


setup-db.js
spaenleh, line 1: Haleluia ! We just need to ensure that the links are in the right order ?

spaenleh (approved):
🌷 Looks good to me ! Also thank you for the nice script ! 💪

setup-db.js
dialexo: Yes, and also that there are no other unrelated `https://github.com/<anything>.sql` links. Though we will not need the script anymore after the TypeORM refactor i guess 🤷

PR #378, opened 2023-05-02
chore: update postgres container to 15

morganridel:
Databases will be slowly updated to 15.2, currently only the one on Dev environement is done.

We match the local database version to the expected deployment version for consistency

### DIFF
--- a/.devcontainer/db/Dockerfile
+++ b/.devcontainer/db/Dockerfile
@@ -1,3 +1,3 @@
-FROM postgres:13
+FROM postgres:15
 COPY db-schema.sql /docker-entrypoint-initdb.d/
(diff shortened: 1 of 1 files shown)


.devcontainer/db/Dockerfile
```diff
@@ -1,3 +1,3 @@
-FROM postgres:13
+FROM postgres:15.2
```
dialexo, line 1: Would it make sense to only use the major version?

morganridel: I put the minor version to be exactly on the same version as AWS, but using the major would be fine too as minor version shouldn't affect the environment consistency that much

dialexo (approved):

spaenleh (approved):

PR #382, opened 2023-06-07
refactor: chat ws hooks

morganridel:
Only publish hooks for now

For some reason, the message is first displayed as anonymous but is correct after refreshing?

image


src/services/chat/ws/hooks.ts
```diff
@@ -1,60 +1,52 @@
...
+  chatService: ChatMessageService,
+  itemService: ItemService,
+) {
+  websockets.register(itemChatTopic, async (req) => {
+    const { channel: itemId, member } = req;
+    // item must exist with read permission
+    const item = await itemService.get(member, repositories, itemId, PermissionLevel.Read);
+    if (!item) {
+      throw new ItemNotFound();
+    }
```
pyphilia, line 22: I think that if you use the service (or even the repository) it already checks if the item exists for you.

pyphilia (commented):
I don't know how websockets work, but I've left a minor comment^^

src/services/chat/ws/hooks.ts
```diff
@@ -1,60 +1,52 @@
...
+  websockets: WebsocketService,
+  chatService: ChatMessageService,
+  itemService: ItemService,
+) {
+  websockets.register(itemChatTopic, async (req) => {
+    const { channel: itemId, member } = req;
+    // item must exist with read permission
+    const item = await itemService.get(member, repositories, itemId, PermissionLevel.Read);
+    if (!item) {
+      throw new ItemNotFound();
```
spaenleh, line 21: Linked to my comment I think here the connection should simply be rejected with `reject(<the error>)` ? Maybe I am missing something, @dialexo can you help ?

spaenleh (commented):
To the best of my knowledge of the backend and WS, I would say it looks good, see my comment about the reject.

spaenleh (approved):
LGTM !

PR #385, opened 2023-06-12
Fix item controller

morganridel:
* bundling the action and item modification in the same transaction
  * is there a reason not to do this that I missed?
* adding await inside transactions, this is what is recomended in the docs
  * image
  * awaiting the transaction themselve is debatable depending on our needs

waiting to see if it doesn't break the CI in other places

### DIFF
--- a/src/services/item/controller.ts
+++ b/src/services/item/controller.ts
@@ -140,11 +140,11 @@ const plugin: FastifyPluginAsync = async (fastify) => {
         body,
         log,
       } = request;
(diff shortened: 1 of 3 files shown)


spaenleh (approved):
🌃 LGTM, but I am not as knowledgeable as others. 

Maybe the idea behind not awaiting the actions was to try to avoid the request failing in case the actions failed ? But there could be a way closer to the docs recommendations.

PR #391, opened 2023-06-14
fix: create membership for copied item

### LINKED ISSUE #390: Copy an Item does not related create membership

pyphilia:
close #390

### DIFF
--- a/src/services/item/repository.ts
+++ b/src/services/item/repository.ts
@@ -18,6 +18,7 @@ import {
   ItemNotFolder,
   ItemNotFound,
   TooManyDescendants,
+  UnexpectedError,
 } from '../../utils/errors';
 import { Member } from '../member/entities/member';
 import { mapById } from '../utils';
@@ -320,10 +321,15 @@ export const ItemRepository = AppDataSource.getRepository(Item).extend({
 
     _fixChildrenOrder(treeItemsCopy);
     // return copy item + all descendants
-    const createdItem = await this.insert([...treeItemsCopy.values()].map(({ copy }) => copy));
+    const createdOp = await this.insert([...treeItemsCopy.values()].map(({ copy }) => copy));
+
+    const newItemRef = createdOp?.identifiers?.[0];
+    if (!newItemRef) {
+      throw new UnexpectedError({ operation: 'copy', itemId: item.id });
+    }
 
     // TODO: copy item + all descendants
-    return this.get(createdItem.identifiers[0].id);
+    return this.get(newItemRef.id);
   },
 
   // UTILS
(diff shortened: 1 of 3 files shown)


src/services/item/service.ts
```diff
@@ -388,6 +389,22 @@ export class ItemService {
...
+        permission: PermissionLevel.Admin,
+      })
+      .catch((e) => {
+        // admin permission already exists and does not need to be added
+        if (e instanceof InvalidMembership) {
+          return;
+        }
+        throw e;
+      });
+
```
morganridel, line 407: Shouldn't this be a single DB transaction? (i.e if the second query fails, we don't want ab Item without membership? What about items that were copied before this change?

It's a bit complicated to rebuild all the repositories everywhere when you need a transaction, I don't know if it could be done better

Here we could do something like this:
```
await this.db.transaction(async (manager) => {
      const repositories = buildRepositories(manager)
      const result = await repositories.itemRepository.copy(item, actor, parentItem)
      await repositories.itemMembershipRepository
        .post({
          item: result,
          member: actor,
          creator: actor,
          permission: PermissionLevel.Admin,
        })
        // Ideally here rather catch the specific exception, so that the item can rollback if this is an unexpected exception
        .catch((e) => {
          // admin permission already exists and does not need to be added
        });
    })
```

pyphilia: It's already one transaction since it's defined in the `controller`. I felt like creating a membership is "related" to creating an item, that's why I've put it in the item service (it's also because of some habits). But I'm open to put it in the controller. However since we run hooks in copy, if you don't have your membership beforehand some of the hooks (that might check your authorization) will fail.

morganridel: You're right, no need to move.
I actually would expect the controller to only handle general request thing (authentication pre-handler is good example) and HTTP requests details, then only calling one service.
The service can call repositories but also other services, so in my opinion the membership creation is at the right place. 

Eventually I'd think about not creating transaction in controllers but only in services, but that would be for a future change

morganridel (approved):
LGTM

src/services/item/repository.ts
```diff
@@ -320,10 +320,10 @@ export const ItemRepository = AppDataSource.getRepository(Item).extend({
 
     _fixChildrenOrder(treeItemsCopy);
     // return copy item + all descendants
-    await this.insert([...treeItemsCopy.values()].map(({ copy }) => copy));
+    const createdItem = await this.insert([...treeItemsCopy.values()].map(({ copy }) => copy));
 
     // TODO: copy item + all descendants
-    return item;
+    return this.get(createdItem.identifiers[0].id);
```
dialexo, line 326: Q: what is `identifiers[0]`? Is an indexed access required, and does it always exist? Consider using the destructuration syntax

src/services/item/service.ts
dialexo: @morganridel I agree, but how do you handle which service should initiate the transaction? E.g. if item service needs the item memberships service, how does it pass the current transaction (also if some other services uses item membership without initiating its own transaction)?

dialexo (commented):
LGTM but let me know what you think about the 2 questions I've left

PR #399, opened 2023-06-16
chore: delete postcheckout husky hook

spaenleh:
This PR removes the `postcheckout` husky hook as it caused unwanted installs when rebasing, checking out other branches etc.
It has been evaluated it was more cumbersome thatn helpful, so it is removed.


dialexo (approved):
Thank you :hamburger: this hook sure was intrusive :stuck_out_tongue:

PR #400, opened 2023-06-20
fix: update deployment workflows

pyphilia:
This PR updates the deployment workflows

ref https://github.com/graasp/graasp-deploy/pull/201


dialexo (approved):
LGTM, are the env vars to disable plugins and non-secret configs such as paths completely deleted?

pyphilia:
We don't use them anymore at least, they shouldn't be used in the core anymore.

PR #403, opened 2023-06-20
fix: prevent deletion of only / last admin on item

### LINKED ISSUE #384: Creators can lose their own admin privilege by themselves
- Prevent removing admin if there are no more admins left
- Should this be allowed in the general case? Should a user be able to "unadmin" themselves? What if they were the creator?
- On the front-end, at least a modal should inform the admin that they are about to lose their right

dialexo:
This PR checks that on deletion of an item membership, the user is not the last admin on the associated item. If that's the case, the server throws a bad request instead (is there a better status code?).

Closes #384


spaenleh (approved):
LGTM ! 👨‍🍳

morganridel (approved):

PR #408, opened 2023-06-21
fix: make membership websockets hooks work again

morganridel:


src/services/itemMembership/index.ts
```diff
@@ -11,6 +11,7 @@ import { ItemMembership } from './entities/ItemMembership';
 import { PurgeBelowParam } from './interfaces/requests';
 import common, { create, createMany, deleteAll, deleteOne, getItems, updateOne } from './schemas';
 import ItemMembershipService from './service';
+import { registerItemMembershipWsHooks } from './ws/hooks';
 
 // import { registerItemMembershipWsHooks } from './ws/hooks';
```
pyphilia, line 16: Delete.

pyphilia (approved):
LGTM but I think @dialexo should have a look

src/services/itemMembership/index.ts
```diff
@@ -36,16 +38,14 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
-      //   registerItemMembershipWsHooks(
-      //     websockets,
-      //     runner,
-      //     itemsDbService,
-      //     dbService,
-      //     taskManager,
-      //     db.pool,
-      //   );
-      // }
+      if (WEBSOCKETS_PLUGIN) {
```
dialexo, line 41: I think you can remove the `WEBSOCKETS_PLUGIN` check, we don't use it anymore for the items

src/services/itemMembership/index.ts
```diff
@@ -36,16 +38,14 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+        registerItemMembershipWsHooks(
+          buildRepositories(),
```
dialexo, line 43: I guess it's fine to pass the repos for now. In the items, I've called `buildRepositories` in each function, but the idea is to replace the injection by letting each service resolve its own dependencies anyway.

src/services/itemMembership/service.ts
```diff
@@ -18,6 +19,7 @@ export class ItemMembershipService {
   itemService: ItemService;
   hosts: Hostname[];
   mailer: MailerDecoration;
+  hooks = new HookManager();
```
dialexo, line 22: You can narrow the generic type on the hook manager to get precise types on the `data` argument for both pre and post hooks individually. This will also provide types for the event names themselves.

See:

https://github.com/graasp/graasp/blob/36223a3508df54ce6ff5a75c689449ed02b547da/src/services/item/service.ts#L29-L35

See the type definition at

https://github.com/graasp/graasp/blob/36223a3508df54ce6ff5a75c689449ed02b547da/src/utils/hook.ts#L4-L56

src/services/itemMembership/service.ts
```diff
@@ -69,6 +71,7 @@ export class ItemMembershipService {
     });
 
     await this._notifyMember(actor, repositories, member, item);
+    await this.hooks.runPostHooks('create', actor, repositories, result);
```
dialexo, line 74: Should we run the prehook as well for consistency before the repo call?

src/services/itemMembership/service.ts
```diff
@@ -127,7 +130,16 @@ export class ItemMembershipService {
...
 
-    return itemMembershipRepository.post({ item, member, creator: actor, permission });
+    const result = await itemMembershipRepository.post({
+      item,
+      member,
+      creator: actor,
+      permission,
+    });
+
+    await this.hooks.runPostHooks('create', actor, repositories, result);
```
dialexo, line 140: Same as above, prehook?

src/services/itemMembership/service.ts
```diff
@@ -146,7 +158,14 @@ export class ItemMembershipService {
...
       memberships.map(async ({ memberId, permission }) => {
         const member = await memberRepository.get(memberId);
-        return itemMembershipRepository.post({ item, member, creator: actor, permission });
+        const result = await itemMembershipRepository.post({
+          item,
+          member,
+          creator: actor,
+          permission,
+        });
+        await this.hooks.runPostHooks('create', actor, repositories, result);
```
dialexo, line 167: Same as above

src/services/itemMembership/service.ts
```diff
@@ -165,7 +184,11 @@ export class ItemMembershipService {
     const iM = await itemMembershipRepository.get(itemMembershipId);
     await validatePermission(repositories, PermissionLevel.Admin, actor, iM.item);
 
-    return itemMembershipRepository.patch(itemMembershipId, data);
+    const result = await itemMembershipRepository.patch(itemMembershipId, data);
+
+    await this.hooks.runPostHooks('update', actor, repositories, result);
```
dialexo, line 189: Same as above

src/services/itemMembership/service.ts
```diff
@@ -196,7 +219,13 @@ export class ItemMembershipService {
       throw new CannotDeleteOnlyAdmin(item);
     }
 
-    return itemMembershipRepository.deleteOne(itemMembershipId, { purgeBelow: args.purgeBelow });
+    const result = await itemMembershipRepository.deleteOne(itemMembershipId, {
+      purgeBelow: args.purgeBelow,
+    });
+
+    await this.hooks.runPostHooks('delete', actor, repositories, result);
```
dialexo, line 226: Same as above

src/services/itemMembership/ws/hooks.ts
```diff
@@ -1,90 +1,74 @@
...
-//       }
-//       websockets.publish(itemMembershipsTopic, item.id, ItemMembershipEvent('create', membership));
-//     },
-//   );
+  // on create:
+  // - notify member of new shared item IF creator != member
+  // - notify item itself of new membership
+  itemMembershipService.hooks.setPostHook(
+    'create',
+    async (member, repositories, membership: ItemMembership) => {
```
dialexo, line 28: If you type the HookManager instance, you will get the typings here for free (and at any usage site), as well as event names

dialexo (changes requested):
Overall LGTM 🌮  I think the types can be improved by specifying the generic. 
Let me know what you think and I'll be happy to approve.
Thank you!

src/services/itemMembership/index.ts
```diff
@@ -36,16 +34,12 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
-      //     taskManager,
-      //     db.pool,
-      //   );
-      // }
+      registerItemMembershipWsHooks(
+        buildRepositories(),
+        websockets,
+        itemService,
+        itemMembershipService,
+      );
```
dialexo, line 42: Looks like a review comment I was adding got removed by the cat walking on the keyboard 🐈 :

In the items I have packed the destructuration and the call into a fastify plugin, so that it's more consistent:

https://github.com/graasp/graasp/blob/36223a3508df54ce6ff5a75c689449ed02b547da/src/services/item/ws/hooks.ts#L180-L184

So the registration now looks like this:

https://github.com/graasp/graasp/blob/36223a3508df54ce6ff5a75c689449ed02b547da/src/services/item/index.ts#L101

We may want to do the same here

src/services/itemMembership/service.ts
```diff
@@ -117,17 +127,28 @@ export class ItemMembershipService {
   async post(
     actor: Actor,
     repositories: Repositories,
-    { permission, itemId, memberId }: { permission: PermissionLevel; itemId: UUID; memberId: UUID },
+    im: { permission: PermissionLevel; itemId: UUID; memberId: UUID },
```
dialexo, line 130: I got confused while reading
```suggestion
    membership: { permission: PermissionLevel; itemId: UUID; memberId: UUID },
```

src/services/itemMembership/service.ts
```diff
@@ -179,7 +218,8 @@ export class ItemMembershipService {
     }
     const { itemMembershipRepository } = repositories;
     // check memberships
-    const { item } = await itemMembershipRepository.get(itemMembershipId);
+    const membership = await itemMembershipRepository.get(itemMembershipId);
+    const item = membership.item;
```
dialexo, line 222:
```suggestion
    const { item } = membership;
```

dialexo (changes requested):
Just a few nitpicks 🌵

src/services/itemMembership/service.ts
morganridel: I think destructuring is often confusing as a syntax and only improves readability in specific cases, but rarely for one element :D I'll follow the customs though

src/declarations.d.ts
```diff
@@ -48,6 +49,9 @@ declare module 'fastify' {
       thumbnails: {
         service: ItemThumbnailService;
       };
+      membserships: {
```
dialexo, line 52:
```suggestion
      memberships: {
```

src/declarations.d.ts
```diff
@@ -48,6 +49,9 @@ declare module 'fastify' {
...
+        service: ItemMembershipService;
+      };
     };
     members: { service: MemberService };
```
dialexo, line 56: I think we used to have the membership service decorated at the top level (and not nested inside items). WDYT?

src/services/itemMembership/index.ts
```diff
@@ -5,25 +5,18 @@ import { FastifyPluginAsync } from 'fastify';
...
-  const {
-    db,
-    items: { service: itemService },
-    mailer,
-    hosts,
-  } = fastify;
-  const itemMembershipService = new ItemMembershipService(itemService, hosts, fastify.mailer);
+  const { db, items, mailer, hosts } = fastify;
+  const itemMembershipService = new ItemMembershipService(items.service, hosts, mailer);
+  items.membserships = { service: itemMembershipService };
```
dialexo, line 19: Considering the above, you would instead have:
```suggestion
  fastify.decorate('memberships', { service: itemMembershipService });
```
So it wouldn't depends on the item service being decorated. However I don't know if it breaks any scoping rules

dialexo (commented):
Let me know what you think about the decoration

src/declarations.d.ts
morganridel: sounds good, I didn't know any rules to determine if it should be nested or not

dialexo (approved):
LGTM 🦜 Thank you!

PR #423, opened 2023-06-23
feat: allow to edit file item extra's alttext

### LINKED ISSUE #422: Allow updating extra of `file` and `s3File` item types
graasp/graasp-sdk#163 introduced a new extra for `file` and `s3File` item types to allow to set an `altText` attribute on those items for a11y. 
graasp/graasp-ui#356 and graasp/graasp-builder#631 implement this new feature.

The schema of the updateItem request needs to be updated to allow this change.

We should make sure that the schema specifically disallows to change the existing keys like `name`, `path`, `size` and `mimetype` as these are set when uploading the file and should not be changed be the user. All other keys should be allowed to be set in order to support future updates to the extra object.

pyphilia:
close #422


src/services/item/plugins/file/test/index.test.ts
```diff
@@ -442,6 +442,35 @@ describe('File Item routes tests', () => {
...
+      ({ item } = await saveItemAndMembership({ item: MOCK_FILE_ITEM, member }));
+    });
+
+    it('Edit file item altText', async () => {
+      const response = await app.inject({
+        method: HttpMethod.PATCH,
+        url: `${ITEMS_ROUTE_PREFIX}/${item.id}`,
+        payload: { extra: { [FILE_ITEM_TYPE]: { altText: 'new name' } } },
+      });
+      console.log(response.json());
```
pyphilia, line 460: Remove console.log

package.json
```diff
@@ -65,7 +65,7 @@
     "@fastify/static": "6.10.2",
     "@fastify/view": "7.4.1",
     "@fastify/websocket": "7.2.0",
-    "@graasp/sdk": "1.0.0",
+    "@graasp/sdk": "github:graasp/graasp-sdk#162-add-alttext-to-settings-fileitem",
```
spaenleh, line 68: It was released in sdk 1.1.0

spaenleh (approved):
Looks good to me, even if the extendSchema stuff looks like magic 🌠

src/services/item/plugins/file/index.ts
```diff
@@ -45,9 +45,8 @@ const basePlugin: FastifyPluginAsync<GraaspPluginFileOptions> = async (fastify,
     },
   });
 
-  // if (!buildFilePath) {
-  //   throw new Error('graasp-plugin-file: buildFilePath is not defined');
-  // }
+  // "install" custom schema for validating document items update
+  extendExtrasUpdateSchema(updateSchema(fileService.type));
```
morganridel, line 49: Does this do anything? I only find lambdas returning schemas (`updateOne` in items fluent schema)
What does this do written alone like this without being assigned somewhere?

src/services/item/plugins/file/test/index.test.ts
```diff
@@ -442,6 +442,35 @@ describe('File Item routes tests', () => {
...
+      expect(response.json()).toMatchObject(new MemberCannotAccess(expect.anything()));
```
morganridel, line 461: Shouldn't this test if the the extra alt text has been modified?

src/services/item/plugins/file/index.ts
pyphilia: It's a dirty legacy trick: `extendExtrasUpdateSchema` as a local yet saved schema that is extended given new schemas. 

Later in `PATCH /items/id` it reuses this saved schema that got updated during the app setup (that goes through all the plugins). It allows to edit specific fields that only exists depending on the item type.

morganridel (approved):

PR #424, opened 2023-06-23
feat: add aggregation endpoint

rayacers:


src/services/action/utils/actions.ts
```diff
@@ -6,3 +6,43 @@ export const getGeolocationIp = (ip: string | number): geoip.Lookup | null => ge
...
+export type AggregateAttribute =
+  | 'user'
+  | 'createdDay'
+  | 'createdTimeOfDay'
+  | 'createdDayOfWeek'
+  | 'actionType'
+  | 'actionLocation'
+  | 'itemId'
+  | 'actionCount';
+export type AggregateFunctionType = 'AVG' | 'COUNT' | 'SUM';
```
spaenleh, line 19: Here you use uppercase literal strings but in the schema you use lowercase.

src/services/action/utils/actions.ts
```diff
@@ -6,3 +6,43 @@ export const getGeolocationIp = (ip: string | number): geoip.Lookup | null => ge
...
+    case 'createdDay':
+      return "date_trunc('day', action.createdAt)";
+    case 'createdTimeOfDay':
+      return 'extract(hour from created_at)';
+    case 'createdDayOfWeek':
+      return 'extract(dow from created_at)';
+    default:
+      throw new Error(`${attribute} Attribute does not exist.`);
+  }
+};
```
spaenleh, line 40: I think you can replace this by an object to do your mapping:
```ts
const aggregateExpressionNames = {
  user: "action.member_id",
  actionType: "action.type",
  ...
}
const expression = aggregateExpressionName[attribute]
```

src/services/action/utils/actions.ts
```diff
@@ -6,3 +6,43 @@ export const getGeolocationIp = (ip: string | number): geoip.Lookup | null => ge
...
+
+export const buildAggregateExpression = (
+  subqueryName: string,
+  func?: AggregateFunctionType,
+  metric?: AggregateAttribute,
+): string => {
+  return func + '(' + subqueryName + '."' + metric + '")';
```
spaenleh, line 47:
```suggestion
  return `${func}(${subqueryName}."${metric}")`;
```

src/services/item/plugins/action/service.ts
```diff
@@ -54,11 +56,18 @@ export class ActionItemService {
...
   ): Promise<Action[]> {
     const { view = Context.Builder, sampleSize = DEFAULT_ACTIONS_SAMPLE_SIZE } = filters;
+
     // get item
     const item = await this.itemService.get(actor, repositories, itemId);
 
-    // check member has admin membership over item
-    await validatePermission(repositories, PermissionLevel.Admin, actor, item);
+    // check permission
+    const permission = actor
```
pyphilia, line 64: This does not make sense.

src/services/item/plugins/action/service.ts
```diff
@@ -54,11 +56,18 @@ export class ActionItemService {
     filters: { view?: Context; sampleSize?: number },
   ): Promise<Action[]> {
     const { view = Context.Builder, sampleSize = DEFAULT_ACTIONS_SAMPLE_SIZE } = filters;
+
     // get item
     const item = await this.itemService.get(actor, repositories, itemId);
```
pyphilia, line 61: Information: This line automatically check the actor has access to the item. It means that it will return if the actor is undefined and the item is public.

src/services/item/plugins/action/service.ts
```diff
@@ -76,9 +85,44 @@ export class ActionItemService {
     return repositories.actionRepository.getForItem(item.path, {
       sampleSize: size,
       view,
+      memberId: permission === PermissionLevel.Admin ? undefined : actor.id,
```
pyphilia, line 88: I think there's a major problem here because you cannot handle `public` data (empty user both means public and admin).

pyphilia: We didn't define the public case together, so either:
- if you don't have time: add `preHandler: fastify.verifyAuthentication` to your endpoints (you can have a look to any endpoint in the codebase). It will ensure `actor` is always defined. But you will need to add the following code in your service so actor has the correct type.
```
 if (!actor) {
      throw new UnauthorizedMember(actor);
    }
```
- you have time: I think that a `public` item should be able to display the aggregated data only. And it shouldn't be possible to get individual actions only if you're signed in.

src/services/item/plugins/action/service.ts
```diff
@@ -54,11 +56,18 @@ export class ActionItemService {
...
+
     // get item
     const item = await this.itemService.get(actor, repositories, itemId);
 
-    // check member has admin membership over item
-    await validatePermission(repositories, PermissionLevel.Admin, actor, item);
+    // check permission
+    const permission = actor
+      ? (await repositories.itemMembershipRepository.getInherited(item, actor, true))?.permission
+      : null;
```
pyphilia, line 66: If you check `actor` beforehand, you don't need to do a "complex" conditional, I think it would be clearer.

src/services/item/plugins/action/service.ts
```diff
@@ -88,20 +132,34 @@ export class ActionItemService {
...
 
-    // get item and check rights
+    // get item
     const item = await this.itemService.get(actor, repositories, payload.itemId);
-    await validatePermission(repositories, PermissionLevel.Admin, actor, item);
 
-    // check membership and get actions
-    const actions = await repositories.actionRepository.getForItem(item.path);
+    // check permission
+    const permission = actor
```
pyphilia, line 139: This does not make sense.

src/services/item/plugins/action/service.ts
```diff
@@ -88,20 +132,34 @@ export class ActionItemService {
...
+      ? (await repositories.itemMembershipRepository.getInherited(item, actor, true))?.permission
+      : null;
+
+    let actions;
+    if (!permission || !actor) {
```
pyphilia, line 144: `!actor` is already checked above.

src/services/item/plugins/action/service.ts
```diff
@@ -88,20 +132,34 @@ export class ActionItemService {
...
+    // get item
     const item = await this.itemService.get(actor, repositories, payload.itemId);
-    await validatePermission(repositories, PermissionLevel.Admin, actor, item);
 
-    // check membership and get actions
-    const actions = await repositories.actionRepository.getForItem(item.path);
+    // check permission
+    const permission = actor
+      ? (await repositories.itemMembershipRepository.getInherited(item, actor, true))?.permission
+      : null;
```
pyphilia, line 141: `actor` is already checked above, so you don't need the conditional.

src/services/item/plugins/action/test/fixtures/actions.ts
```diff
@@ -0,0 +1,88 @@
...
+    item: item,
+    extra: {},
+  } as unknown as Action;
+};
+
+const buildActions = (item: Item, member: Member[]) => {
+  const actions: Action[] = [];
+  let dummyAction;
+
+  dummyAction = getDummyAction(
```
pyphilia, line 35: Why not just have a big array definition instead of pushing into it?

src/services/item/plugins/action/utils.ts
```diff
@@ -8,6 +12,44 @@ export enum ItemActionType {
...
 
+export const validateAggregateRequest = (
+  countGroupBy: AggregateAttribute[],
+  aggregateFuction: AggregateFunctionType,
+  aggregateMetric: AggregateAttribute,
+  aggregateBy: AggregateAttribute[],
+) => {
+  // Aggregate by user is not allowed
+  if (aggregateBy?.includes('user')) {
+    return false;
```
pyphilia, line 23: Throw an error, it will be clearer.

src/services/item/plugins/action/utils.ts
```diff
@@ -8,6 +12,44 @@ export enum ItemActionType {
...
+  }
+
+  // Perform aggregation on a grouping expression is not allowed
+  if (aggregateBy?.includes(aggregateMetric)) {
+    return false;
```
pyphilia, line 28: Throw an error, it will be clearer.

src/services/item/plugins/action/utils.ts
```diff
@@ -8,6 +12,44 @@ export enum ItemActionType {
...
+  }
+
+  // The input of the second stage aggregation should be the output of the first stage aggregation
+  if (!(countGroupBy.includes(aggregateMetric) || aggregateMetric === 'actionCount')) {
+    return false;
```
pyphilia, line 33: Throw an error.

src/services/item/plugins/action/utils.ts
```diff
@@ -8,6 +12,44 @@ export enum ItemActionType {
...
+  }
+  let falseFlag = false;
+  aggregateBy?.forEach((element) => {
+    if (!(countGroupBy.includes(element) || element === 'actionCount')) {
+      falseFlag = true;
```
pyphilia, line 38: Throw an error, then you won't even need to use `falseFlag`.
I personally prefer when validate throw rather than returning a boolean. If you throw, the endpoint should automatically throw as well (don't need to manually use reply). You can get inspiration from `action/utils/errors.ts` file.

pyphilia (changes requested):
Thanks for the PR, I have a concern about public cases, that should at least be forbidden if you don't have time to deal with it.

src/services/action/repositories/action.ts
```diff
@@ -51,6 +57,62 @@ export const ActionRepository = AppDataSource.getRepository(Action).extend({
...
+
+    // Second stage aggregation.
+    const query = AppDataSource.createQueryBuilder()
+      .addSelect(
+        buildAggregateExpression('subquery', aggregateFunction, aggregateMetric),
+        'aggregateResult',
+      )
+      .from('(' + subquery.getQuery() + ')', 'subquery')
+      .setParameter('path', itemPath)
+      .setParameter('view', view);
```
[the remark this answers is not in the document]
rayacers: I tried it before, but it seems that this approach is not allowed in a subquery.

src/services/action/repositories/action.ts
```diff
@@ -51,6 +57,62 @@ export const ActionRepository = AppDataSource.getRepository(Action).extend({
...
+    if (type) {
+      query.setParameter('type', type);
+    }
```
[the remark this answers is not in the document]
rayacers: ditto.

src/services/action/repositories/action.ts
```diff
@@ -51,6 +57,62 @@ export const ActionRepository = AppDataSource.getRepository(Action).extend({
...
+
+    aggregateBy?.forEach((attribute) => {
+      const expression = 'subquery.' + '"' + attribute + '"';
```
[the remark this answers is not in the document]
rayacers: Yes, it is necessary.

src/services/action/utils/actions.ts
```diff
@@ -6,3 +6,32 @@ export const getGeolocationIp = (ip: string | number): geoip.Lookup | null => ge
...
+  createdTimeOfDay: 'extract(hour from created_at)',
+  createdDayOfWeek: 'extract(dow from created_at)',
+};
+
+export const buildAggregateExpression = (
+  subqueryName: string,
+  func?: AggregateFunctionType,
+  metric?: AggregateAttribute,
+): string => {
+  return `${func}(${subqueryName}."${metric}")`;
```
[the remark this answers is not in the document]
rayacers: Yes, the query failed if I don't add the quotes.

src/services/item/plugins/action/service.ts
rayacers: I try to fix it and make it clear. For the individual actions, the workflow is in the following:

1. Check if the user is logged in
2. Check the right through `this.itemService.get(..., PermissionLevel.Read,)`. If it returns, the user has at least read access to the item or the item is public.
3.  Get the user's permission through `repositories.itemMembershipRepository.getInherited(...)`.
4.  If the permission is Admin, get all actions. Otherwise, get actions created by the actor.

In this case, for a public item, unauthorized members cannot access it, and authorized members can get the actions created by themselves even if they don't have any permission. Does that make sense?

For the aggregated data, we only check the right through `this.itemService.get(..., PermissionLevel.Read,)`. If it returns, we return the aggregated data.

pyphilia:
ref https://github.com/graasp/graasp/pull/514

PR #427, opened 2023-06-23
fix: getItemsByCategories only return root items

spaenleh:
This PR fixes ~~a query param typo.~~ the getItemsByCategories for published items to only return root items.

### DIFF
--- a/src/services/item/plugins/published/index.ts
+++ b/src/services/item/plugins/published/index.ts
@@ -22,14 +22,14 @@ const plugin: FastifyPluginAsync = async (fastify) => {
 
   fastify.register(graaspSearchPlugin);
 
-  fastify.get<{ Querystring: { categoryIds: string[] } }>(
+  fastify.get<{ Querystring: { categoryId: string[] } }>(
     '/collections',
     {
       schema: getCollections,
       preHandler: fastify.fetchMemberInSession,
     },
     async ({ query, member }) => {
-      return pIS.getItemsByCategories(member, buildRepositories(), query.categoryIds);
+      return pIS.getItemsByCategories(member, buildRepositories(), query.categoryId);
     },
   );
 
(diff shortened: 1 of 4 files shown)


pyphilia:
It's not a typo, if you use `categoryId` it allows you to have the following url: `endpoint?categoryId=x&categoryId=y` while having `categoryIds` leads to `endpoint?categoryIds=x&categoryIds=y`, or are you using another syntax?

For unpublished children: Are you sure? I think it uses the published table to filter out, but let me know.

spaenleh:
@pyphilia According to https://github.com/graasp/graasp-query-client/blob/4573b8fca8470e38de964c23624c7aec31e81fbc/src/api/routes.ts#L340 the call is made with this url: `endpoint?categoryIds=x&categoryIds=y`. So there is a mismatch at the endoint in the backend. It does not work en using the singular because the parameters are not singular.

spaenleh:
@pyphilia I think the children are also selected because of the use the `@>` operator. So we are selecting all items that are children from the published item. 

I found out that the filtering works as expected as is, but the ids of the categories need to be supplied in the `["A1,A2", "B2"] format in order to be selected as (A1 OR B2) AND (B2)

pyphilia:
@spaenleh I see, but maybe we prefer to change query-client rather than the backend? Everything in the backend uses singular querystrings.

I think I see what you mean now. But it `innerJoin` on `public` on equal paths so only the root is selected at this stage.

spaenleh:
Okay, I assume we could change `query-client`. But I don't know how to communicate that the `categoryId` param can be a join of multiple categories to allow for OR selection. We would still have to change some places in the backend where the plural is used for `categoryIds`, no ?

https://github.com/graasp/graasp/blob/d2e0574bad289411c71f1d7a2191a40b932c4ced/src/services/item/plugins/published/repositories/itemPublished.ts#L96 looks like it joins on all items with the `@>` no ? so not only roots

pyphilia (approved):

PR #457, opened 2023-06-26
fix: return creator for published items

spaenleh:
This PR adds the creator field to the returned published items.

~~Maybe we should filter the properties of the member that are returned ? But this would mean we have to create a new `PublicMember` type that only contains information that is ok to share. I do not know how we would handle that with relation to the type that is defined on the relation ... Maybe there is a way to filter the data directly using typeorm ? Like only exporting certain allowed columns ?~~
Update: The creator properties are handled by the schema, only "id", "name" and "email" are allowed, so we are fine. 😄 

This change is needed for the user avatars to work in the library.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -95,7 +95,7 @@ services:
     image: localstack/localstack
     volumes:
       - ${TMPDIR}/localstack:/tmp/graasp-localstack
-      - ./localstack:/docker-entrypoint-initaws.d
+      - './localstack/init.sh:/etc/localstack/init/ready.d/init-aws.sh'
       - /var/run/docker.sock:/var/run/docker.sock
     environment:
       - SERVICES=s3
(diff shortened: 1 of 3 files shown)


morganridel:
I don't think the "public" data should be handled at the DB level, the DB returns what's stored. For your usecase and with our current architecture, you can maybe create an alternative Item/Member schema, with a reduced scope. With `fluent-json-schema` we can even use [only](https://github.com/fastify/fluent-json-schema/blob/master/docs/API.md#only) to avoid too much dupplication.

This is if you want to have less information in the returned item that we currently have in the `graasp/src/services/item/fluent-schema.ts` `partialMember`

pyphilia (approved):
For me the changes are correct, you can "control" the output with the schemas. Currently it returns `partialMember` defined in the schema definition of `item` (https://github.com/graasp/graasp/blob/main/src/services/item/fluent-schema.ts)

spaenleh:
@morganridel You are right, I checked and only `id, name, email` are transmitted. So there is no more issue about leaking information. Thank you for pointing out the solution using `fluent-schema` but there is no needed change.

PR #465, opened 2023-06-28
fix: add member back in itemLike response

spaenleh:
Should we add the member back in the response of item like ?

Do we consider members as "public" ? 

I think there could be some privacy implications by returning the email. Maybe we could just return `id` and `name` ? 

Currently only users with read access to the item can use the `getItemLikeForItem`.
We should maybe add a publicly accessible call to query just the number of likes of an item. It should also check that that item is published.


pyphilia:
Is it still up to date?  @spaenleh 
There're currently a `items/liked` and `items/<itemId>/liked` endpoints.
The first one returns your own likes so you don't need to return the member (as you did in this PR)
The second endpoint returns the likes for an item. It is public. You can get the count from there. This indeed does not return any member (by design for privacy).

Let me know if I missed something.

spaenleh:
Okay, yes it makes sens, I will close this and we should change the style in sdk to match what the api is sending.

Thank you for having given this a look !

PR #466, opened 2023-06-28
fix: return root published item when querying for a child item

spaenleh:
This PR updates the publish information call to return the root item when querying for a child item that is published.
- single call
  - [x] repository 
  - [x] tests
- multiple call
  - [x] repository
  - [x] tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,6 @@ types
 
 # Secrets
 secret-key
+
+# caches
+.jest-cache
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,3 @@ types
 
 # Secrets
 secret-key
-
-# caches
-.jest-cache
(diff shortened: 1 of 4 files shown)


pyphilia (approved):
LGTM ☔

.gitignore
```diff
@@ -53,4 +53,7 @@ types
 .jest-cache
 
 # Secrets
-secret-key
\ No newline at end of file
+secret-key
+
+# caches
+.jest-cache
```
morganridel, line 59: It is already ignored

morganridel (changes requested):

.gitignore
spaenleh: Ah sorry, my branch was not up-to-date, will remove.

morganridel (approved):

PR #470, opened 2023-06-29
chore: make CI tests run < 5 mins

morganridel:
image
🙏

Note: I observed a difference of up to 5 min (total between 15-20min) in the tests execution on the CI before this PR, so the time displayed here are not really accurate all the time.

# Leverage jest cache

from [Jest's doc](https://jestjs.io/docs/cli#--cache):

> On average, disabling the cache makes Jest at least two times slower.

We can store the jest cache in GitHub, like the node_modules.

Example total CI run before the change: **16m40**
Example total CI run after the change: **13m40**

# Leverage all github CPU

By default, jest will only use one CPU on the GitHub Runner but we have 2 available. Jest will run multiple test suites in parallel on each allowed worker.
However most of our tests are dependent on a database, so this can cause problems as tests can interfere with each other while using the test database. We need one database per worker.

We never had problems because this is not visible with only one worker used on the CI.

So, we configure the action to use 2 workers and add a new test database.
Example total CI runs after the change: **10m17**, **13m53**

# Leverage matrix jobs

Jest can divide the tests in multiples shards. We leverage this feature to run multiple github jobs in parallel.

With 4 shards, we see significant speed up to the total workflow time.

Note: we effectively have 8 database containers running with this.

Example total CI runs after the change: **6m39** (< 5m tests)

# Conclusion

There is a small overhead with creating more github jobs, so the total number of minute used is *a little bit* higher (we mitigate this with the new cache and use of all of the CPUs) but I think the gain in feedback for devs is worth it.

The next steps would be to optimize the tests themselves, to identify where we lose most of the time. I might take a look into it at some points.


src/plugins/__mocks__/datasource.ts
```diff
@@ -35,7 +35,7 @@ import { Member } from '../../services/member/entities/member';
 export const AppDataSource = new DataSource({
   type: 'postgres',
   host: process.env.DB_HOST,
-  port: 5432,
+  port: process.env.CI === 'true' ? 5432 + parseInt(process.env.JEST_WORKER_ID ?? '1') - 1 : 5432,
```
morganridel, line 38: Any idea of a better looking way?

dialexo (approved):
LGTM 👏  Incredible work!

spaenleh:
🎆

PR #471, opened 2023-06-29
chore: cleanup env vars, add mailcatcher

dialexo:
This PR:

- cleans up some unused environment variables
- binds the hostname to 0.0.0.0 by default, renames `EMAIL_LINKS_HOST` to `PUBLIC_URL` and sets defaults in the docker-compose
- allows multiple read replica and renames `DB_READ_REPLICA_HOST` to `DB_READ_REPLICA_HOSTS`
- ~~moves the DB constants to `config.ts`~~
- allows dotenv to override in the env from the .env files (i.e the values in .env.xyz will now take precedence over the docker-compose now) which allows editing core config such as hostname without restarting the container or passing them through CLI
- sorts the github workflows (**WARNING: before merging, ensure that the env vars that have changed name such as `DB_READ_REPLICA_HOSTS` and `PUBLIC_URL` have been changed through the entire deployment workflows chain i.e. graasp-deploy also**)
- adds the mailcatcher mock mailbox container, which UI is available at localhost:1080. The default mail config has been changed in the README to point to it, and setting random config values will yield the current default behaviour (logging in the console)
- sets sane values and instructions to correctly configure one's `.dev.development` in the README

There is a bug on macOS where the devcontainer does not correctly forward the port of the server when the host name passed to fastify is not set to `0.0.0.0`. Using `localhost` will result in `ERR_EMPTY_RESPONSE` in Chrome. Note: this also happens on Linux 🤔 

TODO:
- [x] ~~check where the variable is used and replace it with some other PUBLIC_URL var if needed, e.g. when generating login links~~ the env var is `EMAIL_LINKS_HOST`


morganridel (approved):

dialexo:
> https://github.com/graasp/graasp/actions/runs/5488896562/jobs/10002596471?pr=471

Mmmh by moving the DB env vars to `config.ts`, all the other vars are checked thus they are not set when running the migration steps of the workflow. This is not ideal, so maybe I should revert to putting `process.env.DB_XYZ` in both `datasource.ts` files 🫤

dialexo:
> https://github.com/graasp/graasp/actions/runs/5489086753/jobs/10002938587

No idea what's happening here, I haven't changed existing stuff :thinking:

morganridel:
> > https://github.com/graasp/graasp/actions/runs/5489086753/jobs/10002938587
> 
> No idea what's happening here, I haven't changed existing stuff thinking

The failed test is linked to the mailer file so something must be up (a missing await or misconfiguration of something)

spaenleh:
The issue of the `HOSTNAME` on mac using Docker Desktop Version 4.19.0 (106363) and VSCode Version: 1.80.1 (Universal) and node v16.20.0 seems to have disappeared.

I can simply run `yarn watch` and `localhost:3000` will be accessible from the outside no problems.
@morganridel @dialexo

src/services/auth/utils.ts
```diff
@@ -1,10 +1,17 @@
...
 
 export const getRedirectionUrl = (url?: string) => {
   // TODO: improve check
   // TODO: whitelist graasp domains for redirection
   if (!url || !CLIENT_HOSTS.some(({ hostname }) => url.startsWith(hostname))) {
-    return CLIENT_HOST;
+    return defaultClientHost.hostname;
   }
   return url;
 };
```
dialexo, line 17: Okay I found the issue. The tests had some default behaviour when `CLIENT_HOST` was undefined (as it's the case in the CI) so now that I've removed `CLIENT_HOST` and always provide `BUILDER_CLIENT_HOST` instead, the auth tests should always receive redirection on success. On failure, it depends on `AUTH_CLIENT_HOST`

PR #483, opened 2023-07-13
fix: item relation can be null because of soft-delete

spaenleh:
This PR fixes an issue where if a published item has been put in the trash, the call to `all collections` will fail with a 500 Internal Error. 
This is caused by our use of the soft-delete on items. A soft-deleted item still exists in the `item` table but will be replaced by `null` when returned by typeorm. We did not find a way to make the soft-deletion cascade to entities that reference an item. These entities need to be **very** careful when querying a related item, as it may not be available due to soft-deletion. In that case we would like to not return the entity instance as they are light wrappers around the item, no item means we most likely do not care about the entity. 

The implementation forces us to use `createQueryBuilder` instead of typeorm's `find` because find uses a `leftJoinAndSelect` on relations which allows to return entities that have a null relation which we do not want. We use the `innerJoinAndSelect` to remove those soft-deleted items instead.

:warning: **This affects all entities that have a relation to item and expose an endpoint where we would like to have the `relations: { item: true }`**


src/services/item/plugins/app/appAction/repository.ts
```diff
@@ -48,6 +48,13 @@ export const AppActionRepository = AppDataSource.getRepository(AppAction).extend
       where: { item: { id: In(itemIds) }, member: { id: memberId } },
       relations: { item: true },
     });
+    // todo: should use something like:
+    // but this does not work. Maybe related to the placement of the IN ?
+    // const appActions = await this.createQueryBuilder('actions')
+    //   .innerJoinAndSelect('actions.item', 'item', 'item.id IN (:...itemIds)', { itemIds })
+    //   .where('actions.member = :memberId', { memberId })
+    //   .andWhere('actions.item IN (:...itemIds)', { itemIds })
+    //   .getMany();
```
morganridel, line 57: I think it's missing a join for the member? Also from the entity definition, both member and item can actually be null.
Which means that we can't actually differentiate between an action that is not linked to an Item and an action linked to a soft deleted item...

Maybe in this case we can use `withDeleted: true` in the find options? Maybe we still want the item info when we look at the actions?

But if not it's a tricky. We don't want a inner join because we will lose the actions which are not linked to a member or a item.

pyphilia (approved):
LGTM
Very tricky this soft-delete feature...

src/services/item/plugins/app/appAction/repository.ts
spaenleh: Good catch about the missing join on member ! But I am not sure it is allowed for the item or member to be null ? `nullable: false` means it can't be null right ? And we have a cascade on delete, so my guess would be that we may not want to retrieve appactions that belong to trashed items ... maybe @pyphilia can help.

morganridel (approved):

PR #484, opened 2023-07-17
fix: fix detachedMoveHousekeeping to move to root

### LINKED ISSUE #476: Moving nested item to root make it "disapear"
## STR:
* Create an item inside a folder
* Move the item to root "Owned Items"
* The item is not visible anymore

## Possible explanation:
A nested item do not necesarily have a direct membership (it inherits some), By moving it to the root, it has no assigned membership anymore and isn't returned by the `getOwn` request.

Should a membership be created in the "housekeeping" during a move? I'm not really familiar with the complex process inside the housekeeping. What if it is moved by someone other than the creator? What if the creator is not admin anymore on the item?

pyphilia:
Actually the `detachedMoveHousekeeping` was not correctly refactored, so it never returned the necessary memberships to add. It should be fixed now.

fix #476


src/services/item/test/index.test.ts
```diff
@@ -1966,6 +1967,43 @@ describe('Item routes tests', () => {
...
+
+        const response = await app.inject({
+          method: HttpMethod.POST,
+          url: `/items/move?${qs.stringify({ id: item.id }, { arrayFormat: 'repeat' })}`,
+          payload: {},
+        });
+
+        expect(response.statusCode).toBe(StatusCodes.ACCEPTED);
+
+        // item should have a differnt path
```
morganridel, line 1985:
```suggestion
        // item should have a different path
```

morganridel (approved):
LGTM

All the `setTimeout((), MULTIPLE_ITEMS_LOADING_TIME)` tests make me a bit uneasy though, this make the test flaky andadd 2 seconds to every test. in the future, We could consider using `waitForExpect` (example in h5p test I think) to improve the odds of success and check earlier than 2 seconds. Or better, having a callback when moved is finished, but it's a bit more involved

PR #486, opened 2023-07-17
fix: login-password to use a redirect with the token in query string

spaenleh:
This PR updates the `/m/login-password` endpoint to use a redirect and send the token as a query-param. 
This is to support the workflow of the mobile login from the webauth.

The deep-link uses the protocol set in the `GRAASP_MOBILE_BUILDER` env variable which defaults to `graasp-mobile-builder`.

We could mirror what is happenning with the web endpoints for the redirect url which can be set as a query param in the request, and the backend uses it if it is in the allowed whitelist.

### DIFF
--- a/src/services/auth/plugins/mobile/mobile.test.ts
+++ b/src/services/auth/plugins/mobile/mobile.test.ts
@@ -221,8 +221,10 @@ describe('Mobile Endpoints', () => {
           captcha: MOCK_CAPTCHA,
         },
       });
-      expect(response.statusCode).toEqual(StatusCodes.MOVED_TEMPORARILY);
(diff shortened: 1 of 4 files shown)


src/services/auth/plugins/mobile/mobile.test.ts
```diff
@@ -217,8 +217,8 @@ describe('Mobile Endpoints', () => {
           captcha: MOCK_CAPTCHA,
         },
       });
-      expect(response.statusCode).toEqual(StatusCodes.OK);
-      expect(response.json()).toHaveProperty('t');
+      expect(response.statusCode).toEqual(StatusCodes.MOVED_TEMPORARILY);
+      expect(response.headers.location).toContain('graasp-mobile-builder://auth?t=');
```
pyphilia, line 221: Use env var?

pyphilia (approved):
LGTM 👍

PR #488, opened 2023-07-18
fix: add migration to clean tags

pyphilia:
close #466

### DIFF
--- a/test/migrations/migrations1689666251815/fixture.ts
+++ b/test/migrations/migrations1689666251815/fixture.ts
@@ -1,0 +1,66 @@
+const memberId = '3e901df0-d246-4672-bb01-34269f4c0fed';
+const itemId = '1e901df0-d246-4672-bb01-34269f4c0fed';
+const incorrectItemId = '2e901df0-d246-4672-bb01-34269f4c0fed';
+const itemPath = itemId.replace(/-/g, '_');
+
+const values = {
+  member: [
+    {
+      id: memberId,
+      name: 'my member',
+      email: 'email@email.com',
+      type: 'individual',
+      extra: { hasThumbnail: true, favoriteItems: [itemId, itemId, incorrectItemId] }, // migration should handle dupplicated favorites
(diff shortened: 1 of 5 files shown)


test/migrations/migrations1689666251815/fixture.ts
```diff
@@ -0,0 +1,66 @@
...
+const itemPath = itemId.replace(/-/g, '_');
+
+const values = {
+  member: [
+    {
+      id: memberId,
+      name: 'my member',
+      email: 'email@email.com',
+      type: 'individual',
+      extra: { hasThumbnail: true, favoriteItems: [itemId, itemId, incorrectItemId] }, // migration should handle dupplicated favorites
```
morganridel, line 13: I don't think this part is important for this migration :D

test/migrations/migrations1689666251815/migrations.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    const itemLoginTag = await app.db.query(
+      buildSelectQuery('item_tag', { id: migrationData.item_tag[2].id }),
+    );
+    expect(itemLoginTag).toHaveLength(0);
+
+    const [hiddenTag] = await app.db.query(
+      buildSelectQuery('item_tag', { id: migrationData.item_tag[3].id }),
+    );
+    expect(hiddenTag.type).toEqual(ItemTagType.Hidden);
+  });
```
morganridel, line 67: This is the kind of migration where I'm not sure of the benefits of testing migrations. If your test pass on the CI it will basically pass all the time, but it will be irrelevant as soon as the migration has passed on prod, because we won't rerun the migration ever. It's not like the test will start to fail at some point (I guess?)

morganridel (approved):
LGTM

test/migrations/migrations1689666251815/migrations.test.ts
pyphilia: That's true, maybe it ensures you don't modify the migrations? Tho it shouldn't affect prod in real life.

PR #492, opened 2023-07-19
fix: remove duplicate public tag in migration

pyphilia:
Adds a sql query to remove duplicate tags, it happens only in dev (and stage) since we re-added public to some of the items.


spaenleh (approved):
Looks good ! Thank you for this ! 🩹

PR #496, opened 2023-07-21
fix: improve import zip, fix undefined parent

### LINKED ISSUE #495: Import ZIP bug
For some zip the import does not work.

pyphilia:
The original error was induced because of an undefined parent. 

In the same time I've changed the following:
- simplified the process to fetch the description file content before saving it
- enable export of h5p files
- added simple tests for the import functionality (remove `utils.test.ts` since the function does not exist anymore)

fix #495


.github/workflows/deploy-development.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     # repository name
     name: Graasp
     # Reference reusable workflow file. Using the commit SHA is the safest for stability and security
-    uses: graasp/graasp-deploy/.github/workflows/cintegration-ecs-backend.yml@v1
+    uses: graasp/graasp-deploy/.github/workflows/cintegration-ecs-backend.yml@log-level
```
pyphilia, line 20: MERGE WORKFLOW

src/services/itemMembership/repository.ts
```diff
@@ -317,16 +315,15 @@ export const ItemMembershipRepository = AppDataSource.getRepository(ItemMembersh
 
       if (membershipsBelowToDiscard.length > 0) {
         // remove redundant existing memberships and to create the new one
-        tasks = membershipsBelowToDiscard.map(async (membership) => {
+        for (const membership of membershipsBelowToDiscard) {
           await this.delete(membership.id);
-        });
+        }
```
dialexo, line 320: This now runs sequentially. I think we can execute those in parallel.

PR #502, opened 2023-07-25
fix: add creator in recent and liked collections

spaenleh:
This PR adds the relation "creator" for the recent and liked collections. This way it can be displayed in the library.

### DIFF
--- a/src/services/item/plugins/published/repositories/itemPublished.ts
+++ b/src/services/item/plugins/published/repositories/itemPublished.ts
@@ -58,6 +58,7 @@ export const ItemPublishedRepository = AppDataSource.getRepository(ItemPublished
       .from(Item, 'item')
       .innerJoin('item_published', 'pi', 'pi.item_path = item.path')
       .innerJoin('item_membership', 'im', 'im.item_path @> item.path')
+      .innerJoinAndSelect('item.creator', 'member')
       .where('im.member_id = :memberId', { memberId })
       .andWhere('im.permission IN (:...permissions)', {
         permissions: [PermissionLevel.Admin, PermissionLevel.Write],
--- a/src/services/item/plugins/published/repositories/itemPublished.ts
+++ b/src/services/item/plugins/published/repositories/itemPublished.ts
@@ -81,6 +81,7 @@ export const ItemPublishedRepository = AppDataSource.getRepository(ItemPublished
   async getRecentItems(limit: number = 10): Promise<Item[]> {
     const publishedInfos = await this.createQueryBuilder('item_published')
       .innerJoinAndSelect('item_published.item', 'item')
+      .innerJoinAndSelect('item.creator', 'member')
       .orderBy('item.createdAt', 'DESC')
       .take(limit)
       .getMany();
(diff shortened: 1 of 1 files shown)


dialexo (commented):
Does this retrieve the entire member or only the id? Should we let library fetch twice, or provide the entire data upfront?

spaenleh:
@dialexo According to the schemas and the new types from typeorm, the member is returned with the following properties: 
```json
{
  id: string;
  name: string;
  email: string;
}
```
I think we want to return the member in the same call as the item, as this is what is expected by the library.

dialexo (approved):

src/services/item/plugins/published/repositories/itemPublished.ts
```diff
@@ -129,9 +131,10 @@ export const ItemPublishedRepository = AppDataSource.getRepository(ItemPublished
   // bug: does not take into account child items
   async getLikedItems(limit: number = 10): Promise<Item[]> {
     const itemPublished = await this.createQueryBuilder('item_published')
-      .leftJoinAndSelect('item_published.item', 'i')
-      .innerJoin('item_like', 'il', 'il.item_id = i.id')
-      .groupBy(['i.id', 'item_published.id'])
+      .innerJoinAndSelect('item_published.item', 'item')
+      .innerJoinAndSelect('item.creator', 'member')
+      .innerJoin('item_like', 'il', 'il.item_id = item.id')
+      .groupBy(['item.id', 'member.id', 'item_published.id'])
```
morganridel, line 137: Will this not separate the like count in multiple like count per user?
Like one count per unique value of (item.id, member.id). I guess it's okay because of having only one member per item (creator) in the query

morganridel (approved):
LGTM

src/migrations/1689777747530-default-item-settings.ts
```diff
@@ -1,14 +1,13 @@
...
 
-    public async up(queryRunner: QueryRunner): Promise<void> {
-        await queryRunner.query(`ALTER TABLE "item" ALTER COLUMN "settings" DROP DEFAULT`);
-    }
-
-    public async down(queryRunner: QueryRunner): Promise<void> {
-        await queryRunner.query(`ALTER TABLE "item" ALTER COLUMN "settings" SET DEFAULT '{}'`);
-    }
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "item" ALTER COLUMN "settings" DROP DEFAULT`);
```
morganridel, line 7: This should not be in your PR I think, it should already be done on the last migration on `main`, make sure to run existing migration before running generate

morganridel (changes requested):

morganridel (approved):

src/services/item/plugins/published/repositories/itemPublished.ts
spaenleh: Yes I guess you are right... I was not sure about how to handle the `groupBy`. I was getting an error about a non-aggregated column when omitting it.

PR #505, opened 2023-07-25
504 fix app data reader

### LINKED ISSUE #504: Fix GET app data for reader

pyphilia:
The every app data couldn't be fetch by the reader: we missed the `member` visibility app data.

- I've moved the permission logic to the repository. I've noticed `casl` abilities don't work so I've removed it from `appData` and `authorization` 
- For simplicity, and because we don't use it, I've removed the possibility to add filters for `GET /appData`
- `PATCH /appData` was also missing a throw check.  
- added member to `GET /appAction`

close #504

### DIFF
--- a/src/services/item/plugins/app/appData/test/index.test.ts
+++ b/src/services/item/plugins/app/appData/test/index.test.ts
@@ -197,11 +197,11 @@ describe('App Data Tests', () => {
     describe('Sign In as reader', () => {
       beforeEach(async () => {
         ({ app, actor } = await build());
-        const creator = await saveMember(BOB);
+        member = await saveMember(BOB);
         ({ item, appData, token } = await setUpForAppData(
           app,
           actor,
-          creator,
+          member,
           PermissionLevel.Read,
         ));
       });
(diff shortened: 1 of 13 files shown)


src/services/item/plugins/app/appData/test/index.test.ts
```diff
@@ -394,55 +426,74 @@ describe('Apps Data Tests', () => {
...
+      it('Invalid app data id throws bad request', async () => {
+        const response = await app.inject({
+          method: HttpMethod.PATCH,
+          url: `${APP_ITEMS_PREFIX}/${item.id}/app-data/invalid-id`,
+          payload: { data: updatedData.data },
+        });
+        expect(response.statusCode).toEqual(StatusCodes.BAD_REQUEST);
+      });
+
       // it('Throw if app data is a file', async () => {
```
pyphilia, line 446: I move it to another issue, this test should be enabled back.

morganridel (approved):
Not so familiar with appdata but LGTM,
In which case do we need the service's `validateAppDataPermission` if we can check permission in the repository?

pyphilia:
Actually I would rather say that the subset you can get (and that depends on your permission) was moved to the repository. 

`validateAppDataPermission` is still necessary to check your permission against a given appdata. 

It doesn't sounds 100% consistent but I don't have a better solution for now.

PR #507, opened 2023-07-25
fix: allow multiple extra and settings fields on update

### LINKED ISSUE #500: Allow `flavor` in document item update
https://github.com/graasp/graasp/blob/26e9357a7f7e1977edfe977db3cc56012d056ca0/src/services/item/plugins/document/schemas.ts#L14

Flavor is not allowed for update currently.

+ try to set `additionalProperties(false)` to trigger an error next time

dialexo:
Closes #500 

- Adds test to check if multiple extra and settings fields can be modified
- Fix logic that checks which extra and settings can be updated
- Adds yarn check (for prettier) in CI

### DIFF
--- a/src/migrations/1689777747530-default-item-settings.ts
+++ b/src/migrations/1689777747530-default-item-settings.ts
@@ -1,14 +1,13 @@
-import { MigrationInterface, QueryRunner } from "typeorm";
+import { MigrationInterface, QueryRunner } from 'typeorm';
 
 export class Migrations1689777747530 implements MigrationInterface {
-    name = 'Migrations1689777747530';
+  name = 'Migrations1689777747530';
 
-    public async up(queryRunner: QueryRunner): Promise<void> {
-        await queryRunner.query(`ALTER TABLE "item" ALTER COLUMN "settings" DROP DEFAULT`);
-    }
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "item" ALTER COLUMN "settings" DROP DEFAULT`);
+  }
 
-    public async down(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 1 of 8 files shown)


spaenleh (approved):
🐈 Looks good to me ! 🌟  This might even solve some bugs in the front-end where we only received a single setting in return ...

morganridel (approved):
LGTM

src/migrations/1689777747530-default-item-settings.ts
```diff
@@ -1,14 +1,13 @@
...
 
-    public async up(queryRunner: QueryRunner): Promise<void> {
-        await queryRunner.query(`ALTER TABLE "item" ALTER COLUMN "settings" DROP DEFAULT`);
-    }
-
-    public async down(queryRunner: QueryRunner): Promise<void> {
-        await queryRunner.query(`ALTER TABLE "item" ALTER COLUMN "settings" SET DEFAULT '{}'`);
-    }
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "item" ALTER COLUMN "settings" DROP DEFAULT`);
```
morganridel, line 7: This should not be in your PR I think, it should already be done on the last migration on main, make sure to run existing migration before running generate

morganridel (changes requested):

src/migrations/1689777747530-default-item-settings.ts
dialexo: @morganridel this is Prettier's work, because the files were not formatted properly in #494. By the way note that I've added `yarn check` in this PR to avoid this same problem in the future (the CI will fail if prettier was not applied):

https://github.com/graasp/graasp/blob/b004979795d4aa598e13f8704ccf898099fbdd27/.github/workflows/test.yml#L213-L215

morganridel: alright

morganridel (approved):

PR #508, opened 2023-07-26
chore: load git bash completions in devcontainer

dialexo:
Loads git bash completions for the node user in the devcontainer


spaenleh (approved):
Not sure why we do not need the temp storage ? But looks good ! Always like some good DX upgrade ! 💪

dialexo:
> Not sure why we do not need the temp storage ? But looks good ! Always like some good DX upgrade ! muscle

In the `docker-compose` file, we bind the local `./tmp` to the container's `/tmp/graasp-file-item-storage` folder.
https://github.com/graasp/graasp/blob/cde5967e82c51606efd3a01ee864e7343ade7e60/.devcontainer/docker-compose.yml#L24-L26
So Docker will already create the bind mount at that path

PR #509, opened 2023-07-26
chore: add indices on item membership columns

morganridel:
Adding index on columns where we might use `where` queries. Probably a deeper analysis could lead to useful multi-column indices, but let's not add them randomly without testing queries directly.

I did not add an index on `Item.path` because the column is already `unique` and this [generates an index automatically](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS)
> adding a unique constraint will automatically create a unique B-tree index on the column or group of columns listed in the constraint

I removed the "global" migration test which count the number of test because I think it's not pertinent, especially with a migration like mine. But I'm open to alternatives


src/services/itemMembership/entities/ItemMembership.ts
```diff
@@ -26,14 +27,17 @@ export class ItemMembership extends BaseEntity implements GraaspItemMembership {
...
   })
+  @Index()
   permission: PermissionLevel;
 
   @ManyToOne(() => Member, (member) => member.id, {
     onDelete: 'SET NULL',
   })
+  @Index()
   @JoinColumn({ name: 'creator_id' })
   creator: Member;
```
pyphilia, line 38: Not sure this one is necessary, I can't think of a case where we look for a membership given the creator.

pyphilia (approved):
LGTM 
We've also noticed Graasp Analytics was really slow, most probably by getting `GET /actions` probably we need an index on item path?

src/services/itemMembership/entities/ItemMembership.ts
morganridel: I was actually thinking of getOwn

```
async getOwn(memberId: string): Promise<Item[]> {
    return this.createQueryBuilder('item')
      .innerJoin('item_membership', 'im', 'im.item_path @> item.path')
      .where('creator.id = :id', { id: memberId })
      .andWhere('im.permission = :permission', { permission: PermissionLevel.Admin })
      .andWhere('nlevel(item.path) = 1')
      .orderBy('item.updatedAt', 'DESC')
      .getMany();
  },
```

But this is probably an index on item.creator then, you're right I will change it

PR #513, opened 2023-07-26
fix: return inherit item login schema, prevent put from child

### LINKED ISSUE #512: Cannot get inherited item login for child
Most probably 
```
 .where('item.path <@ :path', { path: itemPath })
```
is inverted

pyphilia:
- Return inherited item login schema from child
- Prevent put of inherited item login schema from child 

fix #512


dialexo (approved):
LGTM :flying_saucer:

PR #515, opened 2023-07-26
Broadcast apps data change events with websockets

swouf:
Use websockets to send events (post, patch, delete) on app data, actions and settings.

Refer to https://github.com/graasp/graasp-apps-query-client/pull/142 for details.

## Backend features

- [x] Implement plugin providing hooks for signaling updates with websockets.
- [x] Use WS plugin for signaling changes in app data
    - [x] 🧪 -> 📋 Test `post`
    - [x] 🧪 -> 📋 Test `patch`
    - [x] 🧪 -> 📋 Test `delete`
- [x] Use WS plugin for signaling changes in app settings
    - [x] 🧪 -> 📋 Test `post`
    - [x] 🧪 -> 📋 Test `patch`
    - [x] 🧪 -> 📋 Test `delete`
- [x] Use WS plugin for signaling changes in app actions
    - [x] 🧪 -> 📋 Test `post`

### DIFF
--- a/src/services/item/plugins/app/appAction/index.ts
+++ b/src/services/item/plugins/app/appAction/index.ts
@@ -21,7 +21,9 @@ const plugin: FastifyPluginAsync = async (fastify) => {
   fastify.register(async function (fastify) {
     fastify.addHook('preHandler', fastify.verifyBearerAuth as preHandlerHookHandler);
 
-    if (WEBSOCKETS_PLUGIN) fastify.register(appActionsWsHooks, { appActionService });
+    if (WEBSOCKETS_PLUGIN) {
+      fastify.register(appActionsWsHooks, { appActionService });
+    }
 
     // create app action
     fastify.post<{ Params: { itemId: string }; Body: Partial<InputAppAction> }>(
--- a/src/services/item/plugins/app/appAction/index.ts
+++ b/src/services/item/plugins/app/appAction/index.ts
@@ -1,8 +1,10 @@
 import { preHandlerHookHandler } from 'fastify';
 import { FastifyPluginAsync } from 'fastify';
 
+import { WEBSOCKETS_PLUGIN } from '../../../../../utils/config';
 import { buildRepositories } from '../../../../../utils/repositories';
 import { ManyItemsGetFilter, SingleItemGetFilter } from '../interfaces/request';
+import { appActionsWsHooks } from '../ws/hooks';
 import { InputAppAction } from './interfaces/app-action';
 import common, { create, getForMany, getForOne } from './schemas';
 import { AppActionService } from './service';
@@ -18,6 +20,8 @@ const plugin: FastifyPluginAsync = async (fastify) => {
   // endpoints accessible to third parties with Bearer token
   fastify.register(async function (fastify) {
     fastify.addHook('preHandler', fastify.verifyBearerAuth as preHandlerHookHandler);
+
+    if (WEBSOCKETS_PLUGIN) fastify.register(appActionsWsHooks, { appActionService });
 
     // create app action
     fastify.post<{ Params: { itemId: string }; Body: Partial<InputAppAction> }>(
--- a/src/services/item/plugins/app/appAction/index.ts
+++ b/src/services/item/plugins/app/appAction/index.ts
@@ -1,7 +1,6 @@
 import { preHandlerHookHandler } from 'fastify';
 import { FastifyPluginAsync } from 'fastify';
 
-import { WEBSOCKETS_PLUGIN } from '../../../../../utils/config';
 import { buildRepositories } from '../../../../../utils/repositories';
 import { ManyItemsGetFilter, SingleItemGetFilter } from '../interfaces/request';
 import { appActionsWsHooks } from '../ws/hooks';
@@ -21,9 +20,7 @@ const plugin: FastifyPluginAsync = async (fastify) => {
   fastify.register(async function (fastify) {
     fastify.addHook('preHandler', fastify.verifyBearerAuth as preHandlerHookHandler);
 
-    if (WEBSOCKETS_PLUGIN) {
-      fastify.register(appActionsWsHooks, { appActionService });
-    }
+    fastify.register(appActionsWsHooks, { appActionService });
 
     // create app action
     fastify.post<{ Params: { itemId: string }; Body: Partial<InputAppAction> }>(
--- a/src/services/item/plugins/app/appAction/service.ts
+++ b/src/services/item/plugins/app/appAction/service.ts
@@ -1,5 +1,6 @@
-import { PermissionLevel } from '@graasp/sdk';
+import { AppAction, PermissionLevel } from '@graasp/sdk';
 
+import HookManager from '../../../../../utils/hook';
 import { Repositories } from '../../../../../utils/repositories';
 import { validatePermission } from '../../../../authorization';
 import { ManyItemsGetFilter, SingleItemGetFilter } from '../interfaces/request';
@@ -7,6 +8,12 @@ import { AppActionNotAccessible } from './errors';
 import { InputAppAction } from './interfaces/app-action';
 
 export class AppActionService {
+  hooks = new HookManager<{
+    post: {
+      pre: { appAction: Partial<InputAppAction>; itemId: string };
+      post: { appAction: AppAction; itemId: string };
+    };
+  }>();
   async post(actorId, repositories: Repositories, itemId: string, body: Partial<InputAppAction>) {
     const { appActionRepository, memberRepository, itemRepository } = repositories;
     // TODO: check member exists
@@ -18,7 +25,15 @@ export class AppActionService {
     // posting an app data is allowed to readers
     await validatePermission(repositories, PermissionLevel.Read, member, item);
 
-    return appActionRepository.post(itemId, actorId, body);
+    await this.hooks.runPreHooks('post', member, repositories, { appAction: body, itemId });
+
+    return appActionRepository.post(itemId, actorId, body).then(async (newAppAction) => {
+      await this.hooks.runPostHooks('post', member, repositories, {
+        appAction: newAppAction,
+        itemId,
+      });
+      return newAppAction;
+    });
   }
 
   async getForItem(
--- a/src/services/item/plugins/app/appAction/service.ts
+++ b/src/services/item/plugins/app/appAction/service.ts
@@ -27,13 +27,12 @@ export class AppActionService {
 
     await this.hooks.runPreHooks('post', member, repositories, { appAction: body, itemId });
 
-    return appActionRepository.post(itemId, actorId, body).then(async (newAppAction) => {
-      await this.hooks.runPostHooks('post', member, repositories, {
-        appAction: newAppAction,
-        itemId,
-      });
-      return newAppAction;
+    const appAction = await appActionRepository.post(itemId, actorId, body);
+    await this.hooks.runPostHooks('post', member, repositories, {
+      appAction,
+      itemId,
     });
+    return appAction;
   }
 
   async getForItem(
--- a/src/services/item/plugins/app/appAction/service.ts
+++ b/src/services/item/plugins/app/appAction/service.ts
@@ -22,7 +22,7 @@ export class AppActionService {
     // check item exists? let post fail?
     const item = await itemRepository.get(itemId);
 
-    // posting an app data is allowed to readers
+    // posting an app action is allowed to readers
     await validatePermission(repositories, PermissionLevel.Read, member, item);
 
     await this.hooks.runPreHooks('post', member, repositories, { appAction: body, itemId });
@@ -81,7 +81,7 @@ export class AppActionService {
     // check item exists
     const item = await itemRepository.get(itemIds[0]);
 
-    // posting an app data is allowed to readers
+    // posting an app action is allowed to readers
     const membership = await validatePermission(repositories, PermissionLevel.Read, member, item);
     const permission = membership?.permission;
     let { memberId: fMemberId } = filters;
--- a/src/services/item/plugins/app/appData/index.ts
+++ b/src/services/item/plugins/app/appData/index.ts
@@ -3,7 +3,6 @@ import { FastifyPluginAsync } from 'fastify';
 
 import { IdParam } from '@graasp/sdk';
 
-import { WEBSOCKETS_PLUGIN } from '../../../../../utils/config';
 import { buildRepositories } from '../../../../../utils/repositories';
 import { ManyItemsGetFilter } from '../interfaces/request';
 import { appDataWsHooks } from '../ws/hooks';
@@ -27,9 +26,7 @@ const appDataPlugin: FastifyPluginAsync = async (fastify) => {
     // origins from the publishers table an build a rule with that.
 
     fastify.register(appDataFilePlugin, { appDataService });
-    if (WEBSOCKETS_PLUGIN) {
-      fastify.register(appDataWsHooks, { appDataService });
-    }
+    fastify.register(appDataWsHooks, { appDataService });
 
     // create app data
     fastify.post<{ Params: { itemId: string }; Body: Partial<InputAppData> }>(
--- a/src/services/item/plugins/app/appData/index.ts
+++ b/src/services/item/plugins/app/appData/index.ts
@@ -1,9 +1,12 @@
+import fastifyView from '@fastify/view';
 import { FastifyPluginAsync } from 'fastify';
 
 import { IdParam } from '@graasp/sdk';
 
+import { WEBSOCKETS_PLUGIN } from '../../../../../utils/config';
 import { buildRepositories } from '../../../../../utils/repositories';
 import { ManyItemsGetFilter } from '../interfaces/request';
+import { appDataWsHooks } from '../ws/hooks';
 import { AppData } from './appData';
 import { InputAppData } from './interfaces/app-data';
 import appDataFilePlugin from './plugins/file';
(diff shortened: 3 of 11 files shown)


src/services/item/plugins/app/appAction/index.ts
```diff
@@ -19,6 +21,8 @@ const plugin: FastifyPluginAsync = async (fastify) => {
   fastify.register(async function (fastify) {
     fastify.addHook('preHandler', fastify.verifyBearerAuth as preHandlerHookHandler);
 
+    if (WEBSOCKETS_PLUGIN) fastify.register(appActionsWsHooks, { appActionService });
```
pyphilia, line 24: Add `{ ... }`

src/services/item/plugins/app/appData/index.ts
```diff
@@ -24,6 +27,7 @@ const appDataPlugin: FastifyPluginAsync = async (fastify) => {
     // origins from the publishers table an build a rule with that.
 
     fastify.register(appDataFilePlugin, { appDataService });
+    if (WEBSOCKETS_PLUGIN) fastify.register(appDataWsHooks, { appDataService });
```
pyphilia, line 30: Add `{ ... }`

src/services/item/plugins/app/appData/service.ts
```diff
@@ -85,7 +101,10 @@ export class AppDataService {
       },
     );
 
-    return appDataRepository.post(itemId, actorId, completeData);
+    return appDataRepository.post(itemId, actorId, completeData).then(async (newAppData) => {
+      await this.hooks.runPostHooks('post', member, repositories, { appData: newAppData, itemId });
+      return newAppData;
+    });
```
pyphilia, line 107: Written like this we expect to not wait for the posthook. I know we don't need to wait for the websockets, but I would add the `await` to be consistent with other parts of the code. What do you think @dialexo ?

src/services/item/plugins/app/appSetting/index.ts
```diff
@@ -22,6 +24,8 @@ const plugin: FastifyPluginAsync = async (fastify) => {
 
   const appSettingService = new AppSettingService(itemService);
 
+  if (WEBSOCKETS_PLUGIN) fastify.register(appSettingsWsHooks, { appSettingService });
```
pyphilia, line 27: Add `{...}`

src/services/item/plugins/app/appSetting/service.ts
```diff
@@ -53,11 +78,22 @@ export class AppSettingService {
     // patching requires admin rights
     await this.itemService.get(member, repositories, itemId, PermissionLevel.Admin);
 
-    const appSetting = await appSettingRepository.get(appSettingId);
-
-    await this.hooks.runPreHooks('patch', member, repositories, appSetting);
-
-    return appSettingRepository.patch(itemId, appSettingId, body);
+    // const appSetting = await appSettingRepository.get(appSettingId);
```
pyphilia, line 81: Remove.

src/services/item/plugins/app/ws/events.ts
```diff
@@ -0,0 +1,86 @@
...
+import { AppAction, AppSetting } from '@graasp/sdk';
+
+import { AppData } from '../appData/appData';
+
+// changes on app entities
+export const appDataTopic = 'app-data';
+export const appActionsTopic = 'app-actions';
+export const appSettingsTopic = 'app-settings';
+// changes on items of given user
+// export const memberItemsTopic = 'item/member';
```
pyphilia, line 13: Remove.

src/services/item/plugins/app/ws/hooks.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  appDataService.hooks.setPostHook('delete', async (member, repositories, { appData, itemId }) => {
+    if (itemId !== undefined && appData.visibility === AppDataVisibility.Item) {
+      websockets.publish(appDataTopic, itemId, AppDataEvent('delete', appData));
+    }
+  });
+}
+
+/**
+ * helper to register app action topic
```
pyphilia, line 60: I think it would make sense to divide this file in 3 parts.

src/services/item/plugins/app/ws/hooks.ts
```diff
@@ -0,0 +1,181 @@
...
+  itemService: ItemService,
+) {
+  websockets.register(appSettingsTopic, async (req) => {
+    const { channel: id, member } = req;
+    const repositories = buildRepositories();
+    const item = await itemService.get(member, repositories, id);
+    await validatePermission(repositories, PermissionLevel.Read, member, item);
+    if (item.type !== ItemType.APP) {
+      throw new Websocket.AccessDeniedError('item is not app');
+    }
```
pyphilia, line 103: Should you factor out this check?

pyphilia (approved):
LGTM but I let alex have the final review, gg!

src/services/item/plugins/app/ws/events.ts
```diff
@@ -0,0 +1,86 @@
...
+interface AppEvent {
+  kind: string;
+  op: AppOperations;
+}
+
+/**
+ * Events that affect an app data
+ */
+interface AppDataEvent extends AppEvent {
+  kind: 'app-data';
```
spaenleh, line 29: Use `appDataTopic` ? Same for the others.

spaenleh (commented):
Agree with the other comments, and a small suggestion.

src/services/item/plugins/app/appAction/index.ts
swouf: Sorry, I don't get it? Can you give be an example?

src/services/item/plugins/app/appData/index.ts
swouf: Sorry, I don't get it? Can you give be an example?

src/services/item/plugins/app/appSetting/index.ts
swouf: Same.

src/services/item/plugins/app/appData/index.ts
spaenleh:
```suggestion
if (WEBSOCKETS_PLUGIN) {
  fastify.register(appDataWsHooks, { appDataService });
}
```

src/services/item/plugins/app/appData/service.ts
```diff
@@ -127,9 +151,13 @@ export class AppDataService {
       throw new PreventUpdateOtherAppData(appDataId);
     }
 
-    await this.hooks.runPreHooks('patch', member, repositories, appData);
```
dialexo, line 130: What is the reason for moving the pre hook before the above checks?

src/services/item/plugins/app/appData/service.ts
```diff
@@ -127,9 +151,13 @@ export class AppDataService {
...
-
-    return appDataRepository.patch(itemId, appDataId, body);
+    return appDataRepository.patch(itemId, appDataId, body).then(async (patchedAppData) => {
+      await this.hooks.runPostHooks('patch', member, repositories, {
+        appData: patchedAppData,
+        itemId,
+      });
+      return patchedAppData;
+    });
```
dialexo, line 160:
```suggestion
    const appData = await appDataRepository.patch(itemId, appDataId, body);
    await this.hooks.runPostHooks('patch', member, repositories, {
      appData: patchedAppData,
      itemId,
    });
    return appData;
```

src/services/item/plugins/app/appSetting/plugins/file/index.ts
```diff
@@ -55,24 +55,28 @@ const basePlugin: FastifyPluginAsync<GraaspPluginFileOptions> = async (fastify,
...
     await appSettingFileService.deleteOne(actor, repositories, appSetting);
   };
   appSettingService.hooks.setPostHook('delete', deleteHook);
 
   // app setting copy hook
-  const hook = async (actor: Member, repositories: Repositories, newAppSettings: AppSetting[]) => {
+  const hook = async (
+    actor: Member,
+    repositories: Repositories,
+    { appSettings, originalItemId, copyItemId },
```
dialexo, line 67: Does Typescript infer the types here (how 🤔) ? Or are they lost to `any`?

src/services/item/plugins/app/appSetting/plugins/file/index.ts
```diff
@@ -55,24 +55,28 @@ const basePlugin: FastifyPluginAsync<GraaspPluginFileOptions> = async (fastify,
...
+    const toCopy = appSettings.filter(isFileSetting);
     if (toCopy.length) {
       await appSettingFileService.copyMany(actor, repositories, toCopy);
     }
   };
   appSettingService.hooks.setPostHook('copyMany', hook);
 
   // prevent patch on app setting file
-  const patchPreHook = async (actor: Actor, repositories: Repositories, appSetting: AppSetting) => {
+  const patchPreHook = async (actor: Actor, repositories: Repositories, { appSetting }) => {
```
dialexo, line 79: Same question about the type of `{ appSetting }`

src/services/item/plugins/app/appSetting/service.ts
dialexo: Now that I think about it, maybe update operations should have the hooks `args` set to both `{ original: item, update: Item }`. But I've just checked and somehow the items are missing the `update` hooks, oops 🤔

src/services/item/plugins/app/appSetting/service.ts
```diff
@@ -53,11 +78,22 @@ export class AppSettingService {
...
+
+    return appSettingRepository
+      .patch(itemId, appSettingId, body)
+      .then(async (patchedAppSetting) => {
+        await this.hooks.runPostHooks('patch', member, repositories, {
+          appSetting: patchedAppSetting,
+          itemId,
+        });
+        return patchedAppSetting;
+      });
```
dialexo, line 96: Same as above, use `async / await`

src/services/item/plugins/app/ws/hooks.ts
dialexo: You could but I think it's fine

dialexo (changes requested):
Thank you for your work 🦞 I've left a few comments

src/services/item/plugins/app/ws/hooks.ts
```diff
@@ -0,0 +1,181 @@
...
+function registerAppActionTopic(
+  websockets: WebsocketService,
+  appActionService: AppActionService,
+  itemService: ItemService,
+) {
+  websockets.register(appActionsTopic, async (req) => {
+    const { channel: id, member } = req;
+    const repositories = buildRepositories();
+    const item = await itemService.get(member, repositories, id);
+    await validatePermission(repositories, PermissionLevel.Admin, member, item);
```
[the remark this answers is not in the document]
swouf: Yes

src/services/item/plugins/app/appData/service.ts
swouf: I think this is a mistake. The checks should be performed before this hook.

src/services/item/plugins/app/appSetting/plugins/file/index.ts
```diff
@@ -55,24 +55,28 @@ const basePlugin: FastifyPluginAsync<GraaspPluginFileOptions> = async (fastify,
   });
 
   // register post delete handler to remove the file object after item delete
-  const deleteHook = async (actor: Actor, repositories: Repositories, appSetting: AppSetting) => {
+  const deleteHook = async (actor: Actor, repositories: Repositories, { appSetting, itemId }) => {
```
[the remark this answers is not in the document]
swouf: No, I think it cannot infer the type here.

src/services/item/plugins/app/appSetting/service.ts
```diff
@@ -137,7 +175,11 @@ export class AppSettingService {
         );
         newAppSettings.push(newSetting);
       }
-      await this.hooks.runPostHooks('copyMany', actor, repositories, newAppSettings);
+      await this.hooks.runPostHooks('copyMany', actor, repositories, {
+        appSettings: newAppSettings,
+        originalItemId: original.id,
+        copyItemId: copy.id,
+      });
```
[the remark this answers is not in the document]
swouf: At the moment, no service uses the pre-hook. But I added to be consistent and if at some point one needs to hook into that service.

swouf:
I just tested it in the dev environment. It works well. Any other comments @pyphilia, @dialexo, @spaenleh?

src/services/item/plugins/app/appData/service.ts
```diff
@@ -56,12 +69,15 @@ export class AppDataService {
...
   ) {
     const { appDataRepository, memberRepository, itemRepository } = repositories;
+
     // check member exists
     if (!actorId) {
       throw new MemberCannotWriteItem();
     }
     const member = await memberRepository.get(actorId);
 
+    await this.hooks.runPreHooks('post', member, repositories, { appData: body, itemId });
```
dialexo, line 79: Should this be moved after the checks (e.g line 103 instead)?

src/services/item/plugins/app/appData/service.ts
```diff
@@ -145,6 +171,8 @@ export class AppDataService {
     }
     const member = await memberRepository.get(memberId);
 
+    await this.hooks.runPreHooks('delete', member, repositories, { appDataId, itemId });
```
dialexo, line 174: Same as above, should we run the checks first and move to line 197?

src/services/item/plugins/app/appSetting/plugins/file/index.ts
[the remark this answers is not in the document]
dialexo: Then I think you should explicitly type them with
```suggestion
  const deleteHook = async (actor: Actor, repositories: Repositories, { appSetting, itemId }: { appSetting: AppSetting, itemId: string }) => {
```

src/services/item/plugins/app/appAction/service.ts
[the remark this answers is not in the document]
dialexo: > Q: what is the use case for broadcasting actions? Currently we don't have events for the items counterpart, but I guess the semantics are up to the app themselves. I'm just worried about the load on the server if we track and relay each individual action in an app, but again it depends on the app. Edit: i've seen later that only admins can subscribe to these events so I guess the load should be alright.

Bumping this question from previous review

src/services/item/plugins/app/appSetting/plugins/file/index.ts
dialexo: Same as above:
```suggestion
    { appSettings, originalItemId, copyItemId }: { appSettings: AppSetting, originalItemId: string, copyItemId: string },
```

src/services/item/plugins/app/appSetting/plugins/file/index.ts
dialexo:
```suggestion
  const patchPreHook = async (actor: Actor, repositories: Repositories, { appSetting }: { appSetting: AppSetting }) => {
```

src/services/item/plugins/app/appSetting/service.ts
```diff
@@ -30,10 +47,17 @@ export class AppSettingService {
     }
     const member = await memberRepository.get(memberId);
 
+    await this.hooks.runPreHooks('post', member, repositories, { appSetting: body, itemId });
```
dialexo, line 50: Should this run after the check below at line 53?

src/services/item/plugins/app/appSetting/service.ts
```diff
@@ -73,14 +103,16 @@ export class AppSettingService {
     }
     const member = await memberRepository.get(memberId);
 
+    await this.hooks.runPreHooks('delete', member, repositories, { appSettingId, itemId });
```
dialexo, line 106: Same as above, run after the checks?

src/services/item/plugins/app/ws/events.ts
dialexo: > Use `appDataTopic` ? Same for the others.

You can use `typeof appDataTopic` if you want the constant string type

src/services/item/plugins/app/ws/events.ts
```diff
@@ -0,0 +1,84 @@
...
+export const appActionsTopic = 'app-actions';
+export const appSettingsTopic = 'app-settings';
+
+type AppOperations = 'post' | 'patch' | 'delete';
+
+/**
+ * All websocket events for app will have this shape
+ */
+interface AppEvent {
+  kind: string;
```
dialexo, line 19: You could use the union of the topics

dialexo (changes requested):
Thank you for your update 🦺 I have a few nitpicks left, otherwise LGTM. Let me know if they make sense

src/services/item/plugins/app/appData/service.ts
swouf: I agree.

src/services/item/plugins/app/appAction/service.ts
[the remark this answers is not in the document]
swouf: I agree that broadcasting all actions may cause an undesirable load on the server in case of a lot of users. But as you mentioned, only admins can subscribe to these events. Moreover, I think that the hook for the actions, which leads to subscribing to the app actions topic, is not often used in apps. The main use case would be to make realtime analytics panels in current and future apps.

src/services/item/plugins/app/appSetting/plugins/file/index.ts
dialexo: Bump, I think this type is still missing 😁

src/services/item/plugins/app/appSetting/service.ts
```diff
@@ -76,11 +106,13 @@ export class AppSettingService {
     // delete an app data is allowed to admins
     await this.itemService.get(member, repositories, itemId, PermissionLevel.Admin);
 
+    await this.hooks.runPreHooks('delete', member, repositories, { appSettingId, itemId });
+
```
dialexo, line 110: You could move this to the line just below after `const appSetting = await appSettingRepository.get(appSettingId);`. The reasoning is that the get could throw, in which case you would have a state where you have dispatched the pre hooks, but the delete will never even be attempted on the object. I don't think it affects much in the rest of the code but this could help in maintaining consistency.

dialexo (changes requested):
Some very small nitpicks left 🔍 I'll be happy to approve otherwise!

PR #520, opened 2023-07-28
File upload oom

pyphilia:
- does not upload in parallel
- Uniquement use streams to upload files. The file size is retrieved once uploaded.
- Use of s3's upload that can upload in chunk. 
- move the transaction to apply on each file instead of all the files. This allows one file to throw but still keep the other ones.

ref #519


src/services/item/plugins/file/service.ts
```diff
@@ -92,80 +88,101 @@ class FileItemService {
...
+  ) {
+    // compute body data from file's fields
+    if (fields) {
+      const fileBody = Object.fromEntries(
+        Object.keys(fields).map((key) => [
+          key,
+          (fields[key] as unknown as { value: string })?.value,
+        ]),
+      );
+    }
```
dialexo, line 140: Not sure why this is used for? Can `fileBody` be read from outside the if scope?

src/services/item/plugins/file/service.ts
```diff
@@ -92,80 +88,101 @@ class FileItemService {
...
+    // check member storage limit
+    await this.checkRemainingStorage(actor, repositories);
+
+    await this.fileService.upload(actor, {
+      file: stream,
+      filepath,
+      mimetype,
     });
```
dialexo, line 148: Just to be sure, so this operation handles backpressure itself?

src/services/item/plugins/thumbnail/index.ts
```diff
@@ -69,14 +70,20 @@ const plugin: FastifyPluginAsync<GraaspThumbnailsOptions> = async (fastify, opti
...
             throw new UploadFileNotImageError();
           }
 
-          await thumbnailService.upload(member, buildRepositories(manager), itemId, file);
+          await thumbnailService.upload(
+            member,
+            buildRepositories(manager),
+            itemId,
+            // TODO: not good
+            file.file as ReadableStream as WriteStream,
```
dialexo, line 85: Not sure what's up here @pyphilia @morganridel

src/services/item/plugins/thumbnail/service.ts
```diff
@@ -19,7 +20,7 @@ export class ItemThumbnailService {
     this.itemService = itemService;
   }
 
-  async upload(actor: Member, repositories: Repositories, itemId: string, file: UploadedFile) {
+  async upload(actor: Member, repositories: Repositories, itemId: string, file: WriteStream) {
```
dialexo, line 23: Not sure if the type makes sense here, how can an uploadable file come from a writable stream? Where does the constraint come from? @morganridel

src/services/member/plugins/thumbnail/service.ts
```diff
@@ -17,8 +19,8 @@ export class MemberThumbnailService {
   }
 
   // upload self avatar
-  async upload(actor: Member, repositories: Repositories, file: SavedMultipartFile) {
-    await this.thumbnailService.upload(actor, actor.id, file);
+  async upload(actor: Member, repositories: Repositories, file: MultipartFile) {
+    await this.thumbnailService.upload(actor, actor.id, file.file as unknown as WriteStream);
```
dialexo, line 23: Corollary from above

src/services/thumbnail/service.ts
```diff
@@ -23,42 +20,24 @@ export class ThumbnailService {
...
-      throw e;
-    } finally {
-      if (existsSync(fileStorage)) {
-        rm(fileStorage, { recursive: true }).catch((e) => console.error(e));
-      } else {
-        // do not throw if folder has already been deleted
-        // log?.error(`${fileStorage} was not found, and was not deleted`);
-      }
-    }
+  async upload(actor: Member, id: string, file: WriteStream) {
```
dialexo, line 26: Same as above

src/services/item/plugins/file/service.ts
```diff
@@ -72,67 +69,66 @@ class FileItemService {
...
 
-    // check rights
-    if (parentId) {
-      const item = await repositories.itemRepository.get(parentId);
-      await validatePermission(repositories, PermissionLevel.Write, actor, item);
+    // duplicate stream to use in thumbnails
+    const streamForThumbnails = new PassThrough();
+    if (MimeTypes.isImage(mimetype)) {
+      stream.pipe(streamForThumbnails);
     }
```
dialexo, line 112: Can a stream be used by 2 different consumers? From my understanding, the stream here is piped both to the file upload and to the thumbnail upload. Does this actually create a copy of the stream data? What if the consumers don't use the data at the same rate?

src/services/member/plugins/thumbnail/index.ts
```diff
@@ -59,14 +63,16 @@ const plugin: FastifyPluginAsync<GraaspThumbnailsOptions> = async (fastify, opti
...
+          if (!file) {
+            throw new UploadEmptyFileError();
+          }
           // check file is an image
           if (!file.mimetype.includes('image')) {
             throw new UploadFileNotImageError();
           }
 
-          await thumbnailService.upload(member, buildRepositories(manager), file);
+          await thumbnailService.upload(member, buildRepositories(manager), file.file);
```
dialexo, line 75: Minor
```suggestion
          const { file: stream } = file;
          await thumbnailService.upload(member, buildRepositories(manager), stream);
```

dialexo (commented):
Thank you for your work :balloon:  Not sure about the stream duplication. Do you have more info on this?

PR #522, opened 2023-07-28
fix: copied item should not inherit public tag from original

spaenleh:
In this PR:
- fix: copied item does not inherit public tag from original
- fix: remove dependency on ItemLike in Entity

### DIFF
--- a/src/services/item/plugins/itemLike/itemLike.ts
+++ b/src/services/item/plugins/itemLike/itemLike.ts
@@ -9,14 +9,12 @@ import {
 } from 'typeorm';
 import { v4 } from 'uuid';
 
-import { ItemLike as GraaspItemLike } from '@graasp/sdk';
-
 import { Item } from '../../../item/entities/Item';
 import { Member } from '../../../member/entities/member';
 
 @Entity()
 @Unique('id', ['creator', 'item'])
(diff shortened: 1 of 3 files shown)


src/services/item/plugins/itemLike/itemLike.ts
```diff
@@ -9,14 +9,12 @@ import {
...
 
-import { ItemLike as GraaspItemLike } from '@graasp/sdk';
-
 import { Item } from '../../../item/entities/Item';
 import { Member } from '../../../member/entities/member';
 
 @Entity()
 @Unique('id', ['creator', 'item'])
-export class ItemLike extends BaseEntity implements GraaspItemLike {
+export class ItemLike extends BaseEntity {
```
pyphilia, line 17: Why did that get removed?

spaenleh: Following the discussion we had about the creator field (privacy reasons), I think it would make sens that the DB objects are something of their own and the types in sdk reflect more the API types, so in this case when getting an ItemLike from the api we would not get the creator. In such a  case, I think we can decouple the DB types from the api types. I can remove this change if it is not deemed appropriate now.

PR #527, opened 2023-07-31
fix: null creator in ws hooks

### LINKED ISSUE #526: [ws] Deleting an item triggers an error in websockets
Screenshot 2023-07-31 at 15 39 02

The error makes sense because we're not joining creators when getting the descendants (to be deleted)

pyphilia:
This error happens for the following case:

- A > B > C with a membership for each of them for another member Bob
- Removing B, by cascade all the memberships of B and C get deleted
- Since A still have a membership, it triggers the websockets "shared items"
- But since C is get via `getDescendants` which does not return the creator, it fails

In this PR:
- I've added `null` to item's creator and itemMembership's creator
- This highlighted the ws hooks that needed more check on these fields
- added a test
- added postHook for move 

close #526

### DIFF
--- a/src/services/item/service.ts
+++ b/src/services/item/service.ts
@@ -31,7 +31,7 @@ export class ItemService {
     update: { pre: { item: Item }; post: { item: Item } };
     delete: { pre: { item: Item }; post: { item: Item } };
     copy: { pre: { original: Item }; post: { original: Item; copy: Item } };
-    move: { pre: { source: Item }; post: { source: Item; destination: Item } };
+    move: { pre: { source: Item; destination: Item }; post: { source: Item; destination: Item } };
   }>();
 
   async post(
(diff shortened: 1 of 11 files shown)


src/services/item/service.ts
```diff
@@ -323,6 +323,13 @@ export class ItemService {
```
dialexo, line 323: Should we run the prehook here as well?

src/services/item/ws/hooks.ts
```diff
@@ -94,9 +96,9 @@ function registerMemberItemsTopic(websockets: WebsocketService, itemService: Ite
   // on create, notify own items of creator with new item IF path is root
   itemService.hooks.setPostHook('create', async (actor, repositories, { item }) => {
     const parentId = getParentFromPath(item.path);
-    if (parentId === undefined) {
+    if (parentId === undefined && actor) {
       // root item, notify creator
-      websockets.publish(memberItemsTopic, item.creator.id, OwnItemsEvent('create', item));
+      websockets.publish(memberItemsTopic, actor.id, OwnItemsEvent('create', item));
```
dialexo, line 101: Is `actor` always the creator here? So shouldn't it simply check `if (parentId === undefined && item.creator)`?

src/services/item/ws/hooks.ts
```diff
@@ -125,9 +127,12 @@ function registerMemberItemsTopic(websockets: WebsocketService, itemService: Ite
   // - notify own items of creator of deleted item IF path is root
   // - notify members that have memberships on this item of delete
   //   (before with prehook, otherwise memberships lost on cascade from db!)
+  // todo: currently items can be hard deleted without going to the trash
+  // --> preventing this behavior leads to delete the following code
```
dialexo, line 131: I don't think we necessarily need to remove the posthook even with the recycle bin? E.g. it should still be true that if an item is recycled then permanently deleted, the change should be broadcasted?

dialexo (changes requested):
Let me know what you think :beer:

dialexo (approved):

Issue #518, opened 2023-07-28
[files] upload file concurrently 


pyphilia:

pyphilia:
Files are not uploaded concur

PR #531, opened 2023-08-02
fix: fix published queries to prevent cross table queries

pyphilia:
Following #518 we found out some queries were crossing tables, leading to huge query results. These results were hidden because of the use of `getMany()` that seems to remove duplicata. Running `getRawMany()` highlights the real number of rows retrieved (Typeorm's generated SQL being `FROM x, y`).

To avoid that, we use the correct repository to start with. 

We cannot really test it as Typeorm hides the duplicates from us. If we really want to test we should do:

```javascript
// IN REPOSITORY

 async _getByCategoriesQuery(...) {
  return itemRepository.createQueryBuilder(...)
}

 async getByCategories(...) {
  const query = this._getByCategoriesQuery()
  return query.getMany()
}

// IN TEST FILE
// this should highlight the duplicates
await PublishedRepository._getByCategoriesQuery().getRawMany()
```


spaenleh (approved):
Changes look good to me, I let the others chime in for further feedback, nice find on how Typeorm handles these ! 💪

src/services/item/plugins/published/repositories/itemPublished.ts
```diff
@@ -39,23 +40,23 @@ export const ItemPublishedRepository = AppDataSource.getRepository(ItemPublished
...
       .leftJoinAndSelect('item.creator', 'creator')
       .innerJoin('item_published', 'ip', 'ip.item_path = item.path')
       .getMany();
+
     return publishedRows;
   },
 
   // return public item entry? contains when it was published
-  async getItemsForMember(memberId: string) {
+  async getItemsForMember(itemRepository: typeof ItemRepository, memberId: string) {
```
morganridel, line 55: At this point I might consider moving the function directly in ItemRepository `getPublishedItemsForMember`, because I often expect a `X` Entity when I call a `XRepository`.
Most of these queries are basically using item published table as a filter but return `Item`

morganridel (approved):

PR #534, opened 2023-08-03
fix: optimize db queries

morganridel:
~PR is not perfectly ready yet, I will try to remove the last bit of formatting problems tomorrow~ Should be better now

It is not a good idea to do x*n DB request when receiving n ids as input, I tried to reduce to a constant x queries (independent on input size) for the item-memberships call while keeping the same behavior.

Also GIST indices on DB work for all the ltree operators (`@>` ...). They were tested with `EXPLAIN ANALYZE` on realistic queries directly on the DB and are retrieving the ancestors instantly.

Also fix `ItemTagRepository.hasManyForMany` which didn't work

Depending on results on performance I should take a look at the following in the future:

* Optimizing `/children` call queries (should be better with the new index)
*  Optimizing queries from the `ItemTag` repository
* Websockets is a pain as it do a `validatePermission` for EACH item channel on app load, sending a lot of DB requests


src/services/item/service.ts
```diff
@@ -112,17 +112,21 @@ export class ItemService {
...
+      if (memberships?.data[id] === undefined) {
         delete result.data[id];
-        if (e instanceof CoreError || e instanceof Error) {
-          result.errors.push(e);
-        }
       }
     }
-    return result;
+
+    return { data: result.data, errors: [...(memberships?.errors ?? []), ...result.errors] };
```
pyphilia, line 129:
```suggestion
    return { data: result.data, errors: result.errors.concat(memberships?.errors ?? []) };
```

src/services/itemMembership/index.ts
```diff
@@ -36,7 +36,7 @@ const plugin: FastifyPluginAsync = async (fastify) => {
       fastify.get<{ Querystring: { itemId: string[] } }>(
         '/',
         { schema: getItems, preHandler: fastify.fetchMemberInSession },
-        async ({ member, query: { itemId: ids }, log }) => {
+        async ({ member, id, query: { itemId: ids }, log }) => {
```
pyphilia, line 39: `id`?

src/services/itemMembership/repository.ts
```diff
@@ -97,26 +97,22 @@ export const ItemMembershipRepository = AppDataSource.getRepository(ItemMembersh
...
+    const ids = items.map((i) => i.id);
+    const query = this.createQueryBuilder('item_membership')
+      .innerJoin('item', 'descendant', 'item_membership.item_path @> descendant.path')
+      .where('descendant.id in (:...ids)', { ids: ids });
     if (memberId) {
       query.andWhere('member.id = :memberId', { memberId });
     }
 
+    query
+      .leftJoinAndSelect('item_membership.item', 'item')
```
pyphilia, line 113: Could you exchange `innerJoin` to `innerJoinAndSelect` and remove this line?

src/services/itemMembership/repository.ts
```diff
@@ -133,6 +129,54 @@ export const ItemMembershipRepository = AppDataSource.getRepository(ItemMembersh
...
+    }
+
+    const ids = items.map((i) => i.id);
+
+    const query = ItemMembershipRepository.createQueryBuilder('item_membership')
+      // Map each membership to the item it can affect
+      .innerJoin('item', 'descendant', 'item_membership.item_path @> descendant.path')
+      // Join for entity result
+      .leftJoinAndSelect('item_membership.member', 'member')
+      .leftJoinAndSelect('item_membership.item', 'item')
```
pyphilia, line 148: Select from the innerjoin ?

src/services/utils.ts
```diff
@@ -53,3 +54,33 @@ export const buildItemLink = (item: Item) => {
...
+  // eslint-disable-next-line prefer-const
+  let [sql, params] = query.getQueryAndParameters();
+
+  params.forEach((value, index) => {
+    if (typeof value === 'string') {
+      sql = sql.replace(`$${index + 1}`, `'${value}'`);
+    }
+    if (typeof value === 'object') {
+      if (Array.isArray(value)) {
+        sql = sql.replace(
```
pyphilia, line 70: Maybe create another file for this debug function.

pyphilia (approved):
It looks really good, I'm gonna test it locally and try to deploy on dev 🍔

src/services/itemMembership/repository.ts
morganridel: No this is not the same join, the first one is just used as a filter to retrieve memberships that are parents, but this is not a join where `item_membership.item_path = item.path`. The second one is for filling the entity that will be returned with the real linked item
I.E: I need to do the two join, at least from what I was trying to do

pyphilia:
I've added a bunch of tests for `validatePermission` and `validatePermissionMany` (I've caught a small bug). I think I've covered most of the cases.

morganridel:
Really great idea to add unit tests, perfect usecase :)

src/services/authorization.test.ts
```diff
@@ -0,0 +1,2660 @@
+import { ItemTagType, PermissionLevel, UUID } from '@graasp/sdk';
```
dialexo, line 1: Very nice all those unit tests :sparkles: for readability purposes, i would suggest adding blank newlines between `it` blocks as well as `describe` blocks. It makes it easier to distinguish the test cases visually.

src/services/authorization.ts
```diff
@@ -26,27 +28,120 @@ const permissionMapping = {
...
+  ]);
+
+  const result: ResultOf<ItemMembership | null> = {
+    data: inheritedMemberships?.data ?? {},
+    errors: [],
+  };
+
+  for (const item of items) {
+    const highest = result.data[item.id]?.permission;
+    const isValid = highest && permissionMapping[highest].includes(permission);
```
dialexo, line 60: Is this equivalent to (i.e. can you replace instances of `permissionMapping` with):
```suggestion
    const isValid = highest && PermissionLevelCompare.lte(permission, highest);
```

src/services/authorization.ts
```diff
@@ -26,27 +28,120 @@ const permissionMapping = {
...
+    const isPublic = tags.data[item.id].includes(ItemTagType.Public);
+
+    // HIDDEN CHECK - prevent read
+    // cannot read if your have read access only
+    if (highest === PermissionLevel.Read || (isPublic && !highest)) {
```
dialexo, line 65: Q: can members with write access view hidden items?

src/services/authorization.ts
```diff
@@ -26,27 +28,120 @@ const permissionMapping = {
...
   item: Item,
 ): Promise<ItemMembership | null> => {
   // get best permission for user
   // but do not fetch membership for signed out member
+
   const inheritedMembership = member
     ? await itemMembershipRepository.getInherited(item, member, true)
     : null;
   const highest = inheritedMembership?.permission;
   const isValid = highest && permissionMapping[highest].includes(permission);
```
dialexo, line 133: Same as above

src/services/item/plugins/itemTag/repository.ts
```diff
@@ -60,14 +60,22 @@ export const ItemTagRepository = AppDataSource.getRepository(ItemTag).extend({
       }),
     );
 
-    const hasTags = await query
+    const hasTags: ItemTag[] = await query
       .andWhere('itemTag.type IN (:...types)', { types: tagTypes })
-      .getOne();
+      .getMany();
```
dialexo, line 65: Info: why did the previous implementation not return many?

src/services/itemMembership/repository.ts
```diff
@@ -97,26 +97,22 @@ export const ItemMembershipRepository = AppDataSource.getRepository(ItemMembersh
...
-      }),
-    );
+    if (items.length === 0) {
+      return { data: {}, errors: [] };
+    }
 
+    const ids = items.map((i) => i.id);
+    const query = this.createQueryBuilder('item_membership')
+      .innerJoin('item', 'descendant', 'item_membership.item_path @> descendant.path')
+      .where('descendant.id in (:...ids)', { ids: ids });
```
dialexo, line 107:
```suggestion
      .where('descendant.id in (:...ids)', { ids });
```

src/services/itemMembership/repository.ts
```diff
@@ -133,6 +129,54 @@ export const ItemMembershipRepository = AppDataSource.getRepository(ItemMembersh
...
+      query.andWhere('item.id not in (:...ids)', { ids: ids });
+    }
+    // Keep only closest membership per descendant
+    query
+      .addSelect('descendant.id')
+      .distinctOn(['descendant.id'])
+      .orderBy('descendant.id')
+      .addOrderBy('nlevel(item_membership.item_path)', 'DESC');
+
+    // annoyingly, getMany removes dupplicate entities, however in this case two items might be linked to the same effective membership
```
dialexo, line 162:
```suggestion
    // annoyingly, getMany removes duplicate entities, however in this case two items might be linked to the same effective membership
```

src/services/utils.ts
```diff
@@ -53,3 +54,33 @@ export const buildItemLink = (item: Item) => {
 
   return itemLink;
 };
+
+// Print a query with parameters already filled in, ready to be sent to the database
+// Only for debug purposes. DO NOT USE in production. (injection risk)
+export const printFilledSQL = <T extends ObjectLiteral>(query: QueryBuilder<T>) => {
```
dialexo, line 60: Consider running a `if (NODE_ENV === 'production') { throw new Error() }` here?

dialexo (approved):
Wonderful work :strawberry: I've left a few questions and minor improvements :+1:

spaenleh (approved):
Very nice ! Nice tests too !

src/services/item/plugins/itemTag/repository.ts
morganridel: The previous implementation didn't work and wasn't used

src/services/authorization.ts
morganridel: I guess so, so you can change the hidden status? I just kept the existing behavior

src/services/authorization.ts
morganridel: I don't know how this part work exactly, I just copied it from the "validateOne" implementation

pyphilia: In fact that might be true, but rather:
```suggestion
 const isValid = highest && PermissionLevelCompare.gte(permission, highest);
```

PR #539, opened 2023-08-08
feat: always redirect mobile auth to universal auth link

### LINKED ISSUE #538: Use http redirect on mobile auth endpoint
Since we now use universal links to login mobile users (#498), we can use the existing `/m/auth` endpoint such that it acts as an intermediate domain between the `auth` front-end and the universal link which should open the app. We can use a 304 header to make the in-app browser redirect to into the mobile app using the universal link.

In the future, we may want to also use the universal link in their "true" form, such that opening a builder, player, or library link will automatically open the app if it is installed, or the browser otherwise, to the specified resource.

dialexo:
Closes #538


spaenleh (approved):
Looks good to me ! Hope it works and it solves our problem !

PR #544, opened 2023-08-11
fix: throw for undefined member id in token

### LINKED ISSUE #543: Handle undefined member id in token

pyphilia:
This PR fix the undefined member id in token by:
- using the member repository `get` instead of raw `findOneById`
- refactor auth utils in a dedicated file + `AuthService` (probably we don't need the decoration anymore). It's still confused because we have many functions doing "similar" logic , but I've left major refactor for later
- regression tests + unit tests of auth utils

Additionally:
- All other functions using  `findOneById` have an additional check over id

close #543

### DIFF
--- a/src/services/auth/plugins/magicLink/service.ts
+++ b/src/services/auth/plugins/magicLink/service.ts
@@ -9,7 +9,6 @@ import { DEFAULT_LANG } from '@graasp/sdk';
 import { JWT_SECRET } from '../../../../utils/config';
 import {
   InvalidToken,
-  MemberNotFound,
   MemberNotSignedUp,
   TokenExpired,
   UnexpectedError,
@@ -53,13 +52,9 @@ export class MagicLinkService {
   async auth(actor: Actor, repositories: Repositories, token: string) {
     try {
       // verify and extract member info
-      const result = await promisifiedJwtVerify(token, JWT_SECRET, {});
-
+      const result = (await promisifiedJwtVerify(token, JWT_SECRET, {})) as jwt.JwtPayload;
       // pre test the user existence to avoid providing a key
-      const member = await repositories.memberRepository.get(result.sub);
-      if (!member) {
-        throw new MemberNotFound(result.sub);
-      }
+      await repositories.memberRepository.get(result.id);
 
       return result;
     } catch (error) {
--- a/src/services/auth/plugins/magicLink/service.ts
+++ b/src/services/auth/plugins/magicLink/service.ts
@@ -9,6 +9,7 @@ import { DEFAULT_LANG } from '@graasp/sdk';
 import { JWT_SECRET } from '../../../../utils/config';
 import {
   InvalidToken,
+  MemberNotFound,
   MemberNotSignedUp,
   TokenExpired,
   UnexpectedError,
@@ -49,10 +50,17 @@ export class MagicLinkService {
     }
   }
 
-  async auth(actor: Actor, repositories: Repositories, token) {
+  async auth(actor: Actor, repositories: Repositories, token: string) {
     try {
       // verify and extract member info
       const result = await promisifiedJwtVerify(token, JWT_SECRET, {});
+
+      // pre test the user existence to avoid providing a key
+      const member = await repositories.memberRepository.get(result.sub);
+      if (!member) {
+        throw new MemberNotFound(result.sub);
+      }
+
       return result;
     } catch (error) {
       // the token caused the error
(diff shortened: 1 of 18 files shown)


src/services/auth/plugins/magicLink/service.ts
```diff
@@ -49,10 +49,13 @@ export class MagicLinkService {
     }
   }
 
-  async auth(actor: Actor, repositories: Repositories, token) {
+  async auth(actor: Actor, repositories: Repositories, token: string) {
     try {
       // verify and extract member info
-      const result = await promisifiedJwtVerify(token, JWT_SECRET, {});
+      const result = (await promisifiedJwtVerify(token, JWT_SECRET, {})) as jwt.JwtPayload;
```
dialexo, line 55: Info: why is the cast required here? Consider adding a comment

src/services/auth/plugins/magicLink/service.ts
```diff
@@ -49,10 +49,13 @@ export class MagicLinkService {
...
+      // pre test the user existence to avoid providing a key
+      await repositories.memberRepository.get(result.id);
```
dialexo, line 57: Info: This will throw if there is no member?

src/services/auth/plugins/mobile/mobile.test.ts
```diff
@@ -310,7 +307,7 @@ describe('Mobile Endpoints', () => {
 
     it('Fail to authenticate if verifier and challenge do not match', async () => {
       const member = await saveMember(BOB);
-      const t = jwt.sign({ id: member.id }, JWT_SECRET);
+      const t = jwt.sign({ sub: member.id }, JWT_SECRET);
```
dialexo, line 310: Info: why did the property name change?

src/services/auth/plugins/mobile/service.ts
```diff
@@ -89,6 +90,9 @@ export class MobileService {
         throw new ChallengeFailed();
       }
 
+      // pre test the user existence to avoid providing a key
+      await repositories.memberRepository.get(memberId);
```
dialexo, line 94: Same question as above

src/services/auth/service.ts
```diff
@@ -0,0 +1,100 @@
...
+    const { challenge, url } = options;
+
+    // generate token with member info and expiration
+    const token = await promisifiedJwtSign({ sub: member.id, challenge }, JWT_SECRET, {
+      expiresIn: `${REGISTER_TOKEN_EXPIRATION_IN_MINUTES}m`,
+    });
+
+    const redirectionUrl = getRedirectionUrl(url);
+    const linkPath = challenge ? '/m/deep-link' : '/auth';
+    const link = new URL(`${linkPath}?t=${token}&url=${redirectionUrl}`, PUBLIC_URL).toString();
```
dialexo, line 53: Arf this will mean a rebase #539. But nevermind, I'll merge from main.

src/services/auth/service.ts
```diff
@@ -0,0 +1,100 @@
...
+    const { challenge, lang, url } = options;
+
+    // generate token with member info and expiration
+    const token = await promisifiedJwtSign({ sub: member.id, challenge }, JWT_SECRET, {
+      expiresIn: `${LOGIN_TOKEN_EXPIRATION_IN_MINUTES}m`,
+    });
+
+    const redirectionUrl = getRedirectionUrl(url);
+    const linkPath = challenge ? '/m/deep-link' : '/auth';
+    const link = new URL(`${linkPath}?t=${token}&url=${redirectionUrl}`, PUBLIC_URL).toString();
```
dialexo, line 84: Note to self: same here

src/services/auth/utils.test.ts
```diff
@@ -0,0 +1,106 @@
...
+import { fetchMemberInSession, verifyMemberInAuthToken, verifyMemberInSession } from './utils';
+
+const buildRequest = (memberId?: string) =>
+  ({
+    session: {
+      get: () => memberId,
+      delete: jest.fn(),
+    },
+    log: { warn: jest.fn() },
+  } as unknown as FastifyRequest & { member?: Member });
```
dialexo, line 21: Not sure about the types here. Is it for testing purposes only?

dialexo (approved):
Thank you for your work :partying_face: looks much cleaner and more robust now. I just have a few questions, which answers could be useful code comments.

src/services/auth/plugins/mobile/mobile.test.ts
pyphilia: It was false in the test 🥲

src/services/auth/utils.test.ts
pyphilia: Yes, test only. It's not easy to mock a fastify request.

PR #549, opened 2023-08-11
chore: ignore authorization header error in sentry

pyphilia:
This error happens like always and is supposed to happen, I think it makes sense to ignore it. Let me know what you think.

Screenshot 2023-08-11 at 17 12 04

### DIFF
--- a/src/sentry.ts
+++ b/src/sentry.ts
@@ -60,6 +60,7 @@ export function initSentry(instance: FastifyInstance): {
       }
       return breadcrumb;
     },
+    ignoreErrors: ['missing authorization header'],
   });
 
   if (SentryConfig.enablePerformance) {


dialexo (changes requested):
Shouldn't we ignore those directly in the sentry dashboard? That way we still receive them but they are filtered at the provider, and we have flexibility in adding / removing them.

spaenleh:
I understand why we would like to ignore it in the dashboard, but there are no "ignore indefinitely" option. For reference these are the options:
Screenshot 2023-08-14 at 08 03 55
So unless we ignore it for a 100 years ...

dialexo:
> I understand why we would like to ignore it in the dashboard, but there are no "ignore indefinitely" option. For reference these are the options: Screenshot 2023-08-14 at 08 03 55 So unless we ignore it for a 100 years ...

Actually if you click on the ignore button directly (not on the arrow menu), it seems the issue is completely ignored. See https://graasp.sentry.io/issues/4264884885/?project=6177699&query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=14d&stream_index=1

pyphilia:
But you still send it to sentry, and we're limited on the number of transactions as well. Won't also it continue to trigger on slack if we ignore it in the dashboard only?

spaenleh:
@dialexo Nice find, i didn't even think about checking the button directly ... 🙃

dialexo:
I've found some docs on errors quota: 
https://docs.sentry.io/product/accounts/quotas/#inbound-data-filters
Do we want to setup inbound filters at https://graasp.sentry.io/settings/projects/graasp-core/filters/data-filters/ ?

pyphilia:
Sponsored plan:
- 5M errors
- 10M performance units
- 10K replays
- 10G Battachments
But we're really far from the max, so it's okay to filter from sentry.

PR #551, opened 2023-08-14
fix: allow bearer token by using attemptVerifyAuthentication instead of fetchMemberInSession

### LINKED ISSUE #550: Allow Bearer token for public endpoints
Currently `fetchMemberInSession` does not include bearer token (mobile access).

pyphilia:
`fetchMemberInSession` only checked the cookie value.

Whereas `attemptVerifyAuthentication` does both `fetchMemberInSession` + `verifyBearerToken` which allows request to correctly get the member from the `Bearer` Authorization.

This effected all the public endpoints (ie. get item, get tags, etc)

close #550

### DIFF
--- a/src/declarations.d.ts
+++ b/src/declarations.d.ts
@@ -73,6 +73,7 @@ declare module 'fastify' {
       captcha: string,
       actionType: RecaptchaActionType,
     ) => Promise<void>;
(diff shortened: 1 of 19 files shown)


dialexo (approved):
LGTM 🛺

Issue #383, opened 2023-06-12
Add HTML (SCORM) support through sandboxed (?) iframe


dialexo:
Import single-file HTML pages as well as SCORM packages (probably not the full spec?)

- Q: can we reuse parts of the H5P implementation into a generic HTML content service?
- Q: security implications? What should we allow (is interweave enough)? What about e.g. JS simulations?
- Can we improve the isolation? e.g. generate s3 signed URL through postMessage API

pyphilia:
We drop this issue because we're not planning to implement it anymore.

PR #556, opened 2023-08-14
refactor: factor out html service for h5p

pyphilia:
In this PR I've extracted an `HtmlService` from the h5p feature. Mainly:
- I moved generic code in `HtmlService` (ie. extracting a zip and uploading its content, deleting a content folder)
- too specific code got moved to `H5PService` (if possible). `H5PService` extends `HtmlService`
- `HtmlService` uses an `HtmlValidator`. `H5PValidator` implements `HtmlValidator`
- Use the general `TMP_FOLDER`

ref #383


pyphilia:
I removed all scorm dependencies, so only the `HtmlService` remained.

src/services/item/plugins/html/h5p/index.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    const h5pStorageRoot = path.join(fileStorage.config.local.storageRootPath, pathPrefix);
+    fs.mkdirSync(h5pStorageRoot, { recursive: true });
+    fastify.register(fastifyStatic, {
+      root: h5pStorageRoot,
+      prefix: routes?.content ?? DEFAULT_H5P_CONTENT_ROUTE,
+      decorateReply: false,
+      setHeaders,
+    });
+  }
```
dialexo, line 124: I guess the exact integration html and the folders to be served are H5P-specific. But in general, wouldn't all html services need a default way to serve the files statically w.r.t to the file service type (i.e. `local`), and ideally the child class should not have to perform the selection? One way to enforce that is to have an abstract function in the parent class `serve` which must be redefined in the children. Then the parent exposes a `handleServe` which performs the local check and calls the subclass's `serve`.

Optionally, it may also want to define a default way to render a templated page (e.g. the parent could "force" the child classes to use moustachejs). This can be done as a default implementation for a method e.g. `render` in the parent class.

src/services/item/plugins/html/h5p/index.ts
```diff
@@ -0,0 +1,225 @@
...
+    with other instances since we use fp to remove the outer scope
+  */
+  await fastify.register(async (fastify) => {
+    fastify.register(fastifyMultipart, {
+      limits: {
+        fields: MAX_NON_FILE_FIELDS,
+        files: MAX_FILES,
+        fileSize: MAX_FILE_SIZE,
+      },
+    });
```
dialexo, line 137: Arguably this could be factored out as well. But probably not in the html service anyway

src/services/item/plugins/html/h5p/service.ts
```diff
@@ -0,0 +1,89 @@
...
+   */
+  downloadH5P(item: H5PItemType, member: Actor, destinationPath: string) {
+    const h5pPath = item.extra.h5p.h5pFilePath;
+    return this.fileService.download(member, {
+      id: item.id,
+      path: path.join(this.pathPrefix, h5pPath),
+      mimetype: this.mimetype,
+      fileStorage: destinationPath,
+    });
+  }
```
dialexo, line 57: Should this be a general feature of the HTML service? i.e. the service coud impose that you can always store and download the original asset / archive ?

src/services/item/plugins/html/h5p/service.ts
```diff
@@ -0,0 +1,89 @@
...
+    // copy .h5p file
+    await this.fileService.copy(actor, {
+      originalPath: path.join(this.pathPrefix, extra.h5p.h5pFilePath),
+      newFilePath: this.buildH5PPath(remoteRootPath, newName),
+    });
+    // copy content folder
+    await this.fileService.copyFolder(actor, {
+      originalFolderPath: path.join(this.pathPrefix, extra.h5p.contentFilePath),
+      newFolderPath: this.buildContentPath(remoteRootPath),
+    });
```
dialexo, line 82: Arguably, copying the assets should be generic as well. But I guess since I wrote above that uploading is not necessarily related to creating an item, then they could be independent as well. Depends on which we chose, but we should stay consistent

src/services/item/plugins/html/service.ts
```diff
@@ -0,0 +1,196 @@
...
+  protected readonly fileService: FileService;
+  protected readonly pathPrefix: string;
+  protected readonly validator: HtmlValidator;
+  protected readonly mimetype: string;
+  protected readonly extension: string;
+
+  protected readonly tempDir: string;
+
+  constructor(
+    fileService: FileService,
```
dialexo, line 37: Maybe this parent class should initialize given the file type?

src/services/item/plugins/html/service.ts
```diff
@@ -0,0 +1,196 @@
...
+    ) => Promise<Item>,
+    parentId?: Item['id'],
+    log?: FastifyBaseLogger,
+  ): Promise<Item> {
+    /*
+      // uppy tries to guess the mime type but fails and falls back to application/octet-stream, so we disable this check for now
+      if (htmlFile.mimetype !== H5P_FILE_MIME_TYPE) {
+        throw new InvalidH5PFileError(htmlFile.mimetype);
+      }
+      */
```
dialexo, line 144: Remove, or provide a way against the default mime type (could be application/octet-stream just to be safe).

src/services/item/plugins/html/service.ts
```diff
@@ -0,0 +1,196 @@
...
+      const contentFolder = this.buildContentPath(targetFolder);
+
+      // save html file
+      await pipeline(htmlFile.file, fs.createWriteStream(savePath));
+      await extract(savePath, { dir: contentFolder });
+
+      const result = await this.validator.validatePackage(contentFolder);
+      if (!result.isValid) {
+        throw new HtmlInvalidFileError((result as { error: string }).error);
+      }
```
dialexo, line 166: Not sure why the cast is required. You could simplify by not returning the errors but throwing them from inside `validatePackage` directly, I think my original API is needlessly complex.

src/services/item/plugins/html/validator.ts
```diff
@@ -0,0 +1,18 @@
...
+   */
+  isExtensionAllowed(extension: string): boolean;
+
+  /**
+   * Validates an extracted html package content
+   */
+  validatePackage(
+    extractedRoot: string,
+    // TODO: type
+  ): Promise<{ isValid: false; error: string } | { isValid: true; manifest?: H5P.Manifest }>;
```
dialexo, line 17: Yeah, we can simplify into a `Promise<T>` where T is generic from the interface depending on the concrete html service type, and instead of returning `isValid, error` we should simply throw from within.

src/utils/config.ts
```diff
@@ -308,7 +307,7 @@ export const STRIPE_DEFAULT_PLAN_PRICE_ID = process.env.STRIPE_DEFAULT_PLAN_PRIC
 // used for hashing password
 export const SALT_ROUNDS = 10;
 
-export const TMP_FOLDER = path.resolve(__dirname, '../..', 'tmp');
+export const TMP_FOLDER = os.tmpdir();
```
dialexo, line 310: Good call, I forgot this existed 😆

dialexo (approved):
Thank you for your work 🌊 I guess it's difficult to define clear boundaries in the system. It's fine by me for now, we should consider improving it in the future (see my comments) by first defining the responsabilities of each part in the architecture.

src/services/item/plugins/html/h5p/index.ts
```diff
@@ -0,0 +1,225 @@
...
+    db,
+    hosts,
+  } = fastify;
+
+  validatePluginOptions(options);
+  const { routes, fileStorage } = options;
+  const { pathPrefix = '' } = fileStorage;
+
+  // todo: file service should be a singleton with mutliple storage drivers
+  const fileService = new FileService(fileStorage.config, fileStorage.type);
```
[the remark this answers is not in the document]
pyphilia: It depends, what if you want to save in different buckets? When implementing scorm, I would maybe want to save it in another bucket `scorm-storage`?

src/services/item/plugins/html/h5p/service.ts
pyphilia: I totally agree, but because of the type and path it's a bit specific. I've opted for a `super.download` and each specific service still have to implement their own download. Let me know what you think. 
Otherwise having an abstract function in the abstract class could also work.

src/services/item/plugins/html/h5p/service.ts
pyphilia: I also agree, but it's difficult to decoupled this function, everything is tight to h5p.

src/services/item/plugins/html/service.ts
pyphilia: Interesting, but then we need to pass the `fileStorage.config`, etc?

src/services/item/plugins/html/service.ts
pyphilia: I don't know it was disabled in the first place? Should I enable it back?

src/services/item/plugins/html/h5p/index.ts
pyphilia: I hear your comment, but would you really put that code in the `service`?  It actually defines endpoints, maybe it makes more sense to create a sub-plugin `localHtmlPluginController`?

src/services/item/plugins/html/h5p/index.ts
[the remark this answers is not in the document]
dialexo: Yes I don't mean using a single storage system. I think however that the initialization and storage access logic should be provided by the parent class, wdyt?

src/services/item/plugins/html/h5p/index.ts
dialexo: True, i'm not sure about the exact boundaries here. But you could imagine that the html services should expose a `serve` method, which all html controllers should call (this can be enforced via an abstract parent html controller for instance)

src/services/item/plugins/html/service.ts
dialexo: Not sure about the behvaiour of uppy: does it always provide the `application/octet-stream` mime type? I guess it's a minor validation anyway, in the sense that it would help us eliminate requests that attempt to upload unwanted filetypes, but it does not really validate if the file really matches the mimetype anyway.

src/services/item/plugins/html/h5p/index.ts
pyphilia: I think I need an example here^^'

src/services/item/plugins/html/h5p/index.ts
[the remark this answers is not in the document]
pyphilia: I think I see now. I'll change that.

pyphilia:
I've moved the file service instanciation in the h5p service.

But the serve thingy I tried but I don't think it makes sense to put it in a separated plugin but for now I think it's okay to let it in the plugin. 

Let me know what you think.

.github/workflows/test.yml
```diff
@@ -196,7 +196,8 @@ jobs:
           H5P_FILE_STORAGE_TYPE: file
           H5P_PATH_PREFIX: h5p-content/
           H5P_STORAGE_ROOT_PATH: /tmp/h5p
-          H5P_TEMP_DIR: /tmp/h5p/extract
+          # not used anymore
+          H5P_TEMP_DIR: /tmp/h5p/extract
```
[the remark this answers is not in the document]
dialexo: > Should you remove it?

Bumping this

dialexo (approved):
A few nitpicks left but I'll let you decide if these changes are needed 👍

src/services/item/plugins/html/h5p/service.ts
```diff
@@ -0,0 +1,99 @@
...
+        contentFilePath: this.buildContentPath(contentId),
+      },
+    };
+  }
+
+  /**
+   * Helper to build the local or remote path of the .h5p file
+   */
+  buildH5PPath = (rootPath: string, filename: string) =>
+    path.join(rootPath, `${filename}.${H5P.H5P_FILE_EXTENSION}`);
```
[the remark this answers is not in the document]
pyphilia: There's no specific reason, I like it when it's scoped? It feels like putting it in a drawer aha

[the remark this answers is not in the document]
dialexo: WDYM by scoped?

dialexo (approved):

src/services/item/plugins/html/h5p/service.ts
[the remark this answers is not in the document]
pyphilia: It's not just a random function hidden outside de plugin (or inside), but it's defined in the service. It was not about the scope in programmation.

PR #557, opened 2023-08-15
refactor: use MAX_TARGETS_FOR_READ_REQUEST for GET item tag

### LINKED ISSUE #548: Factor out max number of item tags one can fetch
https://github.com/graasp/graasp/blob/45c6b95095f9c219ea2952a449447a3876565c56/src/services/item/plugins/itemTag/schemas.ts#L5

Factor out this line/use the same as for item?

pyphilia:
I've used the same constant for get item tag.
I've also modify a little bit an error in s3 (to help debugging).

close #548

### DIFF
--- a/src/services/item/plugins/itemTag/schemas.ts
+++ b/src/services/item/plugins/itemTag/schemas.ts
@@ -1,8 +1,6 @@
-import { ItemTagType } from '@graasp/sdk';
+import { ItemTagType, MAX_TARGETS_FOR_READ_REQUEST } from '@graasp/sdk';
 
 import { UUID_REGEX } from '../../../../schemas/global';
-
-const MAX_ITEMS_FOR_GET = 30;
 
 export default {
   $id: 'http://graasp.org/item-tags/',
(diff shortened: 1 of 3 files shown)


src/services/item/plugins/itemTag/schemas.ts
```diff
@@ -97,7 +95,10 @@ const getMany = {
   querystring: {
     allOf: [
       { $ref: 'http://graasp.org/#/definitions/idsQuery' },
-      { type: 'object', properties: { id: { type: 'array', maxItems: MAX_ITEMS_FOR_GET } } },
+      {
+        type: 'object',
+        properties: { id: { type: 'array', maxItems: MAX_TARGETS_FOR_READ_REQUEST } },
```
dialexo, line 100: You have swapped the constant, but the previous was not removed? What is the difference between the two?

dialexo (changes requested):

src/services/item/plugins/itemTag/schemas.ts
pyphilia: I did remove it? The difference is the number (15 instead of 30 now), but the important point is that it's synced with the constant used for GET /items as well. I needed a constant to refer in `query-client` to correctly split it.

dialexo: Thank you, my bad I missed the line

dialexo (approved):

PR #559, opened 2023-08-15
fix: remove sentry profiling to resolve types

dialexo:
Sentry somehow broke the types between profiling-node and the the rest of the sentry packages.
We remove profiling for now since we had disabled it. TODO see when / if we re-enable it later.


spaenleh (approved):
🧄

PR #565, opened 2023-08-18
fix: use mailer in itemMembership service, fix notification

### LINKED ISSUE #564: [membership] notification email does not work
In `itemMembership/index.ts` we don't pass the right parameter `mailer` because `hosts` is `any`.

pyphilia:
close #564

### DIFF
--- a/src/declarations.d.ts
+++ b/src/declarations.d.ts
@@ -59,8 +59,6 @@ declare module 'fastify' {
     actions: { service: ActionService };
     chat: { service: ChatMessageService };
 
-    hosts: Hostname[];
-
     websockets: WebsocketService;
     h5p: H5PService;
     etherpad: EtherpadItemService;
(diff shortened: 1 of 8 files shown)


src/declarations.d.ts
dialexo, line 1: I've actually removed the `hosts` decoration in #471, I don't think it makes much sense to decorate it at runtime rather than consuming it directly from the constants file. Not sure why I forgot to remove it from the declarations, but I think you should remove the `hosts` property altogether.

dialexo (changes requested):
The change in `declarations.d.ts` is not needed, but you can actually remove the dangling legacy property entirely.

src/declarations.d.ts
pyphilia: Good to know, I was not sure if it was actually used. I've removed it everywhere, the only "big" change is in h5p let me know if it looks correct.

dialexo (approved):
LGTM 🍄

PR #568, opened 2023-08-21
fix: export a zip use stream

### LINKED ISSUE #516: [Export-zip] Export empty folders

### LINKED ISSUE #537: [export-zip] No space left on device
Possible workaround: https://www.npmjs.com/package/yauzl

pyphilia:
This PR refactor `/zip-export` to use `yazl`. This library is supposed to handle streams, so we don't need to download all the files temporarly.

Currently h5p still saves its own temporary file. But this can be handled in a refactor for the download method of `FileService`.

close #537 
close #516

### DIFF
--- a/src/services/item/plugins/importExport/service.ts
+++ b/src/services/item/plugins/importExport/service.ts
@@ -1,10 +1,11 @@
-import archiver, { Archiver } from 'archiver';
 import fs, { existsSync } from 'fs';
 import { mkdir, readFile } from 'fs/promises';
 import mime from 'mime-types';
 import mmm from 'mmmagic';
+import fetch from 'node-fetch';
 import path from 'path';
 import util from 'util';
+import yazl, { ZipFile } from 'yazl';
 
 import { FastifyBaseLogger, FastifyReply } from 'fastify';
 
@@ -200,7 +201,7 @@ export class ImportExportService {
       reply;
       item: Item;
       archiveRootPath: string;
-      archive: Archiver;
+      archive: ZipFile;
       fileStorage: string;
     },
   ) {
@@ -208,9 +209,10 @@ export class ImportExportService {
 
     // save description in file
     if (item.description) {
-      archive.append(item.description, {
-        name: path.join(archiveRootPath, `${item.name}${DESCRIPTION_EXTENSION}`),
-      });
+      archive.addBuffer(
+        Buffer.from(item.description),
+        path.join(archiveRootPath, `${item.name}${DESCRIPTION_EXTENSION}`),
+      );
     }
 
     switch (item.type) {
@@ -220,10 +222,9 @@ export class ImportExportService {
         const { mimetype } =
           (item.extra[ItemType.S3_FILE] as S3FileItemExtra) ||
           (item.extra[ItemType.LOCAL_FILE] as LocalFileItemExtra);
-        const fileStream = await this.fileItemService.download(actor, repositories, {
-          fileStorage,
+        const url = (await this.fileItemService.download(actor, repositories, {
           itemId: item.id,
-        });
+        })) as string;
 
         // build filename with extension if does not exist
         let ext = path.extname(item.name);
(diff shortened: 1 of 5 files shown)


src/services/item/plugins/importExport/service.ts
```diff
@@ -220,10 +222,9 @@ export class ImportExportService {
         const { mimetype } =
           (item.extra[ItemType.S3_FILE] as S3FileItemExtra) ||
           (item.extra[ItemType.LOCAL_FILE] as LocalFileItemExtra);
-        const fileStream = await this.fileItemService.download(actor, repositories, {
-          fileStorage,
+        const url = (await this.fileItemService.download(actor, repositories, {
           itemId: item.id,
-        });
+        })) as string;
```
pyphilia, line 227: Casting necessary because the method `download` is quite a mess. To be handled in another PR.

src/services/item/plugins/importExport/service.ts
```diff
@@ -200,17 +201,18 @@ export class ImportExportService {
       reply;
       item: Item;
```
dialexo, line 202: Is it possible to type it here as a `DiscriminatedItem`? This would allow the extras to be typed correctly below without the need to cast them specifically

dialexo (changes requested):

src/services/item/plugins/importExport/service.ts
pyphilia: I will do the change, but it does not fit well with the Item repository. For example get an item return type is `Item` which does not fit with `DiscriminatedItem`. Changing this would result in changes everywhere in the project. 

So I've opted for casting to `Item` instead, which is indeed better.

pyphilia:
I did the changes you requested! I moved some bunch of code to the controller, which is quite nice I think. 
Only regret is the `fileStorage` I don't think I can delete unless the stream is sent (so I had to "hack" around). But this will disappear once we refactor the h5p/file service.

src/services/item/plugins/importExport/service.ts
```diff
@@ -300,37 +288,31 @@ export class ImportExportService {
...
-    { itemId, reply }: { itemId: UUID; reply: FastifyReply },
+    {
+      item,
+      reply,
+      fileStorage,
+    }: { fileStorage: string; item: Item; reply: FastifyReply; log?: FastifyBaseLogger },
   ) {
-    // check item and permission
-    const item = await this.itemService.get(actor, repositories, itemId);
-
```
dialexo, line 307: Is the check not required anymore or was it moved?

src/services/item/plugins/importExport/service.ts
```diff
@@ -300,37 +288,31 @@ export class ImportExportService {
...
     // init archive
-    const archive = archiver.create('zip', { store: true });
-    archive.on('warning', function (err) {
+    const archive = new yazl.ZipFile();
+    archive.outputStream.on('warning', function (err) {
       if (err.code === 'ENOENT') {
         // log.debug(err);
       } else {
         throw err;
       }
```
dialexo, line 304: Not sure what error is handled here?

src/services/item/plugins/importExport/service.ts
```diff
@@ -300,37 +288,31 @@ export class ImportExportService {
...
     });
-    archive.on('error', function (err) {
+    archive.outputStream.on('error', function (err) {
       throw err;
     });
```
dialexo, line 308: Is this useful?

dialexo (approved):
I have a few questions but otherwise LGTM 🥮

src/services/item/plugins/importExport/service.ts
pyphilia: It moved to the controller (I needed the name, it's also better to check the permission before creating the filestorage).

src/services/item/plugins/importExport/service.ts
pyphilia: Not sure, but I think it doesn't hurt?

dialexo (approved):

PR #576, opened 2023-08-23
feat: implement get current member storage

### LINKED ISSUE #575: Implement GET current member's storage

pyphilia:
Created a `StorageService` to contain functions related to storage.
Kept in mind the maximum storage can be different for each user.

close #575

### DIFF
--- a/src/services/member/plugins/storage/service.ts
+++ b/src/services/member/plugins/storage/service.ts
@@ -1,4 +1,4 @@
-import { FileItemType } from '@graasp/sdk';
+import { FileItemType, MemberStorage } from '@graasp/sdk';
 
 import { Repositories } from '../../../../utils/repositories';
 import { DEFAULT_MAX_STORAGE } from '../../../item/plugins/file/utils/constants';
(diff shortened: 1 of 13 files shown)


src/services/member/plugins/storage/service.ts
```diff
@@ -0,0 +1,39 @@
...
+import { Repositories } from '../../../../utils/repositories';
+import { DEFAULT_MAX_STORAGE } from '../../../item/plugins/file/utils/constants';
+import { StorageExceeded } from '../../../item/plugins/file/utils/errors';
+import { Member } from '../../entities/member';
+
+export class StorageService {
+  getMaximumStorageSize(_actor: Member): number {
+    // todo: depends on user/group
+    return DEFAULT_MAX_STORAGE;
+  }
```
dialexo, line 12: You may want to mark this async already for future-proofing

src/services/member/plugins/storage/service.ts
```diff
@@ -0,0 +1,39 @@
...
+      maximum: this.getMaximumStorageSize(actor),
+    };
+  }
+
+  // check the user has enough storage to create a new item given its size
+  // get the complete storage
+  // todo: include more item types
+  async checkRemainingStorage(
+    actor: Member,
+    type: FileItemType,
```
dialexo, line 26: Is it necessary to pass the item type to the storage service? Shouldn't it be agnostic and only know about sizes?

dialexo (approved):
Overall LGTM, however have you considered using the [Decorator pattern](https://refactoring.guru/design-patterns/decorator) on top of the file service? Pros: it would ensure that all consumers go through the decoration to handle anything that could affect the user's storage (e.g. we could mark the origin file service method as deprecated), cons: it is less flexible and can become a bloated class if we have many other services that need to set constraints through this service. WDYT @pyphilia @morganridel ?

dialexo (approved):

PR #580, opened 2023-08-25
fix(internal): better type safety

### LINKED ISSUE #535: Seed folder extra with empty array for childrenOrder
Having created a couple learning capsules on Builder, every time I update an item, the order changes, so I have to reorder.

To keep with the default behavior on the old Graasp, it might be best to sort by default by creation date.

spaenleh:
This PR should improve type-safety by defining typeguards and a generic Item Entity.

Detail:
- Removed link between sdk types and database Entities. Database types are now independent, the link should be done by a mapping at the service level make the exposed API responses match the types defined in the sdk (public types). For the moment these types match without mapping, but we should introduce one at some point.
- Changed `Item` entity to be generic over `ItemType` and improved the link between `type` and `extra` by using the generic and a mapping from type to extra. This means that by default the itemRepository will return `Item<ItemTypeKeys>` which can be narrowed using provided type-guards (see next bullet) to one of the specific item types with narrowed extra. Pretty sweet ! 🍬 
- Added some type-guard functions to help narrow item types. This was the only solution available as discriminated unions did not work as expected because the Item generic would not narrow to a particular enum value but the whole enum instead...
- Also seed the folder extra with the childrenOrder to fix #535 
- Removed some unused variables
- Miscellaneous cleanup

Hope this improves the developer experience overall !


package.json
```diff
@@ -66,7 +66,7 @@
     "@fastify/view": "7.4.1",
     "@fastify/websocket": "7.2.0",
     "@graasp/etherpad-api": "2.1.1",
-    "@graasp/sdk": "1.2.0",
+    "@graasp/sdk": "github:graasp/graasp-sdk#251-fix-use-discriminateditem-instead-of-item-in-types",
```
spaenleh, line 69: TODO: use merged branch from sdk

src/services/item/plugins/app/test/index.test.ts
```diff
@@ -221,7 +221,7 @@ describe('Apps Plugin Tests', () => {
             Authorization: `Bearer ${MOCK_TOKEN}`,
           },
         });
-        const result = response.json();
+        const result = await response.json();
```
spaenleh, line 224: Not sure we need this ?

pyphilia: I don't think you need it.

spaenleh:
Related to the `childrenOrder`:
- [x] seed the extra on creation
- [x] adapt the tests to expect the filled extra
- [ ] need to add a hook to update the parent order on child creation

src/services/item/entities/Item.ts
```diff
@@ -100,3 +107,46 @@ export class Item extends BaseEntity implements GraaspItem {
...
+  item.type === ItemType.FOLDER;
+export const isH5PItem = (item: Item<ItemTypeKeys>): item is H5PItem => item.type === ItemType.H5P;
+export const isEmbeddedLinkItem = (item: Item<ItemTypeKeys>): item is EmbeddedLinkItem =>
+  item.type === ItemType.LINK;
+export const isLocalFileItem = (item: Item<ItemTypeKeys>): item is LocalFileItem =>
+  item.type === ItemType.LOCAL_FILE;
+export const isS3FileItem = (item: Item<ItemTypeKeys>): item is S3FileItem =>
+  item.type === ItemType.S3_FILE;
+export const isShortcutItem = (item: Item<ItemTypeKeys>): item is ShortcutItem =>
+  item.type === ItemType.SHORTCUT;
```
dialexo, line 152: Are these type guards necessary? I thought the change would allow the typescript compiler to automatically pickup the extra given the disciminated union on `item.type`? IMO type guards are in essence equivalent to casts, since we basically tell the compiler "trust me bro, if this check passes then we promise it's the `is` type".

I.e. does this mean that writing the following in the code somewhere does not provide the appropriate extra typings (as it does in the front-end)?
```ts
if (item.type === ItemType.DOCUMENT) {
    typeof item.extra // <-- should be DocumentItemExtra but is ?
}
```

When typescript recognizes a type union you shouldn't need typeguards: see https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types

If we still need these typeguards, I think we did not gain additional safety, i.e. we could have written the typeguard before as well and from the type system guarantees they would be the same (either previously the casts, or now the type guards, provide the actual "guarantee"), unlike in the SDK where we currently have "true" union type discrimination.

If we don't, then you should drop them as they may introduce type errors in the future. Consumers should be able to expect that checking against `item.type` correctly narrows the extra type.

src/services/item/plugins/app/appAction/service.ts
```diff
@@ -1,9 +1,10 @@
-import { AppAction, PermissionLevel } from '@graasp/sdk';
+import { PermissionLevel } from '@graasp/sdk';
 
 import HookManager from '../../../../../utils/hook';
 import { Repositories } from '../../../../../utils/repositories';
 import { validatePermission } from '../../../../authorization';
 import { ManyItemsGetFilter, SingleItemGetFilter } from '../interfaces/request';
+import { AppAction } from './appAction';
```
dialexo, line 7: Not sure about this, shouldn't the service layer deal with the business-level types from SDK?

For instance in the lines 12-17 below:
```ts
hooks = new HookManager<{
    post: {
      pre: { appAction: Partial<InputAppAction>; itemId: string };
      post: { appAction: AppAction; itemId: string };
    };
  }>();
```

as well as lines 31-36, I think no conversion is involved so the types are still entity-level:
```ts
const appAction = await appActionRepository.post(itemId, actorId, body);
    await this.hooks.runPostHooks('post', member, repositories, {
      appAction,
      itemId,
    });
    return appAction;
```

I think my comment applies generally to all services w.r.t to their underlying entities.

src/services/item/plugins/app/service.ts
```diff
@@ -1,16 +1,19 @@
...
-import { Item } from '../../entities/Item';
+import { Item, isAppItem } from '../../entities/Item';
 import { checkTargetItemAndTokenItemMatch } from './utils';
 
 export class AppService {
   jwtExpiration: number;
-  promisifiedJwtSign: any; // TODO
+  promisifiedJwtSign: (
+    arg1: { sub: AuthTokenSubject },
+    arg2: { expiresIn: string },
```
dialexo, line 15: Should these parameters have more descriptive names? E.g. the original parameter names from `jwtSign`

src/services/item/plugins/app/service.ts
```diff
@@ -61,8 +64,10 @@ export class AppService {
   async getContext(actorId, repositories: Repositories, itemId: string, requestDetails) {
     const { itemRepository, memberRepository } = repositories;
 
-    // TODO: check item is app
     const item = await itemRepository.get(itemId);
+    if (!isAppItem(item)) {
+      throw new Error('Item is not an app');
+    }
```
dialexo, line 70: Same as pointed above, this should be equivalent without changes to `item.type === ItemType.APP`. Is that the case?

src/services/item/plugins/app/service.ts
```diff
@@ -94,7 +99,12 @@ export class AppService {
   }
 
   // used by app : get tree
-  async getItemsByType(actorId: UUID, repositories: Repositories, types: ItemType[], path: string) {
+  async getItemsByType(
+    actorId: UUID,
+    repositories: Repositories,
+    types: UnionOfConst<typeof ItemType>[],
```
dialexo, line 105: Info: what does this type change do?

src/services/item/plugins/app/ws/events.ts
```diff
@@ -1,9 +1,9 @@
 /**
  * App websocket events are registered under these topics
  */
-import { AppAction, AppSetting } from '@graasp/sdk';
-
+import { AppAction } from '../appAction/appAction';
 import { AppData } from '../appData/appData';
+import { AppSetting } from '../appSetting/appSettings';
```
dialexo, line 6: Same here, shouldn't these types be business-logic related hence from SDK?

src/services/item/plugins/embeddedLink/index.ts
```diff
@@ -2,11 +2,11 @@ import fetch from 'node-fetch';
 
 import { FastifyPluginAsync } from 'fastify';
 
-import { EmbeddedLinkItemExtra, ItemType } from '@graasp/sdk';
+import { ItemType } from '@graasp/sdk';
 
 import { Repositories } from '../../../../utils/repositories';
 import { Actor } from '../../../member/entities/member';
-import { Item } from '../../entities/Item';
+import { Item, isEmbeddedLinkItem } from '../../entities/Item';
```
dialexo, line 9: Is this import not used?

src/services/item/plugins/embeddedLink/index.ts
```diff
@@ -46,11 +46,11 @@ const plugin: FastifyPluginAsync<GraaspEmbeddedLinkItemOptions> = async (fastify
...
     { item }: { item: Partial<Item> },
   ) => {
-    const { embeddedLink } = (item?.extra as EmbeddedLinkItemExtra) ?? {};
-
-    if (item.type !== ItemType.LINK || !embeddedLink) {
+    // if the extra is undefined or it does not contain the embedded link extra key, exit
+    if (!item.extra || !(ItemType.LINK in item.extra)) {
       return;
     }
+    const { embeddedLink } = item.extra;
```
dialexo, line 53: Does this mean that this actually type-checks without the need for the typeguards? Good news if that's the case, it would mean the type guards are not needed.

src/services/item/plugins/etherpad/service.ts
```diff
@@ -3,18 +3,13 @@ import { DateTime } from 'luxon';
...
-  PermissionLevel,
-} from '@graasp/sdk';
+import { EtherpadItemExtra, ItemType, PermissionLevel } from '@graasp/sdk';
 
 import { MemberCannotWriteItem } from '../../../../utils/errors';
 import { buildRepositories } from '../../../../utils/repositories';
 import { validatePermission } from '../../../authorization';
 import { Member } from '../../../member/entities/member';
-import { Item } from '../../entities/Item';
+import { Item, isEtherpadItem } from '../../entities/Item';
```
dialexo, line 12: Same question about service-level types

src/services/item/plugins/etherpad/service.ts
```diff
@@ -151,10 +146,10 @@ export class EtherpadItemService {
 
     const checkedMode = await this.checkMode(mode, member, item);
 
-    if (!item.extra?.etherpad) {
+    if (!isEtherpadItem(item) || !item.extra?.etherpad) {
```
dialexo, line 149: Same question about whether type guards are required

src/services/item/plugins/file/service.ts
```diff
@@ -130,20 +122,17 @@ class FileItemService {
...
-            size,
-          },
-          // todo: fix type
-        } as any,
-        parentId,
+          // this is needed because if we directly use `this.fileService.type` then TS widens the type to `string` which we do not want
+          ...(this.fileService.type === ItemType.LOCAL_FILE
+            ? { [ItemType.LOCAL_FILE]: fileProperties }
+            : { [ItemType.S3_FILE]: fileProperties }),
+        } as ItemExtraMap[typeof this.fileService.type],
```
dialexo, line 135: Can't we restrict the type of  `fileService.type` instead?

src/services/item/plugins/importExport/service.ts
```diff
@@ -9,13 +9,22 @@ import yazl, { ZipFile } from 'yazl';
...
+import {
+  Item,
+  isAppItem,
+  isDocumentItem,
+  isEmbeddedLinkItem,
+  isFolderItem,
+  isH5PItem,
+  isLocalFileItem,
+  isS3FileItem,
+} from '../../entities/Item';
```
dialexo, line 27: Same question about business-logic types

src/services/item/plugins/importExport/service.ts
```diff
@@ -205,81 +218,73 @@ export class ImportExportService {
...
-          ext = `.${mime.extension(mimetype)}`;
-        }
-        const filename = `${path.basename(item.name, ext)}${ext}`;
-
-        // add file in archive
-        const res = await fetch(url);
-        archive.addReadStream(res.body, path.join(archiveRootPath, filename));
-
-        break;
+    if (isLocalFileItem(item) || isS3FileItem(item)) {
```
dialexo, line 221: Same question about type guards (in all conditions below as well).

src/services/item/plugins/validation/service.ts
```diff
@@ -18,7 +16,7 @@ import { Repositories } from '../../../../utils/repositories';
 import { validatePermission } from '../../../authorization';
 import FileService from '../../../file/service';
 import { Member } from '../../../member/entities/member';
-import { Item } from '../../entities/Item';
+import { Item, isLocalFileItem, isS3FileItem } from '../../entities/Item';
```
dialexo, line 19: Same question about service types

src/services/item/plugins/validation/service.ts
```diff
@@ -162,40 +160,30 @@ export class ItemValidationService {
...
 
         case ItemValidationProcess.ImageChecking:
-          let filepath = '';
-          let mimetype = '';
-          // check for service type and assign filepath, mimetype respectively
-          if (item?.type === ItemType.S3_FILE) {
-            const s3Extra = item?.extra as S3FileItemExtra;
-            filepath = s3Extra?.s3File?.path;
-            mimetype = s3Extra?.s3File?.mimetype;
+          if (isS3FileItem(item) || isLocalFileItem(item)) {
```
dialexo, line 163: Same question about typeguards

src/services/item/repository.ts
```diff
@@ -78,9 +78,14 @@ export const ItemRepository = AppDataSource.getRepository(Item).extend({
...
     // TODO: extra
     // folder's extra can be empty
-    const parsedExtra = extra ? JSON.parse(JSON.stringify(extra)) : ({} as ItemExtra);
+    let parsedExtra: ItemExtraUnion = extra ? JSON.parse(JSON.stringify(extra)) : {};
     const id = v4();
 
+    // if item is a folder and the extra is empty, seed the childrenOrder
+    if (type === ItemType.FOLDER && !(ItemType.FOLDER in parsedExtra)) {
+      parsedExtra = { folder: { childrenOrder: [] } };
+    }
```
dialexo, line 87: I believe this also shows that the type-guards are not needed?

src/services/item/repository.ts
```diff
@@ -132,7 +137,7 @@ export const ItemRepository = AppDataSource.getRepository(Item).extend({
   },
 
   async getChildren(parent: Item, ordered?: boolean): Promise<Item[]> {
-    if (parent.type !== ItemType.FOLDER) {
+    if (!isFolderItem(parent)) {
```
dialexo, line 140: Same as above

src/services/item/service.ts
```diff
@@ -23,7 +22,7 @@ import { Repositories } from '../../utils/repositories';
 import { filterOutItems, validatePermission, validatePermissionMany } from '../authorization';
 import { Actor, Member } from '../member/entities/member';
 import { mapById } from '../utils';
-import { Item } from './entities/Item';
+import { Item, isFolderItem } from './entities/Item';
```
dialexo, line 25: Same as above

src/services/item/utils.ts
```diff
@@ -1,6 +1,6 @@
-import { FolderItemExtra, ItemType, UUID } from '@graasp/sdk';
+import { UUID } from '@graasp/sdk';
 
-import { Item } from './entities/Item';
+import { Item, isFolderItem } from './entities/Item';
```
dialexo, line 3: Same question about service types

dialexo (changes requested):
Thank you for your huge work :clap: 

I am not convinced with the usage of type guards. Either we can type-check with `item.type === ...` and we should use those to narrow the types and we remove them, or we did not actually improve the type safety of our extras, i.e. we just moved the cast from explicit to implicit into type guards. But from your code, it seems the former is true in which case I would advice removing the type guards entirely.

It is also unclear how the types should be distributed. They seem to be replaced everywhere now, but IMO the entity types should stay confined to the repository layer. The service layer should only deal with service-level types, i.e. SDK. WDYT @spaenleh @pyphilia @morganridel ?

spaenleh:
Thank you for your review, @dialexo. Maybe I need to give more background on what I tried and why it did not work, so we can understand why the proposed implementation (using type-guards) may not be ideal but was - in my research - the only way to increase the type-safety and convenience when coding. 

## Current issue
Typeorm uses classes for its entities, so we can not directly represent our items with a discriminated union like we do in the frontend. 
## Solutions I tried
### Single table inheritance
I define a parent entity with the common fields of an Item and then define a child entity for each specific type. This is better suited for scenarios where you know you want to query the child entities directly and the parent is just a way of being DRY. 

### Generic entity class and sprinkled discriminated union
One "simple way" of solving the class issue is to make the class generic. The generic parameter should be a string literal that would describe the `type` of the item. Then you can use a mapping from `type` to `extra`, so you can link both together. This works great when you instantiate the Item with a literal value:

```ts
class Item<T extends "folder" | "app"> {
	name: string;
	type: T;
	extra: MappingFromTypeToExtra[T];
	constructor(name: string, type: T, extra: MappingFromTypeToExtra[T]) {
		this.name = name;
		this.type = type;
		this.extra = extra;
	}
};

const folderItem = new Item("New Folder", "folder", {folder: {some: "extra"}}); // this will be a Item<"folder">

const appItem = new Item("Awesome App", "app", {app: {url: "http:..."}}) // this will be a Item<"app">

// checking the `type` will correctly narrow the extra 
if (folderItem.type === "folder") {
	// inside there the type is correctly narrowed
	console.log(folderItem.extra) // extra will be of type FolderExtra i.e. {folder: {some: "extra"}}
}

```
But this code does not work when used in Typeorm. The issue is that items created by typeorm are still if the generic type, and not one particular one. 

```ts
type AppItem = Item<"app">;
const oneItem = itemRepository.get("123");
// oneItem will be of type Item<"folder" | "app">

const appItem: AppItem = itemRepository.get("234"); // this line will fail, you can not assign Item<"folder" | "app"> to AppItem
```

In short, TS can not know at compile time what the value of `type` will be to make it specific. The generic is not powerful enough to hint the type system in the right direction. In a way this makes sens, the Item has a generic type because it could be "anything". We keep the link with our mapping from type to extra, but this does not seem to be enough.
## Current solution
The current solution keeps the idea of the generic class with the mapping from type to extra, but adds type-guards to "help" typescript narrow the type correctly. This means that we "know" that if the type if "folder", then the extra will be of type `FolderExtra`. But as this can only be verified at runtime, using type-guards lets us have that narrowing in the code with the typings, allowing us to have a richer coding experience.
### Pros
- If we use the type guards in place of the `item.type === ItemType.SOME_TYPE` we can narrow the type of the extra. This works really well.
- Type-guards are defined in one place (close to the Item entity) and they should not be changed unless we add a new ItemType.
### Cons
- We are essentially telling the compiler "trust me bro"

I think the type-guards in this case are really straight-forward since they just expose our mapping to TS explicitly. 

I understand the concern that type-guards may not be the most beautiful thing in the world, but it was the only way of making this work. So we can either:
- roll with this and accept the use of type-guards to try to increase type-safety in our code
- reject it (I would not mind) and be back to blindly manipulating extras without any form of specific typing

spaenleh:
To address your second comment, yes the separation between service layer and repository layer is still blurry. I chose to not make it clearer in this PR as there was already a lot to be changed. I would like to discuss more in depth the approach we would use to do the mapping between the layer with you @dialexo  @morganridel  and @pyphilia

src/services/item/entities/Item.ts
spaenleh: You are correct, checking `item.type === "folder"` does not narrow the generic as outlined in [this SO post](https://stackoverflow.com/questions/68897217/types-narrowing-with-generics). This really sucks in my opinion as I would have like for it to work in the described way. 

The issue is that we can not define a union type at the level of the class ...

src/services/item/plugins/app/appAction/service.ts
spaenleh: Yes we currently use the entity level types throughout the backend, as no mapping to service types was created in this PR.

src/services/item/plugins/app/service.ts
spaenleh: Checking the item.type like before does not narrow. This is why type-guards are required.

src/services/item/plugins/app/service.ts
spaenleh: Previously `ItemType` was an `enum`. I changed it to be a const object which means that to obtain the union of literal values before we did 
```ts
`${ItemType}
```
but now we need to do:
```ts
(typeof ItemType)[keyof typeof ItemType]
```
`UnionOfConst<T>` is just a generic utility type I defined in sdk to help with this conversion.

src/services/item/plugins/embeddedLink/index.ts
spaenleh: It works here because we have 
```ts
// if the extra is undefined or it does not contain the embedded link extra key, exit
if (!item.extra || !(ItemType.LINK in item.extra)) {
  return;
}
```
above

src/services/item/plugins/etherpad/service.ts
spaenleh: Yes here they are required, other wise it says that `etherpad` does not exist on type `item.extra`

src/services/item/plugins/file/service.ts
spaenleh: This one was a tricky one too. You can't use `fileService.type` as the key here because it the union of literal `"s3File" | "localFile"`. So it will not result in the expected type of FileItemProperties ... The only way to make TS abid to the fact that it was indeed one or the other was by explicitly checking for the value and setting the value again explicitly like I did here with the literal.

dialexo:
Quick note to summarize our conversation, the issue is that there is not native support in TypeORM for discriminated unions. So we have to "hack" around the type system so that it behaves similarly. In particular, while we can store a discriminated union into a structurally subtyped equivalent class, it seems there is no way to perform the reverse mapping from a serialized DB entity into a discriminated union (whether of types or classes). So we have to perform some kind of cast in this direction: the current implementation (ab)uses generics so that the mapping between item.type and item.extra is recognized by the type system, and the conversion is implemented through a generic type guard. Unfortunately, this type guard cannot be an instance method of the entity class.

morganridel:
I think the current frontend isn't any more typesafe than the backend with casting from the entity.

As long as you have your discriminated object, of course you can narrow it, but you do have it in the frontend because you force casted the result from the network API (your "trust me bro" moment).

In the backend, you force cast also from the network (with the database) as well, but since you are supposed to insert in your database from "type-safe" types created on the business logic layer, you should be able to trust the cast when the see the type of your item.

Haven't read the code yet though

PR #590, opened 2023-08-31
fix: align file name truncation with item name limit

spaenleh:
This PR aligns the truncation length of uploaded file names to match the item name limit. The limit is now `500`.

### DIFF
--- a/src/services/item/plugins/file/service.ts
+++ b/src/services/item/plugins/file/service.ts
@@ -12,11 +12,10 @@ import FileService from '../../../file/service';
 import { UploadEmptyFileError } from '../../../file/utils/errors';
 import { Actor, Member } from '../../../member/entities/member';
 import { randomHexOf4 } from '../../../utils';
+import { MAX_ITEM_NAME_LENGTH } from '../../constants';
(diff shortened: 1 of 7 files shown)


src/services/item/plugins/file/service.ts
```diff
@@ -1,31 +1,22 @@
...
 import { UploadEmptyFileError } from '../../../file/utils/errors';
 import { Actor, Member } from '../../../member/entities/member';
 import { randomHexOf4 } from '../../../utils';
-import { Item } from '../../entities/Item';
 import ItemService from '../../service';
 import { ItemThumbnailService } from '../thumbnail/service';
 import { StorageExceeded } from './utils/errors';
 
-const ORIGINAL_FILENAME_TRUNCATE_LIMIT = 20;
+const ORIGINAL_FILENAME_TRUNCATE_LIMIT = 500;
```
dialexo, line 19: Why don't we use the same constant for both use cases?

dialexo (approved):
LGTM but if we want to always keep them in sync, we may want to use the same constant
🫕

dialexo (approved):

PR #595, opened 2023-09-04
fix: fix library link in email

### LINKED ISSUE #594: Published email link is broken
`https://undefined/collections/<item-id>`

pyphilia:
The link was broken because we provided a `URL` object in another `URL` object.

fix #594

### DIFF
--- a/src/services/item/plugins/published/constants.ts
+++ b/src/services/item/plugins/published/constants.ts
@@ -3,8 +3,12 @@ import { Context } from '@graasp/sdk';
 import { CLIENT_HOSTS } from '../../../../utils/config';
 import { Item } from '../../entities/Item';
 
-export const buildPublishedItemLink = (item: Item) => {
-  const library = CLIENT_HOSTS.find(({ name }) => name === Context.Library);
-  const target = new URL(item.id, library?.url);
+export const buildPublishedItemLink = (item: Item): string | null => {
+  const library = CLIENT_HOSTS.find(({ name }) => name === Context.Library)?.url;
+  if (!library) {
+    return null;
+  }
+
+  const target = new URL(`/${item.id}`, library.origin);
   return target.toString();
 };
--- a/src/services/item/plugins/published/constants.ts
+++ b/src/services/item/plugins/published/constants.ts
@@ -1,14 +1,7 @@
-import { Context } from '@graasp/sdk';
-
-import { CLIENT_HOSTS } from '../../../../utils/config';
+import { LIBRARY_HOST } from '../../../../utils/config';
 import { Item } from '../../entities/Item';
 
-export const buildPublishedItemLink = (item: Item): string | null => {
-  const library = CLIENT_HOSTS.find(({ name }) => name === Context.Library)?.url;
-  if (!library) {
-    return null;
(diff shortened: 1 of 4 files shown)


dialexo (changes requested):
I disagree, the bug is unrelated to nesting URL objects. See [this example](https://www.typescriptlang.org/play?#code/MYewdgzgLgBArgJwDYwLwzAUwO4wKoBKAMgBQDkAFlFAA4QBcA9IwOYICG7ENAdCAizIBKAFChIsACaZoaDDnzFyjGuygUyAGnjIhQA), Typescript explicitly provides the following type for the base (i.e. second) parameter: 

```ts
declare var URL: {
    prototype: URL;
    new(url: string | URL, base?: string | URL): URL;
};
```

I would even argue that it's a good practice to only manipulate URL objects, to avoid weird behaviours with paths, injections, etc.

The `undefined` comes from either `library` or `library?.url`.

src/services/item/plugins/published/constants.ts
```diff
@@ -3,8 +3,12 @@ import { Context } from '@graasp/sdk';
 import { CLIENT_HOSTS } from '../../../../utils/config';
 import { Item } from '../../entities/Item';
 
-export const buildPublishedItemLink = (item: Item) => {
-  const library = CLIENT_HOSTS.find(({ name }) => name === Context.Library);
-  const target = new URL(item.id, library?.url);
+export const buildPublishedItemLink = (item: Item): string | null => {
+  const library = CLIENT_HOSTS.find(({ name }) => name === Context.Library)?.url;
```
dialexo, line 7: This computation can be performed outside the function

src/services/item/plugins/published/constants.ts
```diff
@@ -3,8 +3,12 @@ import { Context } from '@graasp/sdk';
...
+  if (!library) {
+    return null;
+  }
```
dialexo, line 10: This is the only change that will affect the resulting URL, but it will be null

src/services/item/plugins/published/constants.ts
```diff
@@ -3,8 +3,12 @@ import { Context } from '@graasp/sdk';
...
+
+  const target = new URL(`/${item.id}`, library.origin);
```
dialexo, line 12: You don't need to use the `origin` property since you can use the URL type directly.

src/services/item/plugins/published/service.ts
```diff
@@ -34,7 +34,10 @@ export class ItemPublishedService {
       .map(({ member }) => member);
 
     const link = buildPublishedItemLink(item);
-
+    if (!link) {
+      console.error('cannot send published email with undefined link');
+      return;
+    }
```
dialexo, line 40: Not sure if this should run only when the function is called. It would make more sense to extract the computation above, and throw directly at the top-level, i.e. when the module is loaded at server start (fail fast).

pyphilia:
@dialexo I believe you I didn't try out. But then I don't get where the undefined comes from, the env is defined in the task def, `find` should be all fine, I might have missed something?

dialexo (approved):

PR #601, opened 2023-09-11
feat: add etherpad plugin docs stubs

dialexo:


src/services/item/plugins/etherpad/README.md
dialexo, line 1: You probably want to read the rendered README at https://github.com/graasp/graasp/blob/docs-etherpad/src/services/item/plugins/etherpad/README.md

src/services/item/plugins/etherpad/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+Note that the master API key must be placed in the root folder of the Etherpad server deployment, see the related [documentation](https://etherpad.org/doc/v1.8.18/#index_authentication):
+
+> Authentication works via a token that is sent with each request as a post parameter. There is a single token per Etherpad deployment. This token will be random string, generated by Etherpad at the first start. It will be saved in APIKEY.txt in the root folder of Etherpad. Only Etherpad and the requesting application knows this key. Token management will not be exposed through this API.
+
+The key format must be a hex string of 64 characters (i.e. regex `[a-f\d]{64}`).
+
+In the Graasp devcontainer, the key is bind-mounted from [this file](../../../../../.devcontainer/etherpad/devApiKey.txt) and the container is already available over the virtual network from the [docker-compose definition](../../../../../.devcontainer/docker-compose.yml).
+
+In a production container environment, you should either bind-mount a secret key, or build your own etherpad image using the `etherpad/etherpad` image as base and write your own secret file, see the [Dockerfile](Dockerfile) as an example.
```
pyphilia, line 25:
```
you should either bind-mount a secret key
```
I don't get how this could be equivalent to building an image. I guess you mean a remote service already exists?

src/services/item/plugins/etherpad/README.md
```diff
@@ -0,0 +1,243 @@
...
+| write      | write pad | write pad |
+| admin      | write pad | write pad |
+
+Note that the ability "read pad" should also enforce that a user **cannot** edit the pad, e.g. by manually constructing an URL.
+
+> Depending on user needs, it may be interesting to set that public pads (i.e. which item is public) are always writable in Builder, even for logged out users and users that only have a read permission
+
+### Player mode
+
+In the Player view, the pad is (currently) always shown in read-only mode. This may be subject to change in the future, depending on user needs, in which case the mapping should be carefuly redesigned not to leak editable pads to unauthorized users.
```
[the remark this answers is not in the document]
pyphilia: Also you don't mention why we didn't give the write permission in player. I remember there was a limitation.

src/services/item/plugins/etherpad/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+Since the granularity of permissions in Graasp is at the item level, we map permissions as follows:
+
+- Each Etherpad item in Graasp is mapped to an Etherpad group (the `groupMapper` is assigned a unique random identifier) which will contain a single pad.
+- Each Graasp member which accesses to an Etherpad item is assigned an Etherpad author (the `authorMapper` is assigned the member id)
+- Sessions are hence allocated between authors and groups, effectively mapping Graasp members to Graasp Etherpad items.
+- In Etherpad, permission is represented by a session (which is stored in a single cookie) **as well as the `padID`**. A user may be given the actual ` padID` with the session, which effectively grants him write permissions on the pad, or a special `readOnlyID` for a given `padID` which only grants read permissions. **It is up to the Graasp back-end and front-ends to never expose both to unpriviliged users**, i.e. a user which does not have at least write permission should _never_ see the actual `padID`.
```
pyphilia, line 117:
```suggestion
- In Etherpad, permission is represented by a session (which is stored in a single cookie) **as well as the `padID`**. A user may be given the actual `padID` with the session, which effectively grants him write permissions on the pad, or a special `readOnlyID` for a given `padID` which only grants read permissions. **It is up to the Graasp back-end and front-ends to never expose both to unpriviliged users**, i.e. a user which does not have at least write permission should _never_ see the actual `padID`.
```

pyphilia (approved):
Incredible documentation! Thank you!

src/services/item/plugins/etherpad/README.md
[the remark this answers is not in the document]
dialexo: I think it was mostly a design choice. The use case was for the SHS course, where the students would create the content in Builder, and then present it in Player

src/services/item/plugins/etherpad/README.md
dialexo: I've tried clarifying it a bit. So in the Etherpad server installation, the key should at exist the root path and named APIKEY.txt. In a containerized deployment, the Etherpad image should be instantiated with this file at /opt/etherpad/APIKEY.txt, in which case you have 2 options: either you bind-mount this key from the host filesystem to the path, or you build your copy of the etherpad container image with the key already baked in.

In our case, since we do not use bind mounts on the ECS instance of the etherpad container, we chose to build our own image with the key using the Dockerfile. Then the container with the key is shipped to ECR and ECS just picks up the image and spins up a container with it.

PR #610, opened 2023-09-14
feat: add pagination to get own items api

LinaYahya:


src/services/item/repository.ts
```diff
@@ -234,15 +235,29 @@ export const ItemRepository = AppDataSource.getRepository(Item).extend({
...
+  ): Promise<Paginated<Item>> {
+    const pageLimit = limit || ITEMS_LIST_LIMIT;
+    const skip = (page - 1) * pageLimit;
+    const [data, totalCount] = await this.createQueryBuilder('item')
       .leftJoinAndSelect('item.creator', 'creator')
       .innerJoin('item_membership', 'im', 'im.item_path @> item.path')
       .where('creator.id = :id', { id: memberId })
       .andWhere('im.permission = :permission', { permission: PermissionLevel.Admin })
       .andWhere('nlevel(item.path) = 1')
+      .andWhere("LOWER(item.name) LIKE '%' || :name || '%'", {
```
morganridel, line 253: I guess this still work well when `name` is empty? Is this a filter?

src/services/item/repository.ts
```diff
@@ -234,15 +235,29 @@ export const ItemRepository = AppDataSource.getRepository(Item).extend({
...
   },
 
-  async getOwn(memberId: string): Promise<Item[]> {
-    return this.createQueryBuilder('item')
+  async getOwn(
+    memberId: string,
+    page: number,
+    name: string,
+    all: boolean,
+    limit: number,
```
morganridel, line 243: I think it might be cool to wrap these properties (`page`, `all`, `limit` in something like a `PageRequest` or something, that could be useful for future pagination too, but this is not absolutely needed right now.

Then for filters and sorting, I don't know if I would try to put it inside this `PageRequest` with arbitrary strings somehow.

Right now, I think you could still make `name` argument optional if it's not required when calling `getOwn`

morganridel (approved):
LGTM, nice to finally have a beginning of pagination, I added small comments. 
We might want to hold off merging this until we have deployed the search feature, to avoid too many changes if we need to rollback the search.

@dialexo could you take a quick look too?

src/services/item/fluent-schema.ts
```diff
@@ -159,8 +160,17 @@ export const getParents = {
 };
 
 export const getOwn = {
+  querystring: S.object()
+    .additionalProperties(false)
+    .prop('page', S.number().default(1))
+    .prop('name', S.string())
+    .prop('limit', S.number().default(ITEMS_LIST_LIMIT))
+    .prop('all', S.boolean().default(false)),
```
pyphilia, line 168: I think it's maybe too much to add `all` just to disable pagination. What about setting a big limit instead? (sorry I think I told the opposite last time).

src/services/item/repository.ts
pyphilia, line 243: I prefer the following
```suggestion 
type PaginationArgs = { 
      page: number,
      all: boolean,
      limit: number,
    }

  async getOwn(
    memberId: string,
    searchArgs: {
      name: string,
    },
    args: PaginationArgs
```

src/services/item/service.ts
```diff
@@ -129,11 +130,18 @@ export class ItemService {
...
   }
 
-  async getOwn(actor: Actor, { itemRepository }: Repositories) {
+  async getOwn(
+    actor: Actor,
+    { itemRepository }: Repositories,
+    page: number = 1,
+    name: string = '',
+    all: boolean = false,
+    limit: number = ITEMS_LIST_LIMIT,
```
pyphilia, line 139: See my comment above.

src/services/item/test/index.test.ts
```diff
@@ -556,10 +556,10 @@ describe('Item routes tests', () => {
         expect(response.statusCode).toBe(StatusCodes.OK);
 
         const data = response.json();
-        expect(data).toHaveLength(items.length);
+        expect(data.data).toHaveLength(items.length);
         items.forEach(({ id }) => {
           expectItem(
-            data.find(({ id: thisId }) => thisId === id),
+            data?.data?.find(({ id: thisId }) => thisId === id),
             items.find(({ id: thisId }) => thisId === id),
```
pyphilia, line 563: It would be interesting to add a test to see if you indeed get the second page, and/or the limit works.

pyphilia (changes requested):
Thanks for the PR! I've added my comments, let me know if you have questions 🦘

src/services/item/fluent-schema.ts
LinaYahya: @pyphilia 
Thanks for your comments, really appreciated!
Do you mean to send a limit as a total count or a big number instead of having all as a flag?

src/services/item/repository.ts
LinaYahya: @morganridel 
yes it works well if the name is an empty string, and maybe the naming is misconceptual but yes name should be as an item name search query

src/services/item/fluent-schema.ts
```diff
@@ -159,8 +160,16 @@ export const getParents = {
 };
 
 export const getOwn = {
+  querystring: S.object()
+    .additionalProperties(false)
+    .prop('page', S.number().default(1))
+    .prop('name', S.string())
+    .prop('limit', S.number().default(ITEMS_LIST_LIMIT)),
```
dialexo, line 167: Is it desirable to let the consumer set an arbitrary limit? Should it be capped? @morganridel @pyphilia

dialexo (approved):

src/services/item/fluent-schema.ts
pyphilia: We still need the full list for some cases, but I agree it should be capped sometime. @LinaYahya could you add a comment about that?

pyphilia (approved):

src/services/item/fluent-schema.ts
LinaYahya: Not sure how we would cap the value <sometime>, the point of we can't set a limit for `limit`, in case of getting the whole list we will send big values! for this we can't do have a maximum bound for it.
@pyphilia  @dialexo

src/services/item/repository.ts
```diff
@@ -242,15 +244,29 @@ export const ItemRepository = AppDataSource.getRepository(Item).extend({
...
       .where('creator.id = :id', { id: memberId })
       .andWhere('im.permission = :permission', { permission: PermissionLevel.Admin })
       .andWhere('nlevel(item.path) = 1')
+      .andWhere("LOWER(item.name) LIKE '%' || :name || '%'", {
+        name: searchArgs.name.toLowerCase().trim(),
+      })
       .orderBy('item.updatedAt', 'DESC')
-      .getMany();
+      .skip(skip)
+      .take(pageLimit)
```
pyphilia, line 267: After a documentation review, it appears it's better to use `limit` and `offset`
https://orkhan.gitbook.io/typeorm/docs/select-query-builder#adding-offset-expression

pyphilia (changes requested):

PR #614, opened 2023-09-22
fix: order parents

### LINKED ISSUE #613: GET /parents returns without order
This causes the breadcrumbs to be in disorder.

pyphilia:
This PR forces `getAncestors` to be ordered. This fixes the problem with the breadcrumbs in the frontend.
I added a bunch of tests there were none for this endpoint.

fix #613


spaenleh (approved):
👏 Looks good to me ! I let others chime in to confirm 🔔

dialexo (approved):

PR #616, opened 2023-09-22
feat: add post bug api service

### LINKED ISSUE #615: POST /bug to send user feedback to graasp email
Create a route that will send a bug report email to graasp admin, 
Basically email would contain those information
- reporter name
- reporter email
- report details

LinaYahya:
closes #615


src/services/bug/controller.ts
```diff
@@ -0,0 +1,35 @@
...
+    async (request, reply) => {
+      const { member, body: data } = request;
+      const { log, mailer } = fastify;
+
+      const subject = `Graasp Builder | User Feedback`;
+      const html = generateEmail(data);
+
+      if (BUGS_RECIVER_EMAIL) {
+        return mailer
+          .sendEmail(subject, BUGS_RECIVER_EMAIL, '', html)
```
pyphilia, line 25:
```suggestion
          .sendEmail(subject, REPORTING_RECEIVER_EMAIL, '', html)
```

src/services/bug/controller.ts
```diff
@@ -0,0 +1,35 @@
...
+    {
+      schema: bugSchema,
+      preHandler: fastify.verifyAuthentication,
+    },
+    async (request, reply) => {
+      const { member, body: data } = request;
+      const { log, mailer } = fastify;
+
+      const subject = `Graasp Builder | User Feedback`;
+      const html = generateEmail(data);
```
pyphilia, line 21: There're actually some util functions to generate an email, it would be better to use them I think https://github.com/graasp/graasp/blob/main/src/plugins/mailer/index.ts

src/services/bug/controller.ts
```diff
@@ -0,0 +1,35 @@
...
+
+      if (BUGS_RECIVER_EMAIL) {
+        return mailer
+          .sendEmail(subject, BUGS_RECIVER_EMAIL, '', html)
+          .then(() => ({ message: 'feedback received successfully' }))
```
pyphilia, line 26: I think it's okay to just `console.log` that.
The returned value of the endpoint could just be an empty `OK`/200 response.

src/services/bug/index.ts
```diff
@@ -0,0 +1,24 @@
...
+      if (fastify.corsPluginOptions) {
+        fastify.register(fastifyCors, fastify.corsPluginOptions);
+      }
+
+      // core routes - require authentication
+      fastify.register(async function (fastify) {
+        fastify.register(bugController);
+      });
+    },
+    { prefix: BUGS_REPORTING_PREFIX },
```
pyphilia, line 20: I think it's clearer if the prefix is clearly stated? Maybe also let's use `report` instead.

src/services/bug/utils.ts
```diff
@@ -0,0 +1,13 @@
...
+<p>I hope this email finds you well. I wanted to bring to your attention a bug that I've encountered while using Graasp. I believe it's important to report this issue to help improve the overall user experience.</p>
+<strong>Bug Details:</strong>
+<ul>
+    <li><strong>Reporter Name:</strong> ${name}</li>
+    <li><strong>Reporter Email:</strong> ${email}</li>
+    <li><strong>Details:</strong> ${details}</li>
+</ul>
+<p>I believe that addressing this bug will contribute to an enhanced user experience for all users of Graasp. Please connect reporter for any further information or assistance in reproducing the issue.</p>
+
+  `;
```
pyphilia, line 13: As suggested above, use util functions instead of this template.

src/services/bug/controller.ts
```diff
@@ -0,0 +1,35 @@
...
+  fastify.post<{ Body: { name: string; email: string; details: string } }>(
+    '/',
+    {
+      schema: bugSchema,
+      preHandler: fastify.verifyAuthentication,
+    },
+    async (request, reply) => {
+      const { member, body: data } = request;
+      const { log, mailer } = fastify;
+
```
pyphilia, line 19: I think you should check for a captcha here, to avoid spamming. You can find example in `auth/magicLink`.

src/services/bug/utils.ts
```diff
@@ -0,0 +1,13 @@
+export const generateEmail = ({ name, email, details }) => `
+<p>Dear Graasp Team,</p>
+
+<p>I hope this email finds you well. I wanted to bring to your attention a bug that I've encountered while using Graasp. I believe it's important to report this issue to help improve the overall user experience.</p>
+<strong>Bug Details:</strong>
+<ul>
+    <li><strong>Reporter Name:</strong> ${name}</li>
+    <li><strong>Reporter Email:</strong> ${email}</li>
+    <li><strong>Details:</strong> ${details}</li>
+</ul>
```
pyphilia, line 10: The wrapper is nice. Maybe let's add the front-end url as well.

pyphilia (changes requested):
Thanks for the PR. I think `Bug` is not a good naming. Maybe let's go with `Reporting` such as `ReportingService`, `services/reporting/controller.ts`, `reportingSchema` etc. So we can also reuse this feature for simple feedback (that may contain just some thanks or suggestions).

**Also since you added an environment variable we need to change the deploy workflow**

src/services/bug/index.ts
```diff
@@ -0,0 +1,24 @@
...
+    async function (fastify) {
+      // add CORS support
+      if (fastify.corsPluginOptions) {
+        fastify.register(fastifyCors, fastify.corsPluginOptions);
+      }
+
+      // core routes - require authentication
+      fastify.register(async function (fastify) {
+        fastify.register(bugController);
+      });
```
spaenleh, line 18: maybe a dumb question, but why do we double-wrap with register here ? @pyphilia ?

spaenleh:
We should have #626 ready before/at the same time that we merge this. @LinaYahya @pyphilia

src/app.ts
```diff
@@ -76,7 +77,8 @@ export default async function (instance: FastifyInstance): Promise<void> {
     await instance
       .register(fp(MemberServiceApi))
       .register(fp(ItemServiceApi))
-      .register(fp(ItemMembershipServiceApi));
+      .register(fp(ItemMembershipServiceApi))
+      .register(fp(BugServicApi));
```
spaenleh, line 81: Rename to `UserFeedbackServiceApi` ?

src/plugins/mailer/index.ts
```diff
@@ -20,9 +20,14 @@ export interface MailerOptions {
   fromEmail: string;
 }
 
+export interface Item {
```
spaenleh, line 23: We already have an entity called `Item`. I think we could use something more descriptive:
```suggestion
export interface ListItem {
```

src/services/userFeedback/controller.ts
```diff
@@ -0,0 +1,46 @@
...
+  fastify.post<{ Body: { name: string; email: string; details: string } }>(
+    '/',
+    {
+      schema: userFeedbackSchema,
+      preHandler: fastify.verifyAuthentication,
+    },
+    async (request, reply) => {
+      const { member, body: data, hostname } = request;
+      const { log, mailer } = fastify;
+      console.log(hostname);
```
spaenleh, line 19: Remove ?

src/services/userFeedback/controller.ts
```diff
@@ -0,0 +1,46 @@
...
+      const subject = `User Feedback`;
```
spaenleh, line 20: Question: Does Sentry provide the issue number in the payload ? In case they do, it would be interesting to have it in order to ease the search of the issue later in the emails.

src/plugins/mailer/index.ts
```diff
@@ -31,6 +36,7 @@ export interface MailerDecoration {
     from?: string,
   ) => Promise<void>;
   translate: (lang: string) => i18n['t'];
+  buildUnorderList: (list: Item[]) => string;
```
spaenleh, line 39:
```suggestion
  buildUnorderedList: (list: Item[]) => string;
```

src/plugins/mailer/index.ts
```diff
@@ -84,6 +90,17 @@ const plugin: FastifyPluginAsync<MailerOptions> = async (fastify, options) => {
   };
 
   const buildText = (text: string) => `<p>${text}</p>`;
+  const buildBoldText = (text: string) => `<strong>${text}</strong>`;
+  const buildUnorderList = (list: Item[]) => `<ul>
```
spaenleh, line 94:
```suggestion
  const buildUnorderedList = (list: Item[]) => `<ul>
```

src/plugins/mailer/index.ts
```diff
@@ -95,6 +112,8 @@ const plugin: FastifyPluginAsync<MailerOptions> = async (fastify, options) => {
     buildText,
     sendEmail,
     translate,
+    buildBoldText,
+    buildUnorderList,
```
spaenleh, line 116:
```suggestion
    buildUnorderedList,
```

src/utils/config.ts
```diff
@@ -329,11 +329,14 @@ export const RECAPTCHA_SCORE_THRESHOLD = 0.5;
 export const GET_MOST_LIKED_ITEMS_MAXIMUM = 50;
 export const GET_MOST_RECENT_ITEMS_MAXIMUM = 50;
 
+// user feedback reporting
+export const USER_FEEDBACK_PREFIX = '/user-feedback';
+export const USER_FEEDBACK_RECEIVER_EMAIL = process.env.USER_FEEDBACK_RECEIVER_EMAIL;
```
spaenleh, line 334: For this environnement variable, I would use the same pattern as for the `RECAPTCHA_SECRET_ACCESS_KEY`. The app should exit if it not set. I think it makes sens to require it, as otherwise the feature is not able to work as expected.

src/utils/config.ts
```diff
@@ -329,11 +329,14 @@ export const RECAPTCHA_SCORE_THRESHOLD = 0.5;
 export const GET_MOST_LIKED_ITEMS_MAXIMUM = 50;
 export const GET_MOST_RECENT_ITEMS_MAXIMUM = 50;
 
+// user feedback reporting
+export const USER_FEEDBACK_PREFIX = '/user-feedback';
```
spaenleh, line 333: I think it would make more ssens to define this in `src/services/userFeedback/index.ts` directly. Like it is done for the `/members` route.

src/services/userFeedback/controller.ts
```diff
@@ -0,0 +1,46 @@
...
+      ${mailer.buildBoldText(USER_FEEDBACK_MAIL.TITLE)}
+      ${mailer.buildUnorderList([
+        { title: USER_FEEDBACK_MAIL.REPORTER_NAME, value: name },
+        { title: USER_FEEDBACK_MAIL.REPORTER_EMAIL, value: email },
+        { title: USER_FEEDBACK_MAIL.HOSTNAME, value: hostname },
+        { title: USER_FEEDBACK_MAIL.DETAILS, value: details },
+      ])}
+      ${mailer.buildText(USER_FEEDBACK_MAIL.CONCLUSION)}`;
+
+      if (USER_FEEDBACK_RECEIVER_EMAIL) {
```
spaenleh, line 34: See my other comment. But I think this env should always be defined.

spaenleh (changes requested):
🦁 Thank you for the changes, I have added a few comments and suggestions for issues that still need to be addressed.

src/services/bug/controller.ts
spaenleh: @LinaYahya this will require a bit more changes (in the frontend as well). Please create a new issue to handle this.

PR #620, opened 2023-09-25
docs: copy app docs

pyphilia:
I've copy-pasted the app docs that existed in its own repo. 
The only change is `app` of "app id" to `key`when requesting the api key.


spaenleh (approved):
LGTM !

PR #625, opened 2023-09-26
feat: improve meilisearch integration resiliency

morganridel:


src/services/item/plugins/published/plugins/search/meilisearch.ts
```diff
@@ -89,15 +91,15 @@ export class MeiliSearchWrapper {
...
+    try {
+      const finishedTask = await (await this.getIndex()).waitForTask(task.taskUid);
       this.logger.info(
-        `Meilisearch task completed: item name => ${itemName}, task type => ${task.type}, result: ${finishedTask.error}`,
+        `Meilisearch ${task.type} task completed (${task.status}): item name => ${itemName}. ${
+          finishedTask.error ? `Error: ${finishedTask.error}` : ''
+        }`,
       );
+    } catch (e) {
+      // catch timeout while waiting
```
pyphilia, line 102: Should we print something here?

src/services/item/plugins/published/plugins/search/meilisearch.ts
```diff
@@ -328,15 +336,17 @@ export class MeiliSearchWrapper {
...
+      await this.storeMissingPdfContent(buildRepositories());
+    }
 
-    this.logger.info('Starting index rebuild...');
+    this.logger.info('REBUILD INDEX: Starting index rebuild...');
     const tmpIndex = await this.getIndex(ROTATING_INDEX);
 
     // Delete existing document if any
-    this.logger.info('Cleaning temporary index...');
+    this.logger.info('REBUILD INDEX: Cleaning temporary index...');
```
pyphilia, line 349: Here you delete the previous temporary index right?

pyphilia (approved):
LGTM 🏯

src/services/item/plugins/published/plugins/search/meilisearch.ts
morganridel: I'm not sure, I can add something, I just don't want to make it confusing because a timeout does not mean the task will fail

src/services/item/plugins/published/plugins/search/meilisearch.ts
morganridel: Yes, But I prefix all the log with the same string so we can filter for it as a single "process"

PR #632, opened 2023-10-03
fix(app): use correct authentication method to deliver context

spaenleh:
I stumbled on this by accident.

Currently the browser sends the user token alongside the bearer token which allows the backend to get the member using the `attemptVerifyAuthentication`. 

When testing locally with a fake subdomain on localhost the user cookie is isolated. In this case the `GET /context` for the app living on `http://apps.localhost:3005` will fail.

I propose the following changes to make the authentication method work as expected (use only the bearer token provided by the app). This needs to be checked in development environnement to ensure that the apps continue to work.

Suggestions and comments appreciated.


src/services/item/plugins/app/test/index.test.ts
```diff
@@ -144,30 +162,26 @@ describe('Apps Plugin Tests', () => {
   });
 
   describe('GET /:itemId/context', () => {
-    let apps;
-    let chosenApp;
+    // let apps;
+    // let chosenApp;
```
spaenleh, line 166: remove ?

spaenleh (commented):
LGTM ! Thank you for fixing the tests !

PR #636, opened 2023-10-09
build: get views for collection

LinaYahya:
get collection views within get collection information


src/services/item/plugins/published/schemas.ts
```diff
@@ -21,6 +21,24 @@ const publishEntry = {
...
+    item: {
+      $ref: 'http://graasp.org/items/#/definitions/item',
+    },
+    creator: {
+      $ref: 'http://graasp.org/members/#/definitions/member',
+    },
+    createdAt: { type: 'string' },
+    itemViews: {
+      type: 'array',
+    },
```
spaenleh, line 37: Do we need to send all the views (as an array that has potentially hundreds of elements) when requesting the collection info ? 
Are we interested in the total sum, in which case it would make sens to only return the total number of views ?

spaenleh (changes requested):
Thank you for this PR.

What should the itemPublish info call return ? I am leaning towards a single number (total number of views) instead of returning the full views data.
the actions contain personal data on who created them, at what time and so on, we do not want to expose that data without agregation.

If the idea was to have that data available for a graph of the temporal evolution I think it would  be better to use a dedicated endpoint that ensures anonymisation by aggregating the data. What do you think ?

src/services/item/plugins/published/schemas.ts
LinaYahya: @spaenleh 
What if we need the created date to have a sparkline, Shall we only return the created date, or what do u suggest?
views
```
[
{ createdDate: ''},
{ createdDate: ''},
]
```

spaenleh: I think that the spark line should be considered as a separate usecase. My justification is as follows: 
- you might want to query for a particular period of time
- you might to get an aggregated result (binning the results into 1h intervals for example)
- etc

So I think this flexibility would be better exposed on a separate endpoint.
I would like to keep the `/informations` endpoint of a collection more "summarized". 
TLDR: this endpoint should return the total number of views. Create a new endpoint to get the sparkline data that allows to pass in more params.

src/services/item/plugins/published/service.ts
```diff
@@ -58,12 +58,12 @@ export class ItemPublishedService {
 
     // item should be public first
     await itemTagRepository.getType(item, ItemTagType.Public, { shouldThrow: true });
-    const itemViews = await actionRepository.getForItem(item.path, {
+    const totalViews = await actionRepository.getForItem(item.path, {
```
spaenleh, line 61: Maybe we should leverage the `getAggregationForItem` function, in order to perform the sum in sql and not in JS (performance concerns).

spaenleh (changes requested):

spaenleh (commented):
We should also note that both methods are limited to either the specified sample size or the default (if sample size is not specified) which is 5000. Otherwise changes look good.

spaenleh (approved):
Looks good !

PR #640, opened 2023-10-09
fix(types): fileStream argument has incompatible typing

spaenleh:
This PR fixes an issue with the typings.
This fix is coherent with the input typing https://github.com/graasp/graasp/blob/20d59793c97ccf3beb83fb8be6cb7af94d0bdd98/src/services/file/service.ts#L58

### DIFF
--- a/src/services/file/repositories/s3.ts
+++ b/src/services/file/repositories/s3.ts
@@ -6,6 +6,7 @@ import fs from 'fs';
 import { StatusCodes } from 'http-status-codes';
 import fetch from 'node-fetch';
 import path from 'path';
+import Readable from 'stream';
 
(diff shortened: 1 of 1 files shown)


morganridel (approved):

src/services/file/repositories/s3.ts
```diff
@@ -6,6 +6,7 @@ import fs from 'fs';
 import { StatusCodes } from 'http-status-codes';
 import fetch from 'node-fetch';
 import path from 'path';
+import Readable from 'stream';
```
spaenleh, line 9: This was wrong. Should have used a named import instead:
```ts
import { Readable } from 'stream';
```

PR #649, opened 2023-10-16
fix: directly extend Error for FileError

spaenleh:


src/services/file/repositories/s3.ts
```diff
@@ -254,12 +254,13 @@ export class S3FileRepository implements FileRepository {
 
       return url;
     } catch (e) {
-      // todo: usign console.log polutes the logs
-      log.error(e);
-      // todo: handle the "Not Found" error
+      if (e.name === 'NotFound') {
```
pyphilia, line 257: Is there a possibility to use `instanceof`?

pyphilia (approved):
LGTM!

src/services/file/repositories/s3.ts
spaenleh: I tried it first, but it does not work: it says that right hand side of `instanceof` must be a class. Not sure what they do with their code but this worked better.

PR #653, opened 2023-10-18
fix: add missing relations in appSettings

### LINKED ISSUE #651: `GET app-setting` does not return property `item` expected in AppSetting type
The `AppSetting` type in the frontends expects an `item: DiscriminatedItem` property, but the backend does not return it in the appSetting object.

We are missing a `relations` option in the TypeORM query.

spaenleh:
In this PR I added the missing relation on `item` for the `GET app-setting` endpoint and improved the return type of repository functions

close #651

### DIFF
--- a/src/services/item/plugins/app/appSetting/repository.ts
+++ b/src/services/item/plugins/app/appSetting/repository.ts
@@ -1,6 +1,7 @@
 import { ItemType } from '@graasp/sdk';
 
 import { AppDataSource } from '../../../../../plugins/datasource';
+import { ItemNotFound } from '../../../../../utils/errors';
 import { AppSetting } from './appSettings';
 import { AppSettingNotFound, PreventUpdateAppSettingFile } from './errors';
 import { InputAppSetting } from './interfaces/app-setting';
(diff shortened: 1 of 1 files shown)


src/services/item/plugins/app/appSetting/repository.ts
```diff
@@ -46,16 +50,21 @@ export const AppSettingRepository = AppDataSource.getRepository(AppSetting).exte
...
       throw new AppSettingNotFound(id);
     }
-    return this.findOneBy({ id });
+    return this.findOne({ where: { id }, relations: { creator: true, item: true } });
   },
 
-  getForItem(itemId: string) {
-    return this.findBy({ item: { id: itemId } });
+  getForItem(itemId: string): Promise<AppSetting[]> {
+    return this.find({
```
pyphilia, line 63: To avoid any error, we check that `itemId` really exists (even if it's not set as optional here). Otherwise `this.find` will return everything which would just explode. If you could add this check!

pyphilia (approved):
LGTM 🍇

PR #655, opened 2023-10-18
fix: add local test db

### LINKED ISSUE #654: Local tests should run on another database

spaenleh:
- fix: remove unnecessary DISABLE_LOGS env variable We can't disable logs at runtime so this option is useless
- fix(test): setup a local db for tests
- chore: remove unused imports and fix schema warnings

close #654


.devcontainer/devcontainer.json
```diff
@@ -41,7 +51,7 @@
     }
   }
   // Use 'forwardPorts' to make a list of ports inside the container available locally.
-  // "forwardPorts": [3000, 5432],
+  // "forwardPorts": [3000, 5434],
```
pyphilia, line 54: Why change this?

.devcontainer/devcontainer.json
```diff
@@ -13,18 +13,28 @@
...
+            "password": "docker-test"
+          },
           {
             "name": "Etherpad database",
             "driver": "PostgreSQL",
             "previewLimit": 50,
-            "server": "db",
-            "port": 5433,
+            "server": "graasp-postgres-etherpad",
+            "port": 5432,
```
pyphilia, line 37: Shouldn't this be `5433`? I'm not quite sure how the settings work, but you changed the value here but not in `docker-compose`. Or was it a bug?

.devcontainer/devcontainer.json
spaenleh: I was doing tests, will revert it to minimise noise.

.devcontainer/devcontainer.json
spaenleh: @pyphilia This took me an entire afternoon to understand:
- inside the dev-container the other containers are addressed by their `container-name` and `internal ports`, in this case the etherpad database can be reached using `host=graasp-postgres-etherpad` and `port=5432` the exposed port `5433` is only used when accessing from the host (your terminal on your machine)
- I tested and while the previous configuration was unable to open a connection with the sql-tools extensions in vs-code, with this one it works.

PR #662, opened 2023-10-25
fix: set member id on post app data

### LINKED ISSUE #661: [app-data] POST app data does not set member id

pyphilia:
Text input cannot set the member because the backend doesn't. 
This should fix the error, I've added a corresponding test.

close #661


spaenleh (approved):
Nice to see that it was "just missing" 🌮

PR #667, opened 2023-10-26
feat: add action triggers to download item, like, unlike, search item

### LINKED ISSUE #663: Trigger actions for library collection preview

LinaYahya:
closes #663


src/services/item/plugins/importExport/index.ts
```diff
@@ -101,7 +102,7 @@ const plugin: FastifyPluginAsync = async (fastify) => {
   );
 
   // download item as zip
-  fastify.route<{ Params: { itemId: string } }>({
+  fastify.route<{ Params: { itemId: string }; Querystring: { type?: string } }>({
```
spaenleh, line 105: If possible I would use a more descriptive argument name:
```suggestion
  fastify.route<{ Params: { itemId: string }; Querystring: { actionType?: string } }>({
```

src/services/item/plugins/published/plugins/search/index.ts
```diff
@@ -17,12 +17,22 @@ export type SearchFields = {
...
-    async ({ member, body }) => {
-      return searchService.search(member, buildRepositories(), body as MultiSearchParams);
+    async (request) => {
+      const { member, body } = request;
+      const repositories = buildRepositories();
+
+      // TODO: chnage type from graasp/sdk
+      const action = {
+        type: 'item-search',
+        extra: { ...(body as any) },
```
spaenleh, line 32: Do you have an issue with the types here ? I am ok to keep it, but if possible I would prefer `unknown`.

spaenleh (changes requested):
Thank you for these changes, I would just change the queryParam name to something clearer.

src/services/item/plugins/itemLike/index.ts
```diff
@@ -34,8 +36,22 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+        // action like item
+        const item = await itemService.get(member, buildRepositories(manager), itemId);
+        // TODO: chnage type from graasp/sdk
+        const action = {
+          item,
+          type: 'item-like',
+          extra: {},
+        };
+
+        await actionService.postMany(member, buildRepositories(manager), request, [action]);
```
pyphilia, line 54: Let's create the action only after the post.

src/services/item/plugins/itemLike/index.ts
```diff
@@ -45,8 +61,23 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+        const item = await itemService.get(member, buildRepositories(manager), itemId);
+
+        // TODO: chnage type from graasp/sdk
+        const action = {
+          item,
+          type: 'item-unlike',
+          extra: {},
+        };
+
+        await actionService.postMany(member, buildRepositories(manager), request, [action]);
```
pyphilia, line 80: Same as above.

src/services/item/plugins/importExport/index.ts
```diff
@@ -111,12 +112,21 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
       } = request;
       const repositories = buildRepositories();
-
-      // check item and permission
       const item = await iS.get(member, repositories, itemId);
 
+      if (actionType) {
+        const action = {
+          item,
+          type: actionType,
```
pyphilia, line 123: Maybe we want to always save an action. Is it possible to set a default value?

src/services/item/plugins/importExport/index.ts
```diff
@@ -101,7 +102,7 @@ const plugin: FastifyPluginAsync = async (fastify) => {
   );
 
   // download item as zip
-  fastify.route<{ Params: { itemId: string } }>({
+  fastify.route<{ Params: { itemId: string }; Querystring: { actionType?: string } }>({
```
pyphilia, line 105: Might be interesting to add a comment about what values can this be? Will it refer to where in the ui we clicked? Can this info be added in the extra and have a global action type?

src/services/item/plugins/published/plugins/search/index.ts
```diff
@@ -17,12 +19,22 @@ export type SearchFields = {
...
-    async ({ member, body }) => {
-      return searchService.search(member, buildRepositories(), body as MultiSearchParams);
+    async (request) => {
+      const { member, body } = request;
+      const repositories = buildRepositories();
+
+      // TODO: chnage type from graasp/sdk
+      const action = {
+        type: 'item-search',
+        extra: body as Partial<Action> & Pick<Action, 'type'>,
```
pyphilia, line 34: You can precise the type like this if I'm not mistaken:
`fastify.post<{ Body: Partial<Action> & Pick<Action, 'type'> }>(`

pyphilia: Also update the schema of the search endpoint accordingly.

pyphilia (changes requested):
Thanks for your PR, I'll approve once you add the action type in sdk!

spaenleh (approved):
🍪 Nice !

src/services/item/plugins/importExport/test/index.test.ts
```diff
@@ -149,11 +149,6 @@ describe('Member routes tests', () => {
       });
 
       expect(response.statusCode).toBe(StatusCodes.ACCEPTED);
-
-      await waitForExpect(async () => {
-        const items = await ItemRepository.find();
-        expect(items).toHaveLength(1);
-      }, 1000);
```
pyphilia, line 156: I don't think this should be removed.
What result do you have?

src/services/item/plugins/itemLike/index.ts
```diff
@@ -34,9 +38,25 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+      } = request;
       return db.transaction(async (manager) => {
-        return itemLikeService.post(member, buildRepositories(manager), itemId);
+        // action like item
+        const item = await itemService.get(member, buildRepositories(manager), itemId);
+        const action = {
+          item,
+          type: Triggers.ItemLike,
+          extra: {
+            itemId: item?.id,
```
pyphilia, line 53:
```suggestion
            itemId: item.id,
```

pyphilia: Maybe let's create the action after performing the like, so that the scope of `action` is smaller.

src/services/item/plugins/itemLike/index.ts
```diff
@@ -45,9 +65,30 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+        // action unlike item
+        const item = await itemService.get(member, buildRepositories(manager), itemId);
+
+        const action = {
+          item,
+          type: Triggers.ItemUnlike,
+          extra: {
+            itemId: item?.id,
+          },
+        };
```
pyphilia, line 83: Same comments as above.

src/services/item/plugins/published/plugins/search/index.ts
```diff
@@ -17,12 +19,21 @@ export type SearchFields = {
...
+    async (request) => {
+      const { member, body } = request;
+      const repositories = buildRepositories();
+
+      const action = {
+        type: Triggers.ItemSearch,
+        extra: body,
+      };
+      await actionService.postMany(member, repositories, request, [action]);
+      return searchService.search(member, repositories, body);
```
pyphilia, line 36: Maybe let's perform the search first, then the post action. Otherwise you might end up having an action even though the search failed.

pyphilia (changes requested):
Thanks for the PR! I'm a bit concerned about the test that you changed tho?

Issue #664, opened 2023-10-25
Add short links feature


ReidyT:
Explanation from @spaenleh :
A short link allows to access content by using a shorter identifier.
A short link is essentially a hard-coded redirection.
A short link is associated with a view: builder / player / library
A short link can be a random string automatically generated or an alias defined by the user.
A short link is unique.
A short link can be created by an administrator and is managed by him.
A short link can be deleted revoked, in which case it won't redirect to the content anymore.
A revoked short link can be used again to link to a different content.
Content targeted by a short link is subject to standard access permissions.
A short link is a set of minimum 6 chars and up to 255
A short link alphabet is: a-z, A-Z, 0-9,-_.*+  Explicitly disallowed chars are ?/#;,:%&()=## Implementation
Short links are stored in a new table.
They have :
- id: unique UUIDv4 for the link -- should this be the short link identifier ? like hello-world ?
- name: the short link identifier
- item_path: FK to an item
- createdAt: creation timestamp

PR #672, opened 2023-11-02
feat: add short links routes (#664)

ReidyT:
This PR implements and tests the issue (graasp/graasp#664).

The implemented routes are:
- POST /short-links
- GET /short-links/:shortLinkId
- GET /short-links/items/:itemId
- PATCH /short-links/:shortLinkId
- DELETE /short-links/:shortLinkId

The tests check that only admin or creator of items can manage short-links. The exception is for the GET /short-links/:shortLinkId to allow everyone to be redirected to the good item using a short link.

> [!IMPORTANT]  
> Do not forget to update the SDK to the latest NPM when published !


src/services/shortLink/service.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (!member) throw new UnauthorizedMember();
+    const item = await itemRepository.get(shortLink.item_id);
+
+    // check that the member can admin the item to be allowed to create short link
+    await validatePermission(repositories, PermissionLevel.Admin, member, item);
+
+    return shortLinkRepository.createOne(shortLink);
+  }
+
+  async get(alias: string, repositories: Repositories) {
```
spaenleh, line 23:
```suggestion
  async getOne(alias: string, repositories: Repositories) {
```

src/services/shortLink/service.ts
```diff
@@ -0,0 +1,87 @@
...
+    const { shortLinkRepository } = repositories;
+    const shortLink = await shortLinkRepository.get(alias);
+
+    return shortLink;
+  }
+
+  async getItem(member: Actor, itemId: string, repositories: Repositories) {
```
spaenleh, line 30:
```suggestion
  async getAllForItem(member: Actor, itemId: string, repositories: Repositories) {
```

src/services/shortLink/service.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (!member) throw new UnauthorizedMember();
+    const item = await itemRepository.get(itemId);
+
+    // check that the member can admin the item to be allowed to create short link
+    await validatePermission(repositories, PermissionLevel.Admin, member, item);
+
+    return await shortLinkRepository.getItem(itemId);
+  }
+
+  async getRealPath(alias: string, repositories: Repositories) {
```
spaenleh, line 42:
```suggestion
  async getRedirection(alias: string, repositories: Repositories) {
```

src/services/shortLink/service.ts
```diff
@@ -0,0 +1,87 @@
...
+      item_id: shortLink.item_id,
+    };
+  }
+
+  async delete(member: Actor, alias: string, repositories: Repositories) {
+    const { shortLinkRepository, itemRepository } = repositories;
+
+    if (!member) throw new UnauthorizedMember();
+    const shortLink = await shortLinkRepository.get(alias);
+    const item = await itemRepository.get(shortLink.item_id);
```
spaenleh, line 59: You can get the `item` by using the `relations: { item: true }` in your typeorm `this.find()` query.

src/services/shortLink/index.ts
```diff
@@ -0,0 +1,78 @@
...
+      {
+        preHandler: fastify.verifyAuthentication,
+      },
+      async ({ member, params: { alias } }, reply) => {
+        const oldLink = await shortLinkService.delete(member, alias, buildRepositories());
+        reply.code(200).send(oldLink);
+      },
+    );
+
+    fastify.patch<{ Params: { alias: string }; Body: ShortLink }>(
```
spaenleh, line 59: Your body may likely not contain the entity type.

src/services/shortLink/service.ts
```diff
@@ -0,0 +1,87 @@
...
+  async update(
+    member: Actor,
+    alias: string,
+    shortLink: Partial<ShortLink>,
+    repositories: Repositories,
+  ) {
+    const { shortLinkRepository, itemRepository } = repositories;
+
+    if (!member) throw new UnauthorizedMember();
+    const itemId = shortLink.item_id ?? (await shortLinkRepository.get(alias)).item_id;
```
spaenleh, line 77: In the schema you exclude the `item_id`, so I am not sure where you would get it from ...

src/services/shortLink/schemas.ts
```diff
@@ -0,0 +1,33 @@
+import { GraaspPlatform } from '@graasp/sdk';
+
+const shortLinkDefinition = {
```
spaenleh, line 3:
```suggestion
const shortLinkPayload = {
```

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+    });
+
+    describe('Not Found', () => {
+      it('Conflict if post short links with already exist alias', async () => {
+        const { item } = await mockPostData({ author: actor });
+        const { item: item2 } = await mockPostData({ author: actor });
+
+        await injectPost(app, mockShortLinkPayload({ item_id: item.id }));
+
+        const response = await injectPost(app, mockShortLinkPayload({ item_id: item2.id }));
```
spaenleh, line 137: To make it more explicit, I would put the `MOCK_ALIAS` in both requests so it is clear that we are trying to re-use the same one.

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+        expectException(response, new ShortLinkDuplication(MOCK_ALIAS));
+      });
+    });
+
+    describe('Not Found', () => {
+      it('Not found if post short links with invalid item id', async () => {
+        const response = await injectPost(app, mockShortLinkPayload());
```
spaenleh, line 144: Here too, maybe specify the itemId you want to use, and add a comment that the item does not exist prior.

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+
+    describe('Authorized Posts', () => {
+      afterEach(async () => {
+        const check = await injectGet(app, MOCK_ALIAS);
+        expect(check.statusCode).toEqual(StatusCodes.OK);
+        expect(check.json().alias).toEqual(MOCK_ALIAS);
+      });
+
+      it('Success if post short links with author member', async () => {
+        const { item } = await mockPostData({ author: actor });
```
spaenleh, line 160: This doe snot read very well, maybe rename `author` to `itemCreator` and the function to something a little more self-explanatory ?

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+  });
+
+  describe('GET /short-links/:itemId', () => {
+    beforeEach(async () => {
+      ({ app, actor } = await build());
+      bob = await saveMember(BOB);
+    });
+
+    it('Not found if get short links with invalid item id', async () => {
+      const response = await injectGet(app, MOCK_ALIAS);
```
spaenleh, line 188: Maybe here use a string like `bubu` instead, to amke it clear that it is not a uuid.

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+      expect(response.statusCode).toEqual(StatusCodes.NOT_FOUND);
+    });
+
+    it('Success even if no connected', async () => {
```
spaenleh, line 192:
```suggestion
    it('Success even if not connected', async () => {
```

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+      const { item } = await mockPostData({ author: actor });
+      const payload = mockShortLinkPayload({ item_id: item.id, alias: MOCK_ALIAS });
+      const post = await injectPost(app, payload);
+      expect(post.statusCode).toEqual(StatusCodes.OK);
+      expect(post.json().alias).toEqual(MOCK_ALIAS);
+
+      logOut(app);
```
spaenleh, line 199: After this, don't you need to resetAllMocks() in an afterEach hook ?

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+      const response = await injectPatch(app, MOCK_ALIAS, { platform: GraaspPlatform.BUILDER });
+      expect(response.statusCode).toEqual(StatusCodes.NOT_FOUND);
+    });
+    describe('Unauthorized and forbidden', () => {
+      it('Unauthorized if patch short links without connected member', async () => {
+        const { item } = await mockPostData({ author: actor });
+        const post = await injectPost(app, mockShortLinkPayload({ item_id: item.id }));
+        expect(post.statusCode).toEqual(StatusCodes.OK);
+
+        logOut(app);
```
spaenleh, line 258: Same, add a resetAllMocks in an afterEach hook

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+        const { item } = await mockPostData({
+          author: actor,
+          member: bob,
+          permission: PermissionLevel.Admin,
+        });
+        const post = await injectPost(app, mockShortLinkPayload({ item_id: item.id }));
+        expect(post.statusCode).toEqual(StatusCodes.OK);
+
+        // sign in as bob
+        ({ app } = await build({ member: bob }));
```
spaenleh, line 291: I dont think you should re-build the app to login as bob ... maybe adding a spy on the attempt authentication would be enough.

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+    });
+
+    describe('Conflict and bad request', () => {
+      it('Conflict if patch short links alias with existing one', async () => {
+        const { item } = await mockPostData({ author: actor });
+        let post = await injectPost(app, mockShortLinkPayload({ item_id: item.id }));
+        expect(post.statusCode).toEqual(StatusCodes.OK);
+        post = await injectPost(
+          app,
+          mockShortLinkPayload({ item_id: item.id, alias: MOCK_ALIAS + 2 }),
```
spaenleh, line 325: Prefer string templates in this case
```suggestion
          mockShortLinkPayload({ item_id: item.id, alias: `${MOCK_ALIAS}2` }),
```

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+        expect(check.statusCode).toEqual(StatusCodes.OK);
+      });
+      it('Unauthorized if delete when not connected', async () => {
+        const { item } = await mockPostData({ author: actor });
+        const payload = mockShortLinkPayload({ item_id: item.id });
+        const post = await injectPost(app, payload);
+        expect(post.statusCode).toEqual(StatusCodes.OK);
+        expect(post.json().alias).toEqual(MOCK_ALIAS);
+
+        logOut(app);
```
spaenleh, line 412: Add the afterEach hook

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+          member: bob,
+          permission: PermissionLevel.Write,
+        });
+        const payload = mockShortLinkPayload({ item_id: item.id });
+        const post = await injectPost(app, payload);
+        expect(post.statusCode).toEqual(StatusCodes.OK);
+        expect(post.json().alias).toEqual(MOCK_ALIAS);
+
+        // log in as bob
+        ({ app } = await build({ member: bob }));
```
spaenleh, line 430: Same comment, not sure this is the way to do it

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+        expect(post.json().alias).toEqual(MOCK_ALIAS);
+
+        ({ app } = await build({ member: bob }));
+
+        const response = await injectDelete(app, MOCK_ALIAS);
+        expect(response.statusCode).toEqual(StatusCodes.OK);
+        expect(response.json().alias).toEqual(MOCK_ALIAS);
+      });
+
+      it('Success when author', async () => {
```
spaenleh, line 460: Actually this works because when you create an item you get admin writes over it. So actually the only case to test is when you have admin on an item, as you can  loose your admin write but still be the creator.

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+        member: bob,
+        permission: PermissionLevel.Admin,
+      }));
+
+      await saveMembership({ item, member: anna, permission: PermissionLevel.Write });
+
+      for (let i = 0; i < numberOfShortLinks; i++) {
+        const response = await injectPost(
+          app,
+          mockShortLinkPayload({ item_id: item.id, alias: MOCK_ALIAS + i }),
```
spaenleh, line 494: Use template here

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+        logOut(app);
+        await attemptFailure(StatusCodes.UNAUTHORIZED);
+      });
+
+      it('Forbidden if unauthorized', async () => {
+        ({ app } = await build({ member: cedric }));
+        await attemptFailure(StatusCodes.FORBIDDEN);
+      });
+
+      it('Forbidden if not admin', async () => {
```
spaenleh, line 522:
```suggestion
      it('Forbidden if write permission only', async () => {
```

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+        const response = await injectGetAll(app, item.id);
+        expect(response.statusCode).toEqual(statusCode);
+      };
+
+      it('Unauthorized when not connected', async () => {
+        logOut(app);
+        await attemptFailure(StatusCodes.UNAUTHORIZED);
+      });
+
+      it('Forbidden if unauthorized', async () => {
```
spaenleh, line 517:
```suggestion
      it('Forbidden if no memberships on item', async () => {
```

src/services/shortLink/test/index.test.ts
spaenleh, line 1: Very nice tests !

src/services/shortLink/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  switch (shortLinkView) {
+    case GraaspPlatform.BUILDER:
+      return BUILDER_HOST;
+    case GraaspPlatform.LIBRARY:
+      return LIBRARY_HOST;
+    case GraaspPlatform.PLAYER:
+      return PLAYER_HOST;
+    default:
+      throw new Error('Invalid Shortlink view');
```
spaenleh, line 16: I see that there is a disconnect between the views we allow in the types and the ones we provide here, not sure if we could write this differently (maybe with a map ?)

src/services/shortLink/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export function getUrlHost({ url }: { name: Context; url: URL }) {
+  const urlStr = url.toString();
+  // if the url ends with /, remove it
+  if (urlStr.endsWith('/')) {
+    return urlStr.slice(0, -1);
+  }
+  return urlStr;
+}
```
spaenleh, line 27: Not sure what the purpose of this is, simply removing the trailing slash from origins ?

src/services/shortLink/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+    case GraaspPlatform.BUILDER:
+      return BUILDER_PREFIX;
+    case GraaspPlatform.LIBRARY:
+      return LIBRARY_PREFIX;
+    case GraaspPlatform.PLAYER:
+      return PLAYER_PREFIX;
+    default:
+      throw new Error('Invalid Shortlink view');
+  }
+}
```
spaenleh, line 43: Do we have something like this in sdk ? If not, maybe this would be something interesting as a source of truth for both front and back.

src/services/shortLink/utils.ts
spaenleh: Maybe see: https://github.com/graasp/graasp-ui/blob/main/src/PlatformSwitch/hooks.ts

src/utils/config.ts
```diff
@@ -300,8 +300,15 @@ export const FILE_ITEM_PLUGIN_OPTIONS = { storageRootPath: FILE_STORAGE_ROOT_PAT
 
 export const ITEMS_ROUTE_PREFIX = '/items';
 export const APP_ITEMS_PREFIX = '/app-items';
+export const SHORT_LINKS_ROUTE_PREFIX = '/short-links';
 export const THUMBNAILS_ROUTE_PREFIX = '/thumbnails';
 
+// Graasp view (builder, player, library) consts
+export const BUILDER_ITEMS_PREFIX = ITEMS_ROUTE_PREFIX;
```
spaenleh, line 307: I think we should decouple the frontend route from the backend route. So use a new constant here, it may have the same value, but the name should clearly state the purpose of the constant.

src/utils/errors.ts
```diff
@@ -397,3 +397,17 @@ export class UnexpectedError extends CoreError {
...
+export class ShortLinkNotFound extends CoreError {
+  constructor(shortLink: string) {
+    const message = `The short link '${shortLink}' not found`;
+    super({ code: 'GERR1009', statusCode: StatusCodes.NOT_FOUND, message: message });
+  }
+}
+
+export class ShortLinkDuplication extends CoreError {
+  constructor(shortLink?: string) {
+    const message = `The short link '${shortLink ?? '<NULL>'}' already exist`;
```
spaenleh, line 410:
```suggestion
    const message = `The short link '${shortLink ?? '<NULL>'}' already exists`;
```

src/utils/errors.ts
```diff
@@ -397,3 +397,17 @@ export class UnexpectedError extends CoreError {
     this.origin = 'unknown';
   }
 }
+
+export class ShortLinkNotFound extends CoreError {
+  constructor(shortLink: string) {
+    const message = `The short link '${shortLink}' not found`;
```
spaenleh, line 403:
```suggestion
    const message = `The short link '${shortLink}' was not found`;
```

src/utils/errors.ts
```diff
@@ -397,3 +397,17 @@ export class UnexpectedError extends CoreError {
...
+    super({ code: 'GERR1009', statusCode: StatusCodes.NOT_FOUND, message: message });
+  }
+}
+
+export class ShortLinkDuplication extends CoreError {
+  constructor(shortLink?: string) {
```
spaenleh, line 409: Why is it optional ?

spaenleh (changes requested):
Thank you for this ! 💯 With tests and everything ! I would like @pyphilia to also have a look.
Maybe it would be nice to have a short description of the feature next to the code (README.md inside the service folder ?) that just explains the general mechanisms of the feature and how it should be used.

Questions:
- should we use the `itemId` or the `itemPath` to link to the item entity ?
- should we specify in the schemas the shape of what is returned ?

src/utils/errors.ts
ReidyT: Oups, it should not :)

src/services/shortLink/test/index.test.ts
ReidyT: The problem is if I'm not using a good UUID format, I will have a bad request. What I want to test here is to be sure to receive a 404 error if an itemId not exist, and not the format of the given UUID.

ReidyT: The problem is if I'm not using a good UUID format, I will have a bad request. What I want to test here is to be sure to receive a 404 error if an itemId not exist, and not the format of the given UUID.

src/services/shortLink/entities/ShortLink.ts
```diff
@@ -0,0 +1,40 @@
...
+  alias: string;
+
+  @Column({ type: 'enum', enum: GraaspPlatform, nullable: false })
+  platform: GraaspPlatform;
+
+  @CreateDateColumn({ type: 'timestamp', default: () => 'NOW()' })
+  createdAt: Date;
+
+  @Column({ type: 'uuid', nullable: false })
+  item_id: string;
```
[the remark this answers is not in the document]
ReidyT: In fact, I need this column to allow to insert a new short link with the item id without having to give the complete Item. Without this column, I'm not able to insert short link like 
`{
    "alias": string,
    "platform": Context,
    "itemId": UUID
}`.

src/services/shortLink/test/index.test.ts
ReidyT: Normally, there is a `jest.clearAllMocks();` in an `afterEach` at top of the test file.

src/services/shortLink/test/index.test.ts
ReidyT: You're right, I will remove this duplicated test.

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+      const response = await injectGet(app, MOCK_ALIAS);
+      expect(response.statusCode).toEqual(StatusCodes.NOT_FOUND);
+    });
+
+    it('Success even if no connected', async () => {
+      const { item } = await mockPostData({ author: actor });
+      const payload = mockShortLinkPayload({ item_id: item.id, alias: MOCK_ALIAS });
+      const post = await injectPost(app, payload);
+      expect(post.statusCode).toEqual(StatusCodes.OK);
+      expect(post.json().alias).toEqual(MOCK_ALIAS);
```
pyphilia, line 197: I like to have a separate section for public tests, what do you think of moving this out? That way you could also start with being disconnected in the first place instead of logging out in the middle of your test.

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+        const { item } = await mockPostData({
+          author: actor,
+          member: bob,
+          permission: PermissionLevel.Write,
+        });
+        const post = await injectPost(app, mockShortLinkPayload({ item_id: item.id }));
+        expect(post.statusCode).toEqual(StatusCodes.OK);
+
+        // sign in as bob
+        ({ app } = await build({ member: bob }));
```
pyphilia, line 273: You build twice the app?

src/services/shortLink/test/index.test.ts
pyphilia: is this resolved? The code is still there.

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+    describe('Unauthorized and forbidden', () => {
+      const attemptFailure = async (statusCode: StatusCodes) => {
+        const response = await injectGetAll(app, item.id);
+        expect(response.statusCode).toEqual(statusCode);
+      };
+
+      it('Unauthorized when not connected', async () => {
+        logOut(app);
+        await attemptFailure(StatusCodes.UNAUTHORIZED);
+      });
```
pyphilia, line 515: What about a separate describe with it's own `app = build()`? so you don't have a test where you login and logout right at the beginning.

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+
+      it('Forbidden if unauthorized', async () => {
+        ({ app } = await build({ member: cedric }));
```
pyphilia, line 518: App is built twice?

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+        await attemptFailure(StatusCodes.FORBIDDEN);
+      });
+
+      it('Forbidden if not admin', async () => {
+        ({ app } = await build({ member: anna }));
```
pyphilia, line 523: App is built twice?

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+        const response = await injectGetAll(app, item.id);
+        expect(response.statusCode).toEqual(StatusCodes.OK);
+        expect(response.json()).toEqual([]);
+      });
+
+      const attemptSuccess = async () => {
+        const response = await injectGetAll(app, item.id);
+        expect(response.statusCode).toEqual(StatusCodes.OK);
+        expect(response.json()).toHaveLength(numberOfShortLinks);
+      };
```
pyphilia, line 540: Don't add a function in the middle of the describe scope. I think it's okay to c/p this code twice.

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+
+      it('Success if author', async () => {
+        await attemptSuccess();
+      });
+
+      it('Success if admin', async () => {
+        ({ app } = await build({ member: bob }));
```
pyphilia, line 547: How many times is the app build here?

src/services/shortLink/service.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    return await shortLinkRepository.getItem(itemId);
+  }
+
+  async getRealPath(alias: string, repositories: Repositories) {
+    const shortLink = await this.get(alias, repositories);
+    const host = getPlatformHost(shortLink.platform);
+    const prefix = getPlatformPrefix(shortLink.platform);
+    return {
+      url: `${getUrlHost(host)}${prefix}/${shortLink.item_id}`,
```
pyphilia, line 47: Maybe here you should use `URL` to build your link?

pyphilia (changes requested):
Thanks for your work! Do you think you could write a short description of the feature (french or english) so we can add that in some tutorials/FAQ later? Thanks!

I have some concerns for naming especially `item_id` (I think you asked about it yesterday we can discuss tomorrow or by zoom). Also stick with `actor` instead of `author`. 
Also I think your tests build too many times the app, I worry it's actually running the apps in the background and never closing them since you reassign the `app` variable each time.

src/services/shortLink/index.ts
```diff
@@ -0,0 +1,78 @@
...
+    );
+
+    fastify.post<{ Body: ShortLink }>(
+      '/',
+      {
+        schema: create,
+        preHandler: fastify.verifyAuthentication,
+      },
+      async ({ member, body: shortLink }, reply) => {
+        const newLink = await shortLinkService.post(member, shortLink, buildRepositories());
```
pyphilia, line 43: Add a transaction wrapper.

src/services/shortLink/index.ts
```diff
@@ -0,0 +1,78 @@
...
+      },
+    );
+
+    fastify.delete<{ Params: { alias: string } }>(
+      '/:alias',
+      {
+        preHandler: fastify.verifyAuthentication,
+      },
+      async ({ member, params: { alias } }, reply) => {
+        const oldLink = await shortLinkService.delete(member, alias, buildRepositories());
```
pyphilia, line 54: Add a transaction wrapper.

src/services/shortLink/index.ts
```diff
@@ -0,0 +1,78 @@
...
+    );
+
+    fastify.patch<{ Params: { alias: string }; Body: ShortLink }>(
+      '/:alias',
+      {
+        schema: update,
+        preHandler: fastify.verifyAuthentication,
+      },
+      async ({ member, body: shortLink, params: { alias } }, reply) => {
+        const updatedLink = await shortLinkService.update(
```
pyphilia, line 66: Add a transaction wrapper.

pyphilia (changes requested):

src/services/shortLink/test/fixtures.ts
```diff
@@ -0,0 +1,116 @@
...
+  return { item };
+};
+
+const inject = async (app, method: HTTPMethods, { url, payload }: { url; payload? }) => {
+  return await app.inject({
+    method,
+    url,
+    payload,
+  });
+};
```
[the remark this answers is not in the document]
ReidyT: You're right, I will remove it.

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+    });
+
+    describe('Authorized Posts', () => {
+      afterEach(async () => {
+        const check = await injectGet(app, MOCK_ALIAS);
+        expect(check.statusCode).toEqual(StatusCodes.OK);
+        expect(check.json().alias).toEqual(MOCK_ALIAS);
+      });
+
+      it('Success if post short links with author member', async () => {
```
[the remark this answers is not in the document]
ReidyT: As @spaenleh suggested, testing the author is equivalent to testing member with admin permission, so I will remove the author's tests.

src/services/shortLink/test/index.test.ts
```diff
@@ -0,0 +1,552 @@
...
+    });
+
+    describe('Unauthorized and forbidden access', () => {
+      afterEach(async () => {
+        const check = await injectGet(app, MOCK_ALIAS);
+        expect(check.statusCode).toEqual(StatusCodes.NOT_FOUND);
+      });
+
+      it('Unauthorized if post short links without connected member', async () => {
+        ({ app } = await build({ member: null }));
```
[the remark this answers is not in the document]
ReidyT: Yes I'm sorry, I haven't push the modifications yet, but I have update this.

src/services/shortLink/service.ts
ReidyT: Yes, using URL is easier, thank you 👍 .

src/services/shortLink/utils.ts
ReidyT: The new class (maybe Singleton) should solve this.

src/services/shortLink/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+  // TODO: set as const
+  const [BUILDER_PREFIX, PLAYER_PREFIX, LIBRARY_PREFIX] = CLIENT_PREFIX;
+
+  switch (shortLinkView) {
+    case GraaspPlatform.BUILDER:
+      return BUILDER_PREFIX;
+    case GraaspPlatform.LIBRARY:
+      return LIBRARY_PREFIX;
+    case GraaspPlatform.PLAYER:
+      return PLAYER_PREFIX;
```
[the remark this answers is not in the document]
ReidyT: The Singleton should solve this

src/services/shortLink/repository.ts
```diff
@@ -0,0 +1,71 @@
...
+  async updateOne(alias: string, postLink: ShortLinkPatchPayload): Promise<ShortLink> {
+    try {
+      if (!alias) throw new ShortLinkNotFound(alias);
+
+      await this.update(alias, postLink);
+      const aliasId = postLink['alias'] ?? alias;
+      return await this.get(aliasId); // return the updated short link
+    } catch (ex) {
+      // if a duplication entry error throw, the postlink must contain the alias
+      throw handleDuplicateEntryError(ex, new ShortLinkDuplication(postLink['alias']));
```
pyphilia, line 68: Does it have to be a duplicate error?

src/services/shortLink/service.ts
```diff
@@ -0,0 +1,86 @@
...
+import { Actor } from '../member/entities/member';
+
+export class ShortLinkService {
+  private itemService: ItemService;
+
+  public constructor(itemService: ItemService) {
+    this.itemService = itemService;
+  }
+
+  async post(member: Actor, repositories: Repositories, shortLink: ShortLinkPostPayload) {
```
pyphilia, line 21: Enforce `Member`
```suggestion
  async post(member: Member, repositories: Repositories, shortLink: ShortLinkPostPayload) {
```

src/services/shortLink/service.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    return shortLinkRepository.createOne(shortLink);
+  }
+
+  async getOne(repositories: Repositories, alias: string) {
+    const { shortLinkRepository } = repositories;
+    return shortLinkRepository.get(alias);
+  }
+
+  async getAllForItem(member: Actor, repositories: Repositories, itemId: string) {
```
pyphilia, line 36:
```suggestion
  async getAllForItem(member: Member, repositories: Repositories, itemId: string) {
```

src/services/shortLink/service.ts
```diff
@@ -0,0 +1,86 @@
...
+  async getRedirection(repositories: Repositories, alias: string) {
+    const shortLink = await this.getOne(repositories, alias);
+    const clientHostManager = ClientHostManager.getInstance();
+
+    return clientHostManager
+      .getItemLink(shortLink.platform as Context, shortLink.item.id)
+      .toString();
+  }
+
+  async delete(member: Actor, repositories: Repositories, alias: string) {
```
pyphilia, line 55:
```suggestion
  async delete(member: Member, repositories: Repositories, alias: string) {
```

src/services/shortLink/service.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    return shortLinkRepository.getItem(itemId);
+  }
+
+  async getRedirection(repositories: Repositories, alias: string) {
+    const shortLink = await this.getOne(repositories, alias);
+    const clientHostManager = ClientHostManager.getInstance();
+
+    return clientHostManager
+      .getItemLink(shortLink.platform as Context, shortLink.item.id)
```
pyphilia, line 51: I guess you're receiving a `URL` object. Maybe it's clearer if the function name is `getItemAsURL`. And maybe provide also `getItemLink` that returns a string link?

src/services/shortLink/service.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    // check that the member can admin the item to be allowed to create short link
+    await this.itemService.get(member, repositories, shortLink.item.id, PermissionLevel.Admin);
+
+    await shortLinkRepository.deleteOne(alias);
+    return shortLink;
+  }
+
+  async update(
+    member: Actor,
```
pyphilia, line 69:
```suggestion
    member: Member,
```

src/services/shortLink/test/fixtures.ts
```diff
@@ -0,0 +1,126 @@
...
+  itemId,
+}: {
+  alias: string;
+  platform?: Context;
+  itemId: string;
+}) => {
+  return {
+    alias,
+    platform,
+    item: { id: itemId },
```
pyphilia, line 28: Wouldn't it be clearer if we use `itemId` instead of `item: {id}`?

src/utils/errors.ts
```diff
@@ -397,3 +397,42 @@ export class UnexpectedError extends CoreError {
...
+      shortLink,
+    );
+  }
+}
+
+export class ShortLinkDuplication extends CoreError {
+  constructor(shortLink: string) {
+    super(
+      {
+        code: 'GERR1010',
```
pyphilia, line 418: Actually create a new "errors" file and extend the error from another factory. So the origin can be short link instead of unknown. There are examples in the repo.

src/utils/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const SHORT_LINKS_ROUTE_PREFIX = '/short-links';
```
pyphilia, line 1: Should this file be in `services/shortLink`?

src/utils/typeormError.ts
```diff
@@ -1,2 +1,21 @@
...
+export const handleDuplicateEntryError = (thrownError: Error, duplicationError: Error) => {
+  if (
+    thrownError instanceof QueryFailedError &&
+    thrownError.driverError.code === DUPLICATE_ENTRY_ERROR_CODE
+  ) {
+    return duplicationError;
+  }
+
+  // if the thrown error is not due to duplication, throw it to the next
+  return thrownError;
```
pyphilia, line 20: I see know, maybe I would rather have 
```js
catch (e) {
  if(isDuplicateEntryError(e)) {
    throw ShortLinkDuplicateError()
  } 
  throw e
}
```

pyphilia (changes requested):
Thanks for the PR! Let me know if you have questions 🪜

src/services/shortLink/test/fixtures.ts
ReidyT: The payload returned by this function is used in an API POST call. The body of the API must contain `item: {id}` so I think it is simpler to return the payload as it must be send to the API. But maybe I missed something ?

src/utils/constants.ts
ReidyT: Yes, it could be. I'll move their in shortLink service.

PR #673, opened 2023-11-02
fix: filter sentry events at the source

spaenleh:
In this PR I add an option to filter events we send to sentry.


pyphilia (approved):
LGTM 👍

PR #680, opened 2023-11-03
fix: update to sdk 2.0.0

### LINKED ISSUE #679: Check whether childrenOrder exists
Because of legacy data, the following fails if the folder does not have a defined order.

https://github.com/graasp/graasp/blob/ac346f70a8a498077f2edd682ba9842108a68a64/src/services/item/service.ts#L108

spaenleh:
- fix: mismatched types from sdk update
- fix: use optional on folder extra that might be empty fix #679

### DIFF
--- a/src/services/item/plugins/published/plugins/search/meilisearch.ts
+++ b/src/services/item/plugins/published/plugins/search/meilisearch.ts
@@ -163,8 +163,9 @@ export class MeiliSearchWrapper {
       content: await this.getContent(item),
       isPublishedRoot: isPublishedRoot,
       isHidden: isHidden,
-      createdAt: item.createdAt,
-      updatedAt: item.updatedAt,
+      // todo: fix these types
+      createdAt: item.createdAt.toISOString(),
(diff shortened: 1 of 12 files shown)


ReidyT (approved):

src/services/item/plugins/published/plugins/search/meilisearch.ts
```diff
@@ -163,8 +163,9 @@ export class MeiliSearchWrapper {
       content: await this.getContent(item),
       isPublishedRoot: isPublishedRoot,
       isHidden: isHidden,
-      createdAt: item.createdAt,
-      updatedAt: item.updatedAt,
+      // todo: fix these types
```
pyphilia, line 166: Still need to be fixed?

pyphilia (approved):

src/services/item/plugins/published/plugins/search/meilisearch.ts
spaenleh: Yes it uses the type from sdk and it expects a string and we give it a Date ...

PR #687, opened 2023-11-10
fix: add missing client hosts and move workflows back here

spaenleh:
This PR refactors the deployment workflows

This PR adds the missing client hosts to support redirection when logging in from `Account` or `Analytics` with the `?url` param.
We also:
- rename deploy workflows
- upgrade node version set in package.json
- upgrade yarn

Variables and workflows have been migrated in:
- [x] dev
- [x] stage
- [x] prod


pyphilia (approved):
LGTM ⚜️

PR #700, opened 2023-11-20
feat: import users with CSV and create file structure based on groups in CSV

marcelmmc:
This PR include the following modifications:
- Adds a route to handle parsing and invitations using a CSV file in the back-end.
- If the CSV contains a ```group_name``` column field, this route will create group folders and invite their corresponding members to these."


src/services/invitation/index.ts
```diff
@@ -55,6 +119,109 @@ const plugin: FastifyPluginAsync = async (fastify, options) => {
...
+            const r = await result;
+            let group_name = '';
+            if (hasGrpCol) group_name = rowInvite.group_name as string;
+            const invite = new Invitation();
+            invite.email = rowInvite['email'];
+            invite.permission = PermissionLevel.Read;
+            if (rowInvite.permission) invite.permission = rowInvite.permission;
+            if (!rowInvite.name) invite.name = '';
+            if (hasGrpCol) invite.item = groupNameToItem.get(group_name) as Item;
+            else invite.item = item;
```
swouf, line 193: 👍🏼

src/services/invitation/index.ts
```diff
@@ -55,6 +119,109 @@ const plugin: FastifyPluginAsync = async (fastify, options) => {
...
+              r.set(group_name, [invite]);
+            } else {
+              r.get(group_name)?.push(invite);
+            }
+            return r;
+          },
+          Promise.resolve(new Map<string, Invitation[]>()),
+        );
+        for await (const arrOfInvitations of invitesPerGroup.values()) {
+          // Perform some asynchronous operation
```
swouf, line 204: Which one? Be more specific.

src/services/invitation/index.ts
```diff
@@ -55,6 +119,109 @@ const plugin: FastifyPluginAsync = async (fastify, options) => {
...
+      // allow some fields for app data and app setting
+      fileSize: 1024 * 1024 * 1024, // For multipart forms, the max file size (Default: Infinity).
+      files: 1024 * 1024 * 1024 * 1, // Max number of file fields (Default: Infinity).
+      // headerPairs: 2000             // Max number of header key=>value pairs (Default: 2000 - same as node's http).
+    },
+  });
+
+  fastify.route<{ Querystring: IdParam; Body: any }>({
+    method: HttpMethod.POST,
+    url: '/invitations/:id/upload_csv',
```
swouf, line 136: Given the current schema for the invitations, I would suggest to stick to it. See line 237 (patch method): `'/:id/invitations/:invitationId'`

I think you should use something like: `':id/invitations/upload_csv'`.

Maybe @pyphilia or @spaenleh have an opinion on that?

src/services/invitation/index.ts
```diff
@@ -1,16 +1,80 @@
...
+type CSVInvite = {
+  email: string;
+  name?: string;
+  group_name?: string;
+  permission?: PermissionLevel;
+};
+
+function createNewFolder(member: Actor, rowInvite: CSVInvite, parentId: string): Promise<Item> {
+  return new Promise((resolve) => {
+    const buildId = uuidv4();
```
swouf, line 43: Do you have to generate the uuid? I thought that would be given by the item service which itself gets it from the DB or something like that. Could you check or confirm, please?

src/services/invitation/index.ts
```diff
@@ -1,16 +1,80 @@
...
+    let buildPath = buildPathFromIds(buildId);
+    if (parentId) buildPath = `${parentId}.${buildPath}`;
```
swouf, line 45: I guess you wouldn't need to specify the path if the uuid is provided by the item service. See my comment above.

src/services/invitation/index.ts
```diff
@@ -55,6 +119,109 @@ const plugin: FastifyPluginAsync = async (fastify, options) => {
...
+      const {
+        member,
+        query: { id: parentId },
+        log,
+      } = request;
+      const errors: Error[] = [];
+      // check rights
+      try {
+        if (!parentId) {
+          throw new Error('User has passed an item located at its root filesystem');
```
swouf, line 148: I would use a more explicit message.

```suggestion
          throw new Error('User is sharing its root which is forbidden.');
```

swouf (dismissed):
Thanks Marcel for this PR. I made a few comments and suggestions. Please, feel free to disagree with my comments and tag @pyphilia or @spaenleh in the conversation if you would like a more expert opinion. 👩🏼‍💻👨🏼‍💻

Overall, that's a nice implementation 👏🏼 and I don't have much to add.

I just have one question❓

In the current implementation (the one in prod), it's the frontend that checks the existence of members and sends invitations or create memberships based on the existence of accounts (see [this mutation](https://github.com/graasp/graasp-query-client/blob/5673ec7ce02ba53af2b53dd965b880edb338499f/src/mutations/membership.ts#L147C1-L274) in the query-client). It seems here that you are only sending invitations. Does it add the memberships directly if the account already exist?

If it doesn't, it would be a good thing to change the behaviour of the invitations (or of the sharing mechanism) to have a single endpoint to manage the invitations/memberships that does the management of creating memberships or sending invitations in a transparent way. We can discuss that in a future meeting.

Good work!

src/services/invitation/index.ts
```diff
@@ -55,6 +119,109 @@ const plugin: FastifyPluginAsync = async (fastify, options) => {
...
+          throw new Error('An CSV file was not uploaded');
+        }
+        const { service: itemService } = items;
+        const stream = file.file;
+
+        const { rows, header } = await getCSV(stream);
+        const hasGrpCol = header.includes('group_name');
+        let groupNameToItem = new Map<string, Item>();
+        if (hasGrpCol) {
+          groupNameToItem = await rows.reduce(
```
spaenleh, line 166: Not sure this needs to be an async reduce...

src/services/invitation/index.ts
```diff
@@ -55,6 +119,109 @@ const plugin: FastifyPluginAsync = async (fastify, options) => {
...
+                r.set(rowInvite.group_name as string, itemCreated);
+              }
+              return r;
+            },
+            Promise.resolve(new Map<string, Item>()),
+          );
+        }
+        const invitesPerGroup = await rows.reduce(
+          async (result: Promise<Map<string, Invitation[]>>, rowInvite: CSVInvite) => {
+            const r = await result;
```
spaenleh, line 184: Why ??

src/services/invitation/index.ts
```diff
@@ -55,6 +119,109 @@ const plugin: FastifyPluginAsync = async (fastify, options) => {
...
+            if (hasGrpCol) invite.item = groupNameToItem.get(group_name) as Item;
+            else invite.item = item;
+            if (!r.has(group_name)) {
+              r.set(group_name, [invite]);
+            } else {
+              r.get(group_name)?.push(invite);
+            }
+            return r;
+          },
+          Promise.resolve(new Map<string, Invitation[]>()),
```
spaenleh, line 201: Why ?

spaenleh (commented):
It looks like you have put all the concerns in the same place. 
We usually divide the logic in 3 parts in the backend:
- index: defines routes and calls the service
- service: defines business logic i.e. parsing a csv file and checking access rights
- repository: handles database related logic, creating the item given some parameters (parentId, etc)

We already have some repository methods that you can use to create an Item (post).

Secondly, it looks like you first create the items, and then create the invitations inside a transaction. Does this mean that if the operation fails halfway, you will have created a partial hierarchy of items ? I would prefer that you first plan everything that must be created and then create everything inside the same transaction so if anything goes wrong it rolls back everything.

src/services/invitation/index.ts
marcelmmc: I checked the query-client, and I just realized that I do not need to create an item object to pass to the item service, I removed this function as it is not needed anymore.

src/services/invitation/index.ts
```diff
@@ -55,6 +119,109 @@ const plugin: FastifyPluginAsync = async (fastify, options) => {
...
+              arrOfInvitations,
+            );
+          });
+        }
+      } catch (e) {
+        errors.push(e);
+      }
+
+      return {
+        data: [],
```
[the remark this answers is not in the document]
marcelmmc: Oh, sorry. I was not sure what to return and how to handle the data in the front-end, I thought the generated invitations should be appended to the data array and then these should be displayed in the pending Invitations ui section, could I get a confirmation from @swouf or @spaenleh?

src/services/invitation/index.ts
marcelmmc: But I would like to wait for the creation of folders, before creating the invitations, could you give me some advice on how to do this part?

src/services/invitation/index.ts
marcelmmc: I have removed the async from this part, please let me know if this part still concerns you

src/services/invitation/index.ts
marcelmmc: I have removed the async from this part, please let me know if this part still concerns you

src/services/invitation/index.ts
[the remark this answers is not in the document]
spaenleh: @marcelmmc Displaying invitations is done using a different `GET` call. This return should return data specific to the action performed.

src/services/invitation/index.ts
spaenleh: In case your async operation of creating folders can be launched at the same time the pattern would be to do an
```ts
await Promise.all(yourListOfCallToCreateItems)
```
This will allow you to wait until all promises in the list have resolved and continue processing after.

src/schemas/global.ts
```diff
@@ -22,6 +22,7 @@ export default {
       required: ['id'],
       properties: {
         id: { $ref: '#/definitions/uuid' },
+        templateId: { $ref: '#/definitions/uuid' },
```
spaenleh, line 25: Not sure this needs to be here. Which template id this the id of, if this specific to the task of a creating the templates ?

src/services/invitation/index.ts
```diff
@@ -55,6 +57,38 @@ const plugin: FastifyPluginAsync = async (fastify, options) => {
...
 
+  fastify.register(fastifyMultipart, {
+    limits: {
+      fields: MAX_NON_FILE_FIELDS, // Max number of non-file fields (Default: Infinity).
+      fileSize: MAX_FILE_SIZE, // For multipart forms, the max file size (Default: Infinity).
+      files: MAX_FILES, // Max number of file fields (Default: Infinity).
+    },
+  });
+
+  fastify.route<{ Querystring: IdParam; Body: any }>({
```
spaenleh, line 68: If you do not need the body please do not type it as `any` remove it from the generic typing.

src/services/invitation/index.ts
```diff
@@ -55,6 +57,38 @@ const plugin: FastifyPluginAsync = async (fastify, options) => {
...
+    handler: async (request, reply) => {
+      const { member, query, log } = request;
+      const errors: Error[] = [];
+      try {
+        const file = await request.file();
+        await db.transaction(async (manager) => {
+          await iS.handleCSVInvitations(member, buildRepositories(manager), query, file);
+        });
+      } catch (e) {
+        console.log(e);
```
spaenleh, line 81: Use fastify-provided logger instead to comply with the logging system.

src/services/invitation/index.ts
```diff
@@ -55,6 +57,38 @@ const plugin: FastifyPluginAsync = async (fastify, options) => {
...
+        errors.push(e);
+      }
+      return {
+        // TO-DO: report invitations in data
+        data: [],
```
spaenleh, line 86: This return shape is what we call a `ResultOf<T>`. It is used when we perform a "many" operation, for example asking for multiple memberships at a time. We use this shape to be able to reply with a partial set of the data and maybe some errors. In your case, I think that the operation should not be considered partial, it should either succeed or fail. So you should simply return the data if it succeeded and an error if it failed.

spaenleh: The issue becomes very apparent when you see that you will always at most push one error in your error array, which will just mean that the whole thing failed.

src/schemas/global.ts
spaenleh: Remove it from here and update the local schema to include the template_id param

src/services/invitation/service.ts
```diff
@@ -136,4 +140,98 @@ export class InvitationService {
     const memberships = invitations.map(({ permission, item }) => ({ item, member, permission }));
     await itemMembershipRepository.createMany(memberships);
   }
+
+  async handleCSVInvitations(
+    actor: Actor,
+    repositories: Repositories,
+    query: IdParam,
```
spaenleh, line 147:
```suggestion
    query: IdParam & {tempalte_id: string},
```

src/services/invitation/service.ts
```diff
@@ -136,4 +140,98 @@ export class InvitationService {
...
+    const { id: parentId, template_id } = query;
+    const parentItem = await repositories.itemRepository.get(parentId);
+    await validatePermission(repositories, PermissionLevel.Admin, actor, parentItem);
+
+    if (!file || file.mimetype != CSV_MIMETYPE) {
+      throw new Error(`
+          An incorrect type of file has been uploaded, 
+          Please upload a file with .csv extension
+        `);
+    }
```
spaenleh, line 159: Moving this check as the first thing in the function above will improve performance, as you should not even try to validate the item existence if you did not get a file.

src/services/invitation/service.ts
```diff
@@ -136,4 +140,98 @@ export class InvitationService {
...
+              await this.itemService.copyMany(actor, repositories, itemsAtFirstLevel, {
+                parentId: itemCreated.id,
+              });
+            }
+            r.set(rowInvite.group_name as string, itemCreated);
+          }
+          return r;
+        },
+        Promise.resolve(new Map<string, Item>()),
+      );
```
spaenleh, line 207: Please refactor this reduce to create an execution plan synchronously (no await) and then perform the async actions. You could create a map with some options that tels you what you will have to do. Also add comments to explain why you do certain things as their purpose is not easy to understand.

src/services/invitation/service.ts
```diff
@@ -136,4 +140,98 @@ export class InvitationService {
...
+    }
+
+    const invitesPerGroup = rows.reduce((result, rowInvite: CSVInvite) => {
+      let group_name = '';
+      if (hasGrpCol) group_name = rowInvite.group_name as string;
```
spaenleh, line 212: Readability: Do not write inline `if` expressions as they are hard to read. Always put the curly braces even if it only encloses a single statement

src/services/invitation/service.ts
```diff
@@ -136,4 +140,98 @@ export class InvitationService {
...
+      const invite = new Invitation();
+      invite.email = rowInvite['email'];
+      invite.permission = PermissionLevel.Read;
+      if (rowInvite.permission) invite.permission = rowInvite.permission;
+      if (!rowInvite.name) invite.name = '';
+      if (hasGrpCol) invite.item = groupNameToItem.get(group_name) as Item;
+      else invite.item = parentItem;
```
spaenleh, line 219: Same here.

src/services/invitation/service.ts
```diff
@@ -136,4 +140,98 @@ export class InvitationService {
...
+
+    const invitesPerGroup = rows.reduce((result, rowInvite: CSVInvite) => {
+      let group_name = '';
+      if (hasGrpCol) group_name = rowInvite.group_name as string;
+      const invite = new Invitation();
+      invite.email = rowInvite['email'];
+      invite.permission = PermissionLevel.Read;
+      if (rowInvite.permission) invite.permission = rowInvite.permission;
+      if (!rowInvite.name) invite.name = '';
+      if (hasGrpCol) invite.item = groupNameToItem.get(group_name) as Item;
```
spaenleh, line 218: Not sure why you explicitly use a cast here ? If possible do not use explicit casts.

src/services/invitation/service.ts
```diff
@@ -136,4 +140,98 @@ export class InvitationService {
...
+    }, new Map<string, Invitation[]>());
+
+    for await (const arrOfInvitations of invitesPerGroup.values()) {
+      await this.postManyForItem(
+        actor,
+        repositories,
+        arrOfInvitations[0].item.id,
+        arrOfInvitations,
+      );
+    }
```
spaenleh, line 235: You can use a `await Promise.all(invitesPerGroup.values().map( async (i) => this.postManyForItem(...)))` instead.

src/services/invitation/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+      error(err, file) {
+        reject(err);
+      },
+    });
+  });
+};
+
+export const regexGen = (firstLvlPath: string) => {
+  return RegExp(`${firstLvlPath}\.[a-zA-Z0-9_]+$`);
+};
```
spaenleh, line 36: Is this supposed to check whether an item is a descendant of the item passed as argument ? Make this purpose clear in the function name please.

spaenleh (changes requested):
Thank you for the changes, see my new comments.

API Collections/Invitations/Upload users csv.bru
```diff
@@ -0,0 +1,15 @@
...
+}
+
+post {
+  url: {{host}}/items/{{item_id}}/invitations/upload-csv
+  body: multipartForm
+  auth: inherit
+}
+
+body:multipart-form {
+  : @file(/Users/basile/Downloads/users.csv)
```
pyphilia, line 14: Change.

API Collections/Invitations/Upload Template users csv.bru
```diff
@@ -0,0 +1,15 @@
...
+}
+
+post {
+  url: {{host}}/items/{{item_id}}/invitations/upload-csv
+  body: multipartForm
+  auth: inherit
+}
+
+body:multipart-form {
+  : @file(/Users/basile/Downloads/users.csv)
```
pyphilia, line 14: Change.

src/services/item/plugins/invitation/errors.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+}
+
+export class NoDataInFile extends GraaspInvitationsError {
+  constructor(data?: unknown) {
+    super(
+      {
+        code: 'GPINVERR010',
+        statusCode: StatusCodes.BAD_REQUEST,
+        message: 'No data was found in teh file. Please send a file with valid data.',
```
pyphilia, line 146:
```suggestion
        message: 'No data was found in the file. Please send a file with valid data.',
```

src/services/item/plugins/invitation/errors.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+
+export class CantCreateStructureInNoFolderItem extends GraaspInvitationsError {
+  constructor(data?: unknown) {
+    super(
+      {
+        code: 'GPINVERR011',
+        statusCode: StatusCodes.BAD_REQUEST,
+        message:
+          'Provided item is not a folder. A structure can not be created inside an item that is not a folder.',
```
pyphilia, line 160:
```suggestion
          'Provided item is not a folder. A structure cannot be created inside an item that is not a folder.',
```

src/services/item/plugins/invitation/index.ts
```diff
@@ -54,6 +58,67 @@ const plugin: FastifyPluginAsync = async (fastify) => {
     },
   );
 
+  // register upload endpoint separately so the multipart otptions only apply to that specific endpoint
```
pyphilia, line 68:
```suggestion
  // register upload endpoint separately so the multipart options only apply to that specific endpoint
```

src/services/item/plugins/invitation/index.ts
```diff
@@ -54,6 +58,67 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+  fastify.register(async (fastify) => {
+    fastify.register(fastifyMultipart, {
+      limits: {
+        fields: MAX_NON_FILE_FIELDS, // Max number of non-file fields (Default: Infinity).
+        fileSize: MAX_FILE_SIZE, // For multipart forms, the max file size (Default: Infinity).
+        files: MAX_FILES, // Max number of file fields (Default: Infinity).
```
pyphilia, line 74: set to one?

src/services/item/plugins/invitation/index.ts
```diff
@@ -54,6 +58,67 @@ const plugin: FastifyPluginAsync = async (fastify) => {
     },
   );
 
+  // register upload endpoint separately so the multipart otptions only apply to that specific endpoint
+  fastify.register(async (fastify) => {
+    fastify.register(fastifyMultipart, {
+      limits: {
+        fields: MAX_NON_FILE_FIELDS, // Max number of non-file fields (Default: Infinity).
```
pyphilia, line 72: Do you need fields?

src/services/item/plugins/invitation/index.ts
```diff
@@ -54,6 +58,67 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+            iS.handleCSVInvitations(
+              member,
+              buildRepositories(manager),
+              itemId,
+              templateId,
+              uploadedFile,
+              memberships.service,
+            ),
+          );
+        } else {
```
pyphilia, line 114: No need for `else`.

src/services/item/plugins/invitation/repository.ts
```diff
@@ -84,16 +87,27 @@ export const InvitationRepository = AppDataSource.getRepository(Invitation).exte
...
           ...invitations,
           item: { path: itemPath },
           creator,
         })),
     );
-    // TODO: optimize
-    return this.getMany(insertResult.identifiers.map(({ id }) => id));
+
+    const ids = insertResult.identifiers.map(({ id }) => id);
+    console.log(ids);
```
pyphilia: Remove.

src/services/item/plugins/invitation/service.ts
```diff
@@ -0,0 +1,358 @@
...
+  TemplateItemDoesNotExist,
+} from './errors';
+import { CSVInvite, parseCSV, verifyCSVFileFormat } from './utils';
+
+export class InvitationService {
+  log: FastifyBaseLogger;
+  mailer: MailerDecoration;
+  itemService: ItemService;
+  memberService: MemberService;
+  // itemMembershipService: ItemMembershipService;
```
pyphilia, line 38: Remove.

pyphilia (changes requested):

src/services/item/plugins/invitation/index.ts
```diff
@@ -49,11 +53,79 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+      async (request) => {
+        const { member, query, params } = request;
+        // this is needed to assert the type of the member to be a Member and not an Actor.
+        // verifyAuthentication preHandler should throw if there is no member, but the type can not be narrowed automatically
+        if (!member) {
+          throw new UnauthorizedMember(member);
+        }
+
+        // We need to get the membership service here because it is defined after the invitation service
+        const { memberships } = fastify;
```
pyphilia, line 90: You can't invert the registering order to prevent this?

src/services/item/plugins/invitation/specifications.md
```diff
@@ -0,0 +1,101 @@
...
+### Add users from a CSV
+
+Route: `POST /items/:id/invitations/upload-csv`
+Body: Multipart Form containing a csv file.
+
+#### CSV File specifications
+
+| Column Name | Description | Required | Default Value |
+|---|---|---|---|
+| `email` | Email adress of the user | Yes | None (required) |
```
pyphilia, line 26:
```suggestion
| `email` | Email address of the user | Yes | None (required) |
```

src/services/item/plugins/invitation/specifications.md
```diff
@@ -0,0 +1,101 @@
...
+#### Invitations
+
+An email that is already invited on an item will be droped. 
+
+Invitations can be set on parents or children of an item that already holds an invitation for the member but they need to follow membership inheritance rules:
+- A children membership can not be more restrictive (lower permission) than its parent memberships
+- A children membership can be more permissive (higher permission) than its parent membership
+
+Example: 
+If we have the following structure: A (parent) > B (children)
```
pyphilia, line 85:
```suggestion
If we have the following structure: A (parent) > B (child)
```

src/services/item/plugins/invitation/specifications.md
```diff
@@ -0,0 +1,101 @@
...
+- `admin` on A only (user has `admin` access to B with inheritance)
+- `read` on B only (user does not have access to A)
+- `write`on B only (user does not have access to A)
+- `admin` on B only (user does not have access to A)
+
+It is forbidden to have:
+- `admin` on A and `write` on B 
+- `admin` on A and `read` on B 
+- `write` on A and `read` on B 
+As these would result in B having a more restrictive permission than A.
```
pyphilia, line 101: It looks like more a membership description, should you move this to another file and have a reference in this one?

src/services/item/plugins/invitation/specifications.md
```diff
@@ -0,0 +1,101 @@
...
+
+A data entry targeting an already existing membership will result in a `Modifying Existing` error.
+
+#### Invitations
+
+An email that is already invited on an item will be droped. 
+
+Invitations can be set on parents or children of an item that already holds an invitation for the member but they need to follow membership inheritance rules:
+- A children membership can not be more restrictive (lower permission) than its parent memberships
+- A children membership can be more permissive (higher permission) than its parent membership
```
pyphilia, line 82:
```suggestion
- A children membership can not be more restrictive (lower permission) than its parent memberships, but can be equal or more permissive (higher permission) than its parent membership
```

src/services/item/plugins/invitation/specifications.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### Handling duplicates
+
+#### Memberships
+
+A data entry targeting an already existing membership will result in a `Modifying Existing` error.
+
+#### Invitations
+
+An email that is already invited on an item will be droped.
```
pyphilia, line 78: So if I understand correctly, you can set an invitation above or under, but never on the same item? can't we allow it as long as it does not break the permission rules? I can be postpone to later and saved in an issue.

src/services/item/plugins/invitation/repository.ts
```diff
@@ -84,16 +87,27 @@ export const InvitationRepository = AppDataSource.getRepository(Invitation).exte
...
+    const ids = insertResult.identifiers.map(({ id }) => id);
+    if (ids.length) {
+      // get the created invitations
+      const res = await this.createQueryBuilder('invitation')
+        .innerJoinAndSelect('invitation.item', 'item')
+        .innerJoinAndSelect('invitation.creator', 'creator')
+        .where('invitation.id IN (:...ids)', { ids })
+        .getMany();
+      return res;
+    }
```
pyphilia, line 109: Is it necessary?

src/services/item/plugins/invitation/service.ts
```diff
@@ -0,0 +1,350 @@
...
+  ): Promise<Invitation[]> {
+    if (!actor) {
+      throw new UnauthorizedMember(actor);
+    }
+    const { invitationRepository } = repositories;
+    const item = await this.itemService.get(actor, repositories, itemId, PermissionLevel.Admin);
+
+    const completeInvitations = await invitationRepository.postMany(invitations, item.path, actor);
+
+    // this.log.debug('send invitation mails');
```
pyphilia, line 104: uncomment?

src/services/item/plugins/invitation/service.ts
```diff
@@ -0,0 +1,350 @@
...
+    itemId: Item['id'],
+    file: MultipartFile,
+    itemMembershipService: ItemMembershipService,
+  ): Promise<{ memberships: ItemMembership[]; invitations: Invitation[] }> {
+    // verify file is CSV
+    verifyCSVFileFormat(file);
+
+    // get parentItem
+    const item = await repositories.itemRepository.get(itemId);
+    await validatePermission(repositories, PermissionLevel.Admin, actor, item);
```
pyphilia, line 240: If you get the item service, you can use `itemService.get()` that validates the permission for you. Also you can request for a specific permission level.

src/services/item/plugins/invitation/service.ts
```diff
@@ -0,0 +1,350 @@
...
+      memberships: ItemMembership[];
+      invitations: Invitation[];
+    }[]
+  > {
+    // verify file is CSV
+    verifyCSVFileFormat(file);
+
+    // get parentItem
+    const parentItem = await repositories.itemRepository.get(parentId);
+    await validatePermission(repositories, PermissionLevel.Admin, actor, parentItem);
```
pyphilia, line 288: same as above.

src/services/item/plugins/invitation/service.ts
```diff
@@ -0,0 +1,350 @@
...
+
+    // check that the template exists
+    const templateItem = await repositories.itemRepository.get(templateId);
+    if (!templateItem.id) {
+      throw new TemplateItemDoesNotExist();
+    }
```
pyphilia, line 294: use `itemService`? This will also valide the permission.

pyphilia: do you always need a template id ? the name of the function did not mention it.

src/services/item/plugins/invitation/service.ts
```diff
@@ -0,0 +1,350 @@
...
+
+    // group unique names from the group column,
+    // and then create the group folders
+    const dataByGroupName = groupby(rows, (r) => r.group_name);
+
+    const res = Array<{
+      groupName: string;
+      memberships: ItemMembership[];
+      invitations: Invitation[];
+    }>();
```
pyphilia, line 331:
```suggestion
    const res:{
      groupName: string;
      memberships: ItemMembership[];
      invitations: Invitation[];
    }[] = []
```
actually not sure which syntax is the best.

pyphilia: Would it make sense to use a map here instead?

pyphilia (changes requested):

src/services/item/plugins/invitation/index.ts
spaenleh: @pyphilia The `invitation` plugin is registered by the item API but depends on the item API and the `memberships` service
The `memberships` API is registered after the `item` API and depends on `item` and `mailer`. 

I think it might be possible if we instantiate the `memberships` as a plugin inside `item` before the invitations.
Or if we move invitations completely out of `item` and as its own service that depends on item and memberships. 

For now I think it is ok to de-structure it inside the request ? 

We have to map-out the services and their dependencies. Or use a tool like ts-syringe for dependency injection.

src/services/item/plugins/invitation/specifications.md
spaenleh: Yes you are right, you can set an invitation above or under an invitation that already exists, but not exactly on the same item for the same user. This is a mirror of how the memberships behave currently (not ideal). 
Because if you can set it on the same item, then you need to check if it is more, or less restrictive. 

I am aware that you can potentially introduce inconsistent invitations with this implementation, so this needs to be checked. Maybe it would cover the same item use-case.

src/services/item/plugins/invitation/service.ts
spaenleh: About the typing, I like the annotation with generics on Arrays, more, as it does feel less like a casting. I think both are equivalent. Let me know which one We should use throughout the codebase.

I tried with the map, but given that I already iterate over the groups, I found it hard (or too involved) to convert to a map. Also in the frontend I was planning to iterate over the elements anyways, so the map would not be very useful I think. Let me know If you think we should still use a map.

src/services/item/plugins/invitation/service.ts
spaenleh: Yes this function expects a template Id, as it is the one that creates the structure from the csv, i have renamed the function to make that clear.

spaenleh: And I will use the itemService instead.

src/services/item/plugins/invitation/service.ts
spaenleh: Okay will use that !

src/services/item/plugins/invitation/repository.ts
spaenleh: I replaced it, in order to not get the result of version. Do you think it should return the result of ?

src/services/item/plugins/invitation/specifications.md
spaenleh: See #929

spaenleh (approved):

src/services/item/plugins/invitation/index.ts
```diff
@@ -49,11 +62,70 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+
+      if (templateId) {
+        return await db.transaction(async (manager) =>
+          iS.createStructureForCSVAndTemplate(
+            member,
+            buildRepositories(manager),
+            itemId,
+            templateId,
+            uploadedFile,
+            itemMembershipService,
```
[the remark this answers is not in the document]
spaenleh: Ah yes sorry

Issue #716, opened 2023-12-07
Implement route to get shortlink from alias


ReidyT:
For the moment, it is only possible to be redirected to the item using the alias.

The mobile also need a route to retrieve the item id and the platform without being automatically redirected.

PR #717, opened 2023-12-07
feat: implement the get short link route

ReidyT:
see #716 for more information about this PR.

### DIFF
--- a/src/services/item/plugins/shortLink/index.ts
+++ b/src/services/item/plugins/shortLink/index.ts
@@ -6,7 +6,7 @@ import { ShortLinkAvailable, ShortLinkPatchPayload, ShortLinkPostPayload } from 
 
 import { UnauthorizedMember } from '../../../../utils/errors';
 import { buildRepositories } from '../../../../utils/repositories';
-import { create, update } from './schemas';
+import { create, restricted_get, update } from './schemas';
 import { SHORT_LINKS_LIST_ROUTE, ShortLinkService } from './service';
 
 const plugin: FastifyPluginAsync = async (fastify) => {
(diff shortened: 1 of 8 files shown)


src/services/item/plugins/shortLink/index.ts
```diff
@@ -24,6 +24,22 @@ const plugin: FastifyPluginAsync = async (fastify) => {
       reply.code(StatusCodes.MOVED_TEMPORARILY).redirect(path);
     });
 
+    // WARNING: Do not return the entire item, because this route is no protected !
```
pyphilia, line 27:
```suggestion
    // WARNING: Do not return the entire item, because this route is not protected !
```

src/services/item/plugins/shortLink/test/index.test.ts
```diff
@@ -95,6 +96,21 @@ describe('Short links routes tests', () => {
       });
     });
 
+    describe('GET /short-links/alias/:alias without connected member', () => {
```
pyphilia, line 99: Add classic test when you are connected?

src/services/item/plugins/shortLink/repository.ts
```diff
@@ -54,11 +54,20 @@ export const ShortLinkRepository = AppDataSource.getRepository(ShortLink).extend
...
+  async get(alias: string, returnItemIdOnly: boolean = false): Promise<ShortLink> {
     if (!alias) throw new ShortLinkNotFound(alias);
 
+    const selectItemIdOnly = {
+      select: {
+        item: {
+          id: true,
+        },
+      },
+    };
```
pyphilia, line 66: I wonder if it would not be better to go as described here? https://github.com/typeorm/typeorm/issues/379 Otherwise you do a join without using it at all. Let me know if it changes to much stuff in your previous code.

pyphilia (changes requested):
Thanks for this quick PR! The test is really nice. I have one concern let me know if you have questions 🔆

pyphilia (approved):

PR #725, opened 2023-12-13
feat: allow to query app settings by name

spaenleh:
In this PR I add the possibility to query app settings by name. This is done with an optional query parameter. This will allow consumers of the API to get only the list of settings that have a specific name. This will let them type the array as having a specific type linked to that name. This should improve the DX when writing apps.

### DIFF
--- a/src/services/item/plugins/app/appData/test/index.test.ts
+++ b/src/services/item/plugins/app/appData/test/index.test.ts
@@ -207,6 +207,22 @@ describe('App Data Tests', () => {
         expect(receivedAppData.length).toEqual(appDataOfType.length);
       });
 
+      it('Get empty data for type that does not exist', async () => {
+        const response = await app.inject({
+          method: HttpMethod.GET,
+          url: `${APP_ITEMS_PREFIX}/${item.id}/app-data?type=impossible-type`,
+          headers: {
+            Authorization: `Bearer ${token}`,
+          },
+        });
+        const appDataOfType = appData.filter((d) => d.type === 'impossible-type');
+        const receivedAppData = await response.json();
+        expect(appDataOfType.length).toEqual(0);
+        expect(response.statusCode).toEqual(StatusCodes.OK);
+        expectAppData(appDataOfType, receivedAppData);
+        expect(receivedAppData.length).toEqual(appDataOfType.length);
+      });
+
       it('Get app data with invalid item id throws', async () => {
         const response = await app.inject({
           method: HttpMethod.GET,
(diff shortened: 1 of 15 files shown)


src/services/item/plugins/app/appData/test/index.test.ts
```diff
@@ -189,6 +191,22 @@ describe('App Data Tests', () => {
...
+          },
+        });
+        const appDataOfType = appData.filter((d) => d.type === 'other-type');
+        const receivedAppData = await response.json();
+        expect(appDataOfType.length).toBeGreaterThanOrEqual(1);
+        expect(response.statusCode).toEqual(StatusCodes.OK);
+        expectAppData(appDataOfType, receivedAppData);
+        expect(receivedAppData.length).toEqual(appDataOfType.length);
+      });
+
```
pyphilia, line 209: Add a test to check what happens if you don't have matching app data?

src/services/item/plugins/app/appSetting/repository.ts
```diff
@@ -60,13 +60,14 @@ export const AppSettingRepository = AppDataSource.getRepository(AppSetting).exte
...
 
-  getForItem(itemId: string): Promise<AppSetting[]> {
+  getForItem(itemId: string, name?: string): Promise<AppSetting[]> {
     if (!itemId) {
       throw new ItemNotFound(itemId);
     }
     return this.find({
       where: {
         item: { id: itemId },
+        ...(name ? { name } : undefined),
```
pyphilia, line 70: I never tried, but does `...undefined` works?

src/services/item/plugins/app/appSetting/test/index.test.ts
```diff
@@ -125,6 +137,20 @@ describe('Apps Settings Tests', () => {
...
+          method: HttpMethod.GET,
+          url: `${APP_ITEMS_PREFIX}/${item.id}/app-settings?name=new-setting`,
+          headers: {
+            Authorization: `Bearer ${token}`,
+          },
+        });
+        expect(response.statusCode).toEqual(StatusCodes.OK);
+        expectAppSettings(response.json(), [appSettings.find((s) => s.name === 'new-setting')]);
+      });
+
```
pyphilia, line 153: Add a test with empty result?

pyphilia (approved):
LGTM, I added some suggestions for the tests, but it looks good! 🎁

src/services/item/plugins/app/appSetting/repository.ts
spaenleh: Looks like it does !

PR #738, opened 2024-01-04
feat: add optional name to copy API

LinaYahya:


src/services/item/controller.ts
```diff
@@ -27,6 +27,9 @@ import { Ordered } from './interfaces/requests';
 import { ItemSearchParams } from './types';
 import { ItemOpFeedbackEvent, memberItemsTopic } from './ws/events';
 
+export interface ItemName {
+  [key: string]: string;
+}
```
spaenleh, line 32: I generally prefer type aliases for custom defined types.
Maybe we could change the type naming to express better that is is related to an association between the item Id and the a new name. Below is a suggestion:
```suggestion
export type ItemIdToNameMap = {
  [key: string]: string;
}
```
The type name is not final, please feel free to choose a better one 😄

src/services/item/controller.ts
```diff
@@ -297,21 +300,22 @@ const plugin: FastifyPluginAsync = async (fastify) => {
 
   fastify.post<{
     Querystring: IdsParams;
-    Body: { parentId: string };
+    Body: { parentId: string; itemsName?: ItemName };
```
spaenleh, line 303:
```suggestion
    Body: { parentId: string; newNames?: ItemName };
```

src/services/item/fluent-schema.ts
```diff
@@ -269,7 +269,7 @@ export const copyMany = {
   querystring: S.object()
     .prop('id', S.array().maxItems(MAX_TARGETS_FOR_MODIFY_REQUEST))
     .extend(idsQuery),
-  body: S.object().additionalProperties(false).prop('parentId', uuid),
+  body: S.object().additionalProperties(false).prop('parentId', uuid).prop('itemsName', S.object()),
```
spaenleh, line 272: Can we make the schema a bit more precise ?

src/services/item/service.ts
```diff
@@ -430,7 +431,13 @@ export class ItemService {
   }
 
   /////// -------- COPY
-  async copy(actor: Actor, repositories: Repositories, itemId: UUID, args: { parentId?: UUID }) {
+  async copy(
+    actor: Actor,
+    repositories: Repositories,
+    itemId: UUID,
+    args: { parentId?: UUID },
+    name?: string,
```
spaenleh, line 439: If possible I would keep the newName in the `args` argument kind of like a optional argument instead of adding a new parameter to the function.
```suggestion
    args: { parentId?: UUID; newName?: string },
```

src/services/item/service.ts
```diff
@@ -454,6 +461,7 @@ export class ItemService {
     }
 
     const descendants = await itemRepository.getDescendants(item);
+    item.name = name || item.name;
```
spaenleh, line 464: We could make the renaming a bit more explicit by using an if statement:
```suggestion
    if (args.name) {
      // assign a new name to the copied item
      // this supports the duplicate item feature
      item.name = args.name
    }
```

spaenleh (changes requested):
Thank you for this change ! 🌳 
I have left some comments mainly nitpicks on naming and style opinions. Let me know if you want to discuss some comments.

spaenleh (approved):
Awesome !

LinaYahya:
copyItems by itself will append `-copy` to all items

PR #740, opened 2024-01-08
fix: add logger to fileService

### LINKED ISSUE #699: Add logger in FileService to replace console.error
In the FileService, there are some console.error calls when we get an error.

This leads to hard to navigate logs in AWS because it does not print to a JSON object, but line by line.
We should use the fastify logger if possible in this module.

spaenleh:
This PR adds a logger to the fileService to remove the ugly `console.error` printouts in aws logs.
fix #699


pyphilia (approved):
💯

PR #746, opened 2024-01-08
fix: add transaction in import post

### LINKED ISSUE #743: Add transactions in import zip
Without transactions, the logic does not work (getting an item that just have been created).
With one transaction at the very top it won't save partial results.
So the optimal solution is to add transactions at key steps of the import process.

pyphilia:
close #743


spaenleh (approved):
Woop Woop 🚀  firework

PR #752, opened 2024-01-15
feat: include app data, setting and actions in analytics

### LINKED ISSUE #668: Export app actions in analytics export
This would be nice to have all actions in the exported zip link.

pyphilia:
close #668

### DIFF
--- a/src/services/action/utils/export.test.ts
+++ b/src/services/action/utils/export.test.ts
@@ -94,14 +94,14 @@ describe('exportActionsInArchive', () => {
 
     // call on success callback
     expect(result).toBeTruthy();
-    // create files for all views, items, members and memberships, chat messages
-    expect(writeFileSyncMock).toHaveBeenCalledTimes(views.length + 5);
+    // create files for all views, items, members and memberships, chat messages, apps
+    expect(writeFileSyncMock).toHaveBeenCalledTimes(views.length + 6);
     const files = fs.readdirSync(storageFolder);
     expect(files.length).toBeTruthy();
 
     // assume only 2 files exist in the folder
     const [folder, zip] = files;
     expect(zip.includes(baseAnalytics.item.name)).toBeTruthy();
-    expect(fs.readdirSync(path.join(storageFolder, folder)).length).toEqual(views.length + 5);
+    expect(fs.readdirSync(path.join(storageFolder, folder)).length).toEqual(views.length + 6);
   });
 });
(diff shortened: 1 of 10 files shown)


src/services/action/utils/export.test.ts
```diff
@@ -93,14 +94,14 @@ describe('exportActionsInArchive', () => {
 
     // call on success callback
     expect(result).toBeTruthy();
-    // create files for all views, items, members and memberships, chat messages
-    expect(writeFileSyncMock).toHaveBeenCalledTimes(views.length + 5);
+    // create files for all views, items, members and memberships, chat messages, apps
+    expect(writeFileSyncMock).toHaveBeenCalledTimes(views.length + 6);
```
spaenleh, line 98: This offset seems a bit magic, I am not sure how we could make it more readable and adaptable to the logic behind it ?

src/services/item/plugins/action/base-analytics.test.ts
```diff
@@ -27,7 +29,7 @@ describe('Base Analytics', () => {
     app.close();
   });
 
-  it('Members should be cleaned', async () => {
+  it.only('Members should be cleaned', async () => {
```
spaenleh, line 32: Remove only ?

src/services/item/plugins/action/base-analytics.test.ts
```diff
@@ -66,5 +79,21 @@ describe('Base Analytics', () => {
...
+    for (const aa of appActions) {
+      expect(aa.member?.createdAt).toBeUndefined();
+    }
+    for (const ad of appData) {
+      expect(ad.member?.createdAt).toBeUndefined();
+      expect(ad.creator?.createdAt).toBeUndefined();
+    }
+    for (const as of appSettings) {
+      expect(as.creator?.createdAt).toBeUndefined();
+    }
```
spaenleh, line 97: Not sure what this is supposed to test ? The non-existence of some member properties ?

spaenleh (changes requested):
Looks good, just some comments about comprehension and a remaining `only`

src/services/item/plugins/action/base-analytics.test.ts
pyphilia: yes, it was a preventive validation to avoid leaking password. Now it test we don't leak createdAt (that is, we only returns the minimal member). I think I can make it more exhaustive.

src/services/item/plugins/action/base-analytics.test.ts
pyphilia: idk 🤷‍♀️

src/services/action/utils/export.test.ts
pyphilia: idk 🤷‍♀️

spaenleh (approved):
🗳️

PR #755, opened 2024-01-15
fix: disable bad words validation

spaenleh:


src/services/item/plugins/validation/processes/badWordsDetection.ts
```diff
@@ -9,19 +9,20 @@ export const buildWordList = (badWordsFilter: BadWordsFilter): void => {
...
+  // const contents = documents?.filter(Boolean);
+  // const badWordsFilter = new BadWordsFilter();
+  // buildWordList(badWordsFilter);
+  // const suspiciousFields: string[] = [];
+  // for (const { value, name } of contents) {
+  //   if (badWordsFilter.isProfane(value)) {
+  //     suspiciousFields.push(name);
+  //   }
+  // }
+  // return suspiciousFields;
```
swouf, line 27: As precaution, you could leave a comment to remind us in the future why there is this commented code here? Maybe with a link to the related issue.

src/services/item/plugins/recycled/test/ws.test.ts
```diff
@@ -164,9 +164,14 @@ describe('Recycle websocket hooks', () => {
...
-        expect(selfDelete).toMatchObject(SelfItemEvent('delete', updatedParent));
-        expect(childDelete).toMatchObject(ChildItemEvent('delete', updatedChild));
+        // todo: this part of the test seemed flaky as the order of the updates may not be guarantied
+        // const [selfDelete, childDelete] = itemUpdates;
+        expect(itemUpdates.find((v) => v.kind === 'self')).toMatchObject(
+          SelfItemEvent('delete', updatedParent),
+        );
+        expect(itemUpdates.find((v) => v.kind === 'child')).toMatchObject(
+          ChildItemEvent('delete', updatedChild),
+        );
```
swouf, line 174: Why do you need to change that? Seems legit but I don't get it.

swouf (changes requested):
Allright 👍🏼 

I just suggest that you leave a comment to explain what happened and what was removed. From a user point of view, it looks like the filter is still on and one must look at the code to see that it was disabled. We should make sure it's easy to understand that there is no filter.

src/services/item/plugins/recycled/test/ws.test.ts
spaenleh: So this is not directly related to the issue with the badWords.
This test has been flaky in the CI for some time. It looks like there is supposed to be two updates that are received but they may not arrive in the expected order. So the test fails because it gets the second one before the first one. This is just to fix that flakyness.

src/services/item/plugins/validation/processes/badWordsDetection.ts
spaenleh: Yes I will add more detail to explain why this is commented out and possible solutions.

src/services/item/plugins/recycled/test/ws.test.ts
pyphilia, line 174: Maybe .toContain (or similar) could have been better.

pyphilia (approved):

PR #757, opened 2024-01-16
fix: do not fail login with password if score is low

### LINKED ISSUE #394: Define a lower captcha threshold for endpoints that don't send emails?
Sign in with password don't really need a high security -> auto fill often fails

spaenleh:
In this PR:
- fix #394 by adding a parameter in `verifyCaptcha` which decides if a low score should fail the request or not.


pyphilia (approved):
tests? 👀

PR #759, opened 2024-01-16
refactor: use class for favorite item

pyphilia:


spaenleh (commented):
Changes look good, but it looks like the tests did not like it. 🤷

pyphilia:
I've added test at repository level, they will be a bit duplicata from the controller one, but maybe we can further remove some in the controller.

spaenleh (commented):
I see that you use the `build()` method in the tests, so you effectively instantiate the whole server to test our custom repository. Should we try to do "just" unit tests on the custom repository ? We would stub everything else and just ensure that the logic of our custom functions are correct ? Would that be possible ? Do you think it could be lighter/faster ?

spaenleh (approved):
Looks good, as this way of testing allows to test the repository end-2-end with the sql and DB

PR #763, opened 2024-01-17
fix: add dummy bad words validation

spaenleh:


src/services/item/plugins/recycled/test/ws.test.ts
```diff
@@ -315,7 +315,7 @@ describe('Recycle websocket hooks', () => {
       });
     });
 
-    it('admins receive recycle bin create update when item is recycled', async () => {
+    it.only('admins receive recycle bin create update when item is recycled', async () => {
```
pyphilia, line 318: Remove only!

pyphilia (changes requested):

pyphilia (approved):

PR #771, opened 2024-01-19
fix: normalise invitation email

### LINKED ISSUE #769: [bug] Invitation emails are case-sensitive
### Problem statement
When a user registers without using the email link sent by the invitation, it is **possible** for them to use a differently cased email. This means that the invitation they receive may not be matched because it is matched 1:1 (case-sensitive).

### Steps to reproduce
- Invite a user, by using mixed case in the email (i.e. `Bob-iNvitaTionS@example.com`)
  - (in the backend the email is not transformed to lowercase (normalised))
- User receives email about invitation but chooses to register manually (using a different case, for example `bob-invitationS@example.com`)
  -  (in the backend the email is lowercased and a new member is created with the normalised version of the email)
  - the backend checks if there are invitations for this (normalised) email -> there are no invitations because the invitation email has not been normalised
- User does not get the items he was invited to.

See this comment:
https://github.com/graasp/graasp/blob/b4ac8c76278d2209e35d21a55a0a80f90000f88f/src/services/member/service.ts#L33-L36

### Culprit
But when we receive an invitation we do not save the normalised email in the database. We save it as-is.

### Solution
Normalise emails we receive before saving them in the DB

spaenleh:
In this PR:
- fix #769

### DIFF
--- a/src/services/invitation/service.ts
+++ b/src/services/invitation/service.ts
@@ -130,7 +130,7 @@ export class InvitationService {
     // invitations to memberships is triggered on register: no actor available
     const { invitationRepository, itemMembershipRepository } = repositories;
     const invitations = await invitationRepository.find({
-      where: { email: member.email },
+      where: { email: member.email.toLowerCase() },
       relations: { item: true },
     });
     const memberships = invitations.map(({ permission, item }) => ({ item, member, permission }));
--- a/src/services/invitation/repository.ts
+++ b/src/services/invitation/repository.ts
@@ -66,10 +66,15 @@ export const InvitationRepository = AppDataSource.getRepository(Invitation).exte
    * @param invitation Invitation to create
    */
   async postMany(
-    invitations: Partial<Invitation>[],
+    partialInvitations: Partial<Invitation>[],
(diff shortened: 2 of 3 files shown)


src/services/invitation/service.ts
```diff
@@ -130,7 +130,7 @@ export class InvitationService {
     // invitations to memberships is triggered on register: no actor available
     const { invitationRepository, itemMembershipRepository } = repositories;
     const invitations = await invitationRepository.find({
-      where: { email: member.email },
+      where: { email: member.email.toLowerCase() },
```
pyphilia, line 133: should the email have been lowercase from the beginning (on registration) as well?

pyphilia (approved):
Thanks for the fix!

src/services/invitation/service.ts
spaenleh: It is being lowercased on registration [see in the `post` method of the `MemberService`](https://github.com/graasp/graasp/blob/391e0479aabf92a64f2363fe1b86c6ffacb311f3/src/services/member/service.ts#L36)

PR #777, opened 2024-01-22
feat: add geolocation endpoints

### LINKED ISSUE #775: Add geolocation for items

pyphilia:
In the end I've added already lots of endpoints, while thinking of the implementation. 

close #775

### DIFF
--- a/src/services/item/controller.ts
+++ b/src/services/item/controller.ts
@@ -25,7 +25,6 @@ import {
 } from './fluent-schema';
 import { Ordered } from './interfaces/requests';
 import { ItemGeolocation } from './plugins/geolocation/ItemGeolocation';
-import { PartialItemGeolocation } from './plugins/geolocation/errors';
 import { ItemSearchParams } from './types';
 import { ItemOpFeedbackEvent, memberItemsTopic } from './ws/events';
 
@@ -52,19 +51,12 @@ const plugin: FastifyPluginAsync = async (fastify) => {
         body: data,
       } = request;
 
-      // lat and lng should exist together
-      const { geolocation } = data;
-      const { lat, lng } = geolocation || {};
-      if ((lat && !lng) || (lng && !lat)) {
-        throw new PartialItemGeolocation({ lat, lng });
-      }
-
       return await db.transaction(async (manager) => {
         const repositories = buildRepositories(manager);
         const item = await itemService.post(member, repositories, {
           item: data,
           parentId,
-          ...(geolocation ? { geolocation } : {}),
+          geolocation: data.geolocation,
         });
         await actionItemService.postPostAction(request, reply, repositories, item);
         return item;
--- a/src/services/item/controller.ts
+++ b/src/services/item/controller.ts
@@ -24,6 +24,8 @@ import {
   updateMany,
 } from './fluent-schema';
 import { Ordered } from './interfaces/requests';
+import { ItemGeolocation } from './plugins/geolocation/ItemGeolocation';
+import { PartialItemGeolocation } from './plugins/geolocation/errors';
 import { ItemSearchParams } from './types';
 import { ItemOpFeedbackEvent, memberItemsTopic } from './ws/events';
 
@@ -34,7 +36,10 @@ const plugin: FastifyPluginAsync = async (fastify) => {
 
   // create item
   // question: add link hook here? or have another endpoint?
-  fastify.post<{ Querystring: ParentIdParam; Body: Partial<Item> }>(
+  fastify.post<{
+    Querystring: ParentIdParam;
+    Body: Partial<Item> & Pick<ItemGeolocation, 'lat' | 'lng'>;
+  }>(
     '/',
     {
       schema: items.extendCreateSchema(),
@@ -46,11 +51,19 @@ const plugin: FastifyPluginAsync = async (fastify) => {
         query: { parentId },
         body: data,
       } = request;
+
+      // lat and lng should exist together
+      const { lat, lng } = data;
+      if ((lat && !lng) || (lng && !lat)) {
+        throw new PartialItemGeolocation({ lat, lng });
+      }
+
       return await db.transaction(async (manager) => {
         const repositories = buildRepositories(manager);
         const item = await itemService.post(member, repositories, {
           item: data,
           parentId,
+          ...(data.lng && data.lat ? { geolocation: { lng: data.lng, lat: data.lat } } : {}),
         });
         await actionItemService.postPostAction(request, reply, repositories, item);
         return item;
--- a/src/services/item/controller.ts
+++ b/src/services/item/controller.ts
@@ -53,7 +53,8 @@ const plugin: FastifyPluginAsync = async (fastify) => {
       } = request;
 
       // lat and lng should exist together
-      const { lat, lng } = data;
+      const { geolocation } = data;
+      const { lat, lng } = geolocation || {};
       if ((lat && !lng) || (lng && !lat)) {
         throw new PartialItemGeolocation({ lat, lng });
       }
@@ -63,7 +64,7 @@ const plugin: FastifyPluginAsync = async (fastify) => {
         const item = await itemService.post(member, repositories, {
           item: data,
           parentId,
-          ...(data.lng && data.lat ? { geolocation: { lng: data.lng, lat: data.lat } } : {}),
+          ...(geolocation ? { geolocation } : {}),
         });
         await actionItemService.postPostAction(request, reply, repositories, item);
         return item;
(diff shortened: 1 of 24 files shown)


src/services/item/plugins/geolocation/repository.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  constructor(manager?: EntityManager) {
+    if (manager) {
+      this.repository = manager.getRepository(ItemGeolocation);
+    } else {
+      this.repository = AppDataSource.getRepository(ItemGeolocation);
+    }
+  }
+
+  async getForItem(item: Item): Promise<ItemGeolocation | null> {
```
ReidyT, line 19: I'm not sure that `getForItem` is the best name to describe the function, but maybe it is only me. For me, something like `getLocationByItem` or just `getByItem` is clearer.

src/services/item/plugins/geolocation/schemas.ts
```diff
@@ -0,0 +1,72 @@
...
+    createdAt: { type: 'string' },
+    updatedAt: { type: 'string' },
+    item: {
+      $ref: 'http://graasp.org/items/#/definitions/item',
+    },
+  },
+  required: ['lat', 'lng'],
+};
+
+export const getForItem = {
```
ReidyT, line 19: Same here,

src/services/item/plugins/geolocation/service.ts
```diff
@@ -0,0 +1,94 @@
...
+import { ItemGeolocationNotFound } from './errors';
+
+export class ItemGeolocationService {
+  private itemService: ItemService;
+
+  constructor(itemService: ItemService) {
+    this.itemService = itemService;
+  }
+
+  async getForItem(actor: Actor, repositories: Repositories, itemId: Item['id']) {
```
ReidyT, line 17: And here

src/services/item/plugins/geolocation/specifications.md
ReidyT, line 1: Nice :)

src/services/item/plugins/geolocation/specifications.md
```diff
@@ -0,0 +1,16 @@
...
+
+Geolocation attached to a hierarchy of items.
+Geolocation can be inherited (a folder gives its geolocation to its children).
+
+## Endpoints
+
+- `GET /items/:id/geolocation`: return geolocation for item, inherits from parent
+- `GET /items/geolocation?lat1<lat1>&lat2<lat2>&lng1<lng1>&lng2<lng2>`: return geolocations and their items within bounding box
+- `PUT /items/:id/geolocation`: save geolocation for item
+- `POST /items/map`: create an item including its geolocation
```
ReidyT, line 11: I'm thinking why could we not just use `/items` path to create a new item ? If the geolocation is given, it will create it. Otherwise, it will only create the item.

src/services/item/controller.ts
```diff
@@ -46,11 +51,19 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+      if ((lat && !lng) || (lng && !lat)) {
+        throw new PartialItemGeolocation({ lat, lng });
+      }
+
       return await db.transaction(async (manager) => {
         const repositories = buildRepositories(manager);
         const item = await itemService.post(member, repositories, {
           item: data,
           parentId,
+          ...(data.lng && data.lat ? { geolocation: { lng: data.lng, lat: data.lat } } : {}),
```
pyphilia, line 66: I don't need the check anymore

src/services/item/plugins/geolocation/ItemGeolocation.ts
```diff
@@ -0,0 +1,45 @@
...
+  @JoinColumn({ referencedColumnName: 'path', name: 'item_path' })
+  item: Item;
+
+  @Column({ type: 'float' })
+  lat: number;
+
+  @Column({ type: 'float' })
+  lng: number;
+
+  // easier for display in frontend
```
pyphilia, line 36: remove comment

src/services/item/plugins/geolocation/errors.ts
```diff
@@ -0,0 +1,37 @@
+import { StatusCodes } from 'http-status-codes';
+
+import { ErrorFactory } from '@graasp/sdk';
+
+const PLUGIN_NAME = 'graasp-plugin-item-geolocation';
+
+/**
+ * Errors thrown by the chat tasks
+ */
```
pyphilia, line 9: Change comment

src/services/item/plugins/geolocation/index.test.ts
```diff
@@ -0,0 +1,323 @@
...
+        item: expect.objectContaining({
+          id: ig.item.id,
+          creator: expect.objectContaining({ id: ig.item.creator.id }),
+        }),
+      }),
+    );
+  }
+};
+
+describe('Favorite', () => {
```
pyphilia, line 36: Change title

src/services/item/service.ts
```diff
@@ -56,16 +56,16 @@ export class ItemService {
   async post(
     actor: Actor,
     repositories: Repositories,
-    args: { item: Partial<Item>; parentId?: string },
+    args: { item: Partial<Item>; parentId?: string; geolocation?: { lat: number; lng: number } },
```
pyphilia, line 59: Use Pick and ItemGeolocation.

pyphilia:
So in this PR I've included basic endpoints such as:
- post item with geoloc (through POST /items)
- post / update geoloc 
- delete an item geoloc
- get geoloc for item
- get geoloc (and items) within a bounding box

Later I'll need to add more filtering especially for the last endpoint.

src/services/item/controller.ts
ReidyT: You still have to check that lat and long are defined, don't you ?

src/services/item/plugins/geolocation/index.test.ts
```diff
@@ -0,0 +1,323 @@
...
+      it('Throws if no geolocation', async () => {
+        const res = await app.inject({
+          method: HttpMethod.GET,
+          url: `${ITEMS_ROUTE_PREFIX}/${item.id}/geolocation`,
+        });
+        expect(res.statusCode).toBe(StatusCodes.NOT_FOUND);
+        expect(res.json()).toMatchObject(new ItemGeolocationNotFound(expect.anything()));
+      });
+
+      it('Throws if id is incorrect', async () => {
```
ReidyT, line 125: Maybe `Throws if id doesn't have good format` or something like that is a little bit more explicit ? It's maybe only me, but I thought it was to test an id that not exist 😅.

src/services/item/plugins/geolocation/repository.test.ts
```diff
@@ -0,0 +1,231 @@
...
+
+      // noise
+      const geolocParent = { lat: 1, lng: 2, item: parentItem, country: 'de' };
+      await rawRepository.save(geolocParent);
+
+      const res = await repository.getByItem(item);
+      expect(res).toMatchObject({
+        lat: geoloc.lat,
+        lng: geoloc.lng,
+        country: geolocParent.country,
```
ReidyT, line 116: Shouldn't it be `geoloc.country` instead ?

src/services/item/plugins/geolocation/repository.test.ts
```diff
@@ -0,0 +1,231 @@
...
+      expect(geoloc).toMatchObject({ lat, lng, country: 'AQ' });
+    });
+    it('create new geolocation that does not have a country', async () => {
+      const { item } = await saveItemAndMembership({ member: actor });
+
+      const lat = 1;
+      const lng = 2;
+      await repository.put(item.path, lat, lng);
+      const geoloc = await rawRepository.findOneBy({ lat, lng });
+      expect(geoloc).toMatchObject({ lat, lng, country: null });
```
ReidyT, line 215: I'm not sure to understand the difference between this test, and the previous one. How it is possible that this test return a null country ?

Is there any reason to not enforce to always receive a country when saving a lat and long ? I mean, saving the country in the database when inserting or updating lat and long ?

src/services/item/plugins/geolocation/ItemGeolocation.ts
```diff
@@ -0,0 +1,45 @@
...
+  @Column({ type: 'character varying', nullable: true, length: 4 })
```
ReidyT, line 37: If I have understand correctly, the GeoRepository raise an exception if lat, long are invalid (if the country is not found). Is there a reason to set country to nullable ? It is possible using the backend to insert a null country ?

ReidyT (approved):
Great work 🙌, I just left few comments 😄.

src/services/item/plugins/geolocation/repository.test.ts
pyphilia: I'm accepting points that are in the ocean I guess? but we could enforce points on earth?

ReidyT: Yeah I didn't thought about that. I think if it is acceptable to have points in the ocean or something like that, it is not necessary to enforce points on earth ?

PR #786, opened 2024-01-25
fix: throw if item is not folder for copy move

pyphilia:
check for copy and move whether

### DIFF
--- a/src/services/item/repository.ts
+++ b/src/services/item/repository.ts
@@ -76,6 +76,11 @@ export const ItemRepository = AppDataSource.getRepository(Item).extend({
       settings = DEFAULT_ITEM_SETTINGS,
       creator,
     } = args;
+
+    if (parent && !isItemType(parent, ItemType.FOLDER)) {
+      throw new ItemNotFolder(parent);
+    }
+
     // TODO: extra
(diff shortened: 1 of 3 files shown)


src/services/item/repository.ts
```diff
@@ -357,6 +368,11 @@ export const ItemRepository = AppDataSource.getRepository(Item).extend({
     creator: Member,
     parentItem?: Item,
   ): Promise<{ copyRoot: Item; treeCopyMap: Map<string, { original: Item; copy: Item }> }> {
+    // cannot move inside non folder item
```
pyphilia, line 371: fix comment

ReidyT (approved):
LGTM 👍. I'm just thinking if it could be useful to encapsulate the logic into a util function, because it is repeated multiple times ? Maybe not 😅.

pyphilia:
> LGTM 👍. I'm just thinking if it could be useful to encapsulate the logic into a util function, because it is repeated multiple times ? Maybe not 😅.

Thanks for reviewing! I see what you mean, I think it's a bit too short (if you're talking about the if(not folder)) to factor out, but it's open for discussion^^

PR #793, opened 2024-01-29
feat: allow to filter out item by type in accessible

ReidyT:
Add the query parameter `types` to allow to filter by item types in GET /items/accessible and /items/:id/children.

### DIFF
--- a/src/services/item/test/index.test.ts
+++ b/src/services/item/test/index.test.ts
@@ -37,7 +37,7 @@ import { Item } from '../entities/Item';
 import { ItemGeolocation } from '../plugins/geolocation/ItemGeolocation';
 import { ItemTagRepository } from '../plugins/itemTag/repository';
 import { ItemRepository } from '../repository';
-import { SortBy } from '../types';
+import { Ordering, SortBy } from '../types';
 import {
   expectItem,
   expectManyItems,
@@ -1012,7 +1012,7 @@ describe('Item routes tests', () => {
           method: HttpMethod.GET,
           url: `/items/accessible`,
           query: {
-            sorty: SortBy.ItemCreatorName,
+            sortBy: SortBy.ItemCreatorName,
             ordering: 'asc',
             permissions: [PermissionLevel.Read],
           },
@@ -1051,7 +1051,7 @@ describe('Item routes tests', () => {
           method: HttpMethod.GET,
           url: `/items/accessible`,
           query: {
-            sorty: SortBy.ItemCreatorName,
+            sortBy: SortBy.ItemCreatorName,
             ordering: 'asc',
             permissions: [PermissionLevel.Write, PermissionLevel.Admin],
           },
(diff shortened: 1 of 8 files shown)


src/services/item/types.ts
```diff
@@ -23,4 +23,5 @@ export type ItemSearchParams = {
   sortBy?: SortBy;
   ordering?: Ordering;
   permissions?: PermissionLevel[];
+  itemType?: (typeof ItemType)[keyof typeof ItemType];
```
spaenleh, line 26: use `UnionOfConst<typeof ItemType` ?

spaenleh (commented):
Nice ! If I understand correctly this allows to pass a single itemType to filter on ? Should we already add the possibility to filter with multiple itemTypes ? Lets say, `folder & links` ?

src/services/item/types.ts
ReidyT: Yes I  updated the code, now I use unionOfConst and we can use an array.

src/services/item/test/index.test.ts
```diff
@@ -1025,6 +1025,55 @@ describe('Item routes tests', () => {
...
+          },
+        });
+
+        expect(response.statusCode).toBe(StatusCodes.OK);
+
+        const { data, totalCount } = response.json();
+        expect(totalCount).toEqual(folders.length);
+        expect(data).toHaveLength(folders.length);
+        folders.forEach((folder, idx) => {
+          expectItem(data[idx], folder);
```
spaenleh, line 1072: Does this rely on the creation order ? So it will break if we somehow change how we order elements ?

src/services/item/test/index.test.ts
```diff
@@ -1253,6 +1324,42 @@ describe('Item routes tests', () => {
...
+        const { item: child2 } = await saveItemAndMembership({
+          item: getDummyItem({ name: 'child2', type: ItemType.FOLDER }),
+          member,
+          parentItem: parent,
+        });
+        const children = [child2];
+
+        const response = await app.inject({
+          method: HttpMethod.GET,
+          url: `/items/${parent.id}/children?types=folder`,
```
spaenleh, line 1348: Use que same syntax as for the `Accessible` ? with the Query ? Why write the query manually ?

spaenleh (approved):
Looks good, just a question and an inconsistency in how you create the request url in a test.

src/services/item/test/index.test.ts
```diff
@@ -1025,6 +1025,55 @@ describe('Item routes tests', () => {
...
+          item: { type: ItemType.APP },
+        });
+
+        const folders = [itemFolder1, itemFolder2];
+
+        const response = await app.inject({
+          method: HttpMethod.GET,
+          url: `/items/accessible`,
+          query: {
+            sorty: SortBy.ItemCreatorName,
```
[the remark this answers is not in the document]
ReidyT: Oh this is a bad copy and paste, I will update all test using sorty thanks !

src/services/item/test/index.test.ts
ReidyT: Yes it fails if the order is not respected... Maybe you're right, I will remove this control in this test case because it is not what it should test...

PR #798, opened 2024-01-31
feat: add search for geolocation endpoint

### LINKED ISSUE #796: full text search for geolocation endpoint

pyphilia:
Some explanations:

This PR uses PostgresQL's native full text search, based on vectors from a string (mainly listing all words and setting weight on them). Also, if the parsing language is correctly set it can remove stop words and stem the words. For example if we have "I have black cats" it will be parsed as vectors:

- 1. for english: [ have, black, cat ]
- 2. for french: [ I, have, black, cats ]

With these, if we search ('english', 'cats) it will match vectors 1 and 2.
But if we search ('french', 'cats') it will match only vector 2 (because it cannot infer that cats = cat).

The main problem we have is thus languages, since we allow anyone to create content in any language. 

close #796


src/migrations/1706698413211-full-text-search.ts
```diff
@@ -0,0 +1,64 @@
...
+      setweight(to_tsvector('french', COALESCE(replace(extra, '\\u0000', '')::jsonb->'file'->'content','{}')), 'D') || ' ' ||
+      setweight(to_tsvector('italian', COALESCE(replace(extra, '\\u0000', '')::jsonb->'file'->'content','{}')), 'D') || ' ' ||
+      setweight(to_tsvector('german', COALESCE(replace(extra, '\\u0000', '')::jsonb->'file'->'content','{}')), 'D') || ' ' ||
+      setweight(to_tsvector('spanish', COALESCE(replace(extra, '\\u0000', '')::jsonb->'file'->'content','{}')), 'D') || ' ' ||
+      setweight(to_tsvector('english', COALESCE(replace(extra, '\\u0000', '')::jsonb->'s3File'->'content','{}')), 'D') || ' ' ||
+      setweight(to_tsvector('french', COALESCE(replace(extra, '\\u0000', '')::jsonb->'s3File'->'content','{}')), 'D') || ' ' ||
+      setweight(to_tsvector('italian', COALESCE(replace(extra, '\\u0000', '')::jsonb->'s3File'->'content','{}')), 'D') || ' ' ||
+      setweight(to_tsvector('german', COALESCE(replace(extra, '\\u0000', '')::jsonb->'s3File'->'content','{}')), 'D') || ' ' ||
+      setweight(to_tsvector('spanish', COALESCE(replace(extra, '\\u0000', '')::jsonb->'s3File'->'content','{}')), 'D')
```
pyphilia, line 38: It's probably not a so good idea to add all the languages. But if we handle only the item language it means by default all existing item will have the english language, and only once you edit this setting you might get a good search. 
But maybe a good alternative would be to handle english and french by default (since they are our main languages) and all the other languages will "only" work if the setting is correct. (if the language is correctly set to french, finding  "chat" from "chats" works, but it won't if it's set to english)
```
vector(english) || vector(french) || vector(item language)  
```
That means we need to set a language! in a new column? What do you think?

src/services/item/plugins/geolocation/service.ts
```diff
@@ -42,20 +42,16 @@ export class ItemGeolocationService {
...
-      lat2,
-      lng1,
-      lng2,
-    }: {
+    query: {
       lat1: ItemGeolocation['lat'];
       lat2: ItemGeolocation['lat'];
       lng1: ItemGeolocation['lng'];
       lng2: ItemGeolocation['lng'];
+      search?: string[];
```
spaenleh, line 50: This is called `keywords` in the service no ?

spaenleh (approved):
Looks good, though you might have a typo 🍪

PR #805, opened 2024-02-05
fix: filter out recycled items on get geolocation

### LINKED ISSUE #804: cannot return geolocation if one item is deleted

pyphilia:
fix #804

### DIFF
--- a/src/services/item/plugins/geolocation/repository.ts
+++ b/src/services/item/plugins/geolocation/repository.ts
@@ -74,7 +74,7 @@ export class ItemGeolocationRepository {
       // inner join to filter out recycled items
       .innerJoinAndSelect('ig.item', 'item')
       .leftJoinAndSelect('item.creator', 'member')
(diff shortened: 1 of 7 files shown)


src/services/item/plugins/geolocation/repository.ts
```diff
@@ -71,7 +71,8 @@ export class ItemGeolocationRepository {
 
     const geoloc = this.repository
       .createQueryBuilder('ig')
-      .leftJoinAndSelect('ig.item', 'item')
+      // inner join to filter out recycled items
+      .innerJoinAndSelect('ig.item', 'item')
```
spaenleh, line 75: Ah yes ... this is the real pain of using the soft-delete feature ... I recall having to hunt down usage like these with Morgan ... Not fun.

spaenleh (approved):
Looks good, but maybe the `omit` was not such a great idea after all... see the type errors you get...

PR #814, opened 2024-02-07
feat: get geolocation within item

### LINKED ISSUE #813: Get geolocations within item

pyphilia:
close #813

### DIFF
--- a/src/services/item/plugins/geolocation/repository.ts
+++ b/src/services/item/plugins/geolocation/repository.ts
@@ -6,6 +6,7 @@ import { ALLOWED_SEARCH_LANGS } from '../../../../utils/config';
 import { Actor } from '../../../member/entities/member';
 import { Item } from '../../entities/Item';
 import { ItemGeolocation } from './ItemGeolocation';
+import { MissingGeolocationSearchParams } from './errors';
 
 export class ItemGeolocationRepository {
   private repository: Repository<ItemGeolocation>;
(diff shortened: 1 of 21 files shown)


src/services/item/plugins/geolocation/repository.ts
```diff
@@ -59,28 +62,54 @@ export class ItemGeolocationRepository {
...
+      lat2?: ItemGeolocation['lat'];
+      lng1?: ItemGeolocation['lng'];
+      lng2?: ItemGeolocation['lng'];
       keywords?: string[];
     },
+    parentItem?: Item,
   ): Promise<ItemGeolocation[]> {
-    const [minLat, maxLat] = [lat1, lat2].sort((a, b) => a - b);
-    const [minLng, maxLng] = [lng1, lng2].sort((a, b) => a - b);
+    // should include at least parentItem or all lat/lng
```
ReidyT, line 73: It is possible to reflect this constraint in the type of the function to enforce to call it correctly ?
If it is too much code or too difficult, I think it is interesting to add a comment in the method's doc (Javadoc) to indicate this constraint.

Here could be a solution ? Maybe not !

```
type Boundaries = {
  lat1: number;
  lat2: number;
  lng1: number;
  lng2: number;
}

type ParentItem = string;

type GetItemsInParams =
  | { parentItem: ParentItem }
  | { boundaries: Boundaries }
  | { parentItem: ParentItem; boundaries: Boundaries };

const validateParams = (params: GetItemsInParams) => {
  if (!('parentItem' in params) && !('boundaries' in params)) {
    throw new Error('Either parentItem or boundaries is required');
  }
}
```

src/services/item/plugins/geolocation/service.ts
```diff
@@ -43,15 +43,22 @@ export class ItemGeolocationService {
     actor: Actor,
     repositories: Repositories,
     query: {
-      lat1: ItemGeolocation['lat'];
-      lat2: ItemGeolocation['lat'];
-      lng1: ItemGeolocation['lng'];
-      lng2: ItemGeolocation['lng'];
+      parentItemId?: Item['id'];
```
ReidyT, line 46: Maybe it could be interesting to do the same as in the repository ? Trying to enforce the mandatory of types.

spaenleh (approved):

PR #830, opened 2024-02-20
fix: send websockets after the transactions are done

### LINKED ISSUE #806: Snd the web socket events after the transaction
For now, the WS is send inside the move transaction. It should be send after to allow the query-client to invalidate the queries correctly.

ReidyT:


ReidyT:
This PR will be split into multiple PR for each action (move, copy, ...). This one will be close. The move PR can be found to #835.

PR #835, opened 2024-02-21
feat: use one transaction per move item and send ws after the transaction

### LINKED ISSUE #831: Update the web socket feedback of the move items
The idea is to send a web socket containing the updated data each time a transaction is terminated for the moved item. At the end of all the transactions (when moving multiple items), a feedback web socket is sent to display that the task is terminated in the frontend. It could be great to inform which item wasn't move when an error occurred.

To accomplish this, the transaction may have to be go down (not in the controller anymore ?).

The same things will be done for copy, delete, recycle and validate when this one is working.

ReidyT:
- use one transaction for each move item
- publish websocket post move item after the transaction
- publish websocket feedback after all the transactions

For more details, please see #831 

# Validation of the Implementation

**Protocol of the Test**:

- Run each test five times to obtain an average time.
- The time is computed as the highest end time minus the smallest start time.
    - This is because the maximum number of moves is 20, so to move 100 items, it will require 5 API calls.
- For each test, the items and actions are reset.
    - There are 100 items, each with 10 children => 1000 items.
    - There are 1000 actions per item => 1,000,000 actions.
- For each test with web sockets, the builder is forced to refresh.

**Tests 1 - Parallel vs Series with Multiple Transactions without Intermediate Web Sockets**

|                    | In Series | In Parallel |
|:-------------------|:----------|:------------|
|        Time 1 (ms) |   26,916  |   33,155    |
|        Time 2 (ms) |   24,872  |   27,055    |
|        Time 3 (ms) |   30,011  |   35,195    |
|        Time 4 (ms) |   28,483  |   25,567    |
|        Time 5 (ms) |   29,027  |   34,580    |
| **Avg. Time (ms)** | **27,862**| **31,110**  |

Results: Parallel seems to be slower than in series. Also, in series, the feedback isn’t sent at the end of the 20 items but at the whole end.

> [!IMPORTANT]  
> In parallel, each promise has its own thread freezing the database for the time that the move is run. This is not acceptable.

> [!NOTE]
> Maybe use a thread pool (running 5 threads, in each thread having 20 items to move in series…)

**Test 2 - Only One Transaction, moving items are executed in parallel, without web sockets**

|                    | In Parallel |
|:-------------------|:------------|
|        Time 1 (ms) |   27,455     |
|        Time 2 (ms) |   27,808     |
|        Time 3 (ms) |   26,761     |
|        Time 4 (ms) |   33,306     |
|        Time 5 (ms) |   30,993     |
| **Avg. Time (ms)** | **29,265**  |

**Test 3 - Parallel vs Series with Multiple Transactions and web sockets**

> [!IMPORTANT]  
> Refresh the browser to be sure to have a web socket connection!!!

The problem principally comes from the rerender of the table… Indeed, when disabling the invalidation, the manual manipulations conduct to a re-render, involving the HTTP Requests.

|                    | In Series | In Parallel |
|:-------------------|:----------|:------------|
|        Time 1 (ms) |   71,963  |   54,972    |
|        Time 2 (ms) |   64,392  |   79,857    |
|        Time 3 (ms) |   75,210  |   74,077    |
|        Time 4 (ms) |   68,794  |   40,444    |
|        Time 5 (ms) |   63,661  |   51,855    |
| **Avg. Time (ms)** |   68,804  |   60,241    |

Results: In parallel is a little bit faster.

### DIFF
--- a/src/services/item/service.ts
+++ b/src/services/item/service.ts
@@ -23,7 +23,7 @@ import {
   UnauthorizedMember,
 } from '../../utils/errors';
 import HookManager from '../../utils/hook';
-import { Repositories } from '../../utils/repositories';
+import { Repositories, buildRepositories } from '../../utils/repositories';
 import { filterOutItems, validatePermission, validatePermissionMany } from '../authorization';
 import { Actor, Member } from '../member/entities/member';
 import { mapById } from '../utils';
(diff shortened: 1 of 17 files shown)


src/services/item/ws/services.ts
```diff
@@ -0,0 +1,115 @@
...
+    websockets.publish(itemTopic, sourceParentId, ChildItemEvent('delete', source));
+  }
+
+  const destParentId = getParentFromPath(destination.path);
+  if (destParentId) {
+    websockets.publish(itemTopic, destParentId, ChildItemEvent('create', destination));
+  }
+};
+
+export const publishAfterMoved = (
```
pyphilia, line 65: I don't really like the name. Actually if you create a class to group functions (later copy, update, etc) you won't need to pass the websocket service everytime.

```ts
class ItemWebsocketsService {
   constructor(websockets: WebsocketService) { ... }

   publishTopicsForMove({ source, destination, sourceParentId }) { ... }

   publishFeedbacksForMove(...) { ... }
}
```
and use 
```ts
const itemWebsocketsService = new ItemWebsocketsService(...)

itemWebsocketsService.publishTopicsForMove()
```

src/services/item/test/ws.test.ts
```diff
@@ -433,18 +434,24 @@ describe('Item websocket hooks', () => {
       });
       expect(response.statusCode).toBe(StatusCodes.ACCEPTED);
 
-      await waitForExpect(async () => {
-        expect((await ItemRepository.findOneBy({ id: childItem.id }))?.path).toContain(
-          newParentItem.path,
-        );
-      });
+      // When the websocket is not received, indicating that the transaction
```
pyphilia, line 437:
```suggestion
      // When the websocket is not received, indicates that the transaction
```

src/services/item/ws/services.ts
```diff
@@ -0,0 +1,115 @@
...
+  const successIds = success.map((i) => i.id);
+  const failedIds = itemIds.filter((id) => !successIds.includes(id));
+
+  if (success.length) {
+    websockets.publish(
+      memberItemsTopic,
+      memberId,
+      ItemOpFeedbackEvent(
+        'move',
+        success.map((i) => i.id),
```
pyphilia, line 97:
```suggestion
        successIds,
```

src/services/item/ws/services.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+  if (failed.length) {
+    failed.forEach((e) => {
+      log.error(e);
+      websockets.publish(
+        memberItemsTopic,
+        memberId,
+        ItemOpFeedbackEvent('move', failedIds, { error: e }),
+      );
+    });
```
pyphilia, line 113: Why not having one feedback with the errors within the "success" message and let the frontend trigger the multiple notifications? I know we discussed it sorry..

pyphilia (changes requested):
Thanks for the PR! Here are some ideas to refactor the code, plus some questions 🐈

src/services/item/service.ts
```diff
@@ -389,13 +390,6 @@ export class ItemService {
       await itemRepository.checkHierarchyDepth(parentItem, levelsToFarthestChild);
     }
 
-    // post hook
-    // question: invoque on all items?
-    await this.hooks.runPreHooks('move', actor, repositories, {
-      source: item,
-      destinationParent: parentItem,
-    });
```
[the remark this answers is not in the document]
ReidyT: I didn't found one listening for move preHooks...

src/services/item/ws/services.ts
ReidyT: Yes, I started to do something similar in copy. It's a good idea, I will add it in move PR too.

src/services/item/ws/services.ts
ReidyT: Because there are some changes to do... but I can actually do this now.

PR #838, opened 2024-02-23
fix: use sdk esm version

spaenleh:
Update to use esm version of SDK.

The `HttpMethods` enum was changed which leads to a lot of changes but in fact there are not no many things to adapt.


ReidyT (approved):
LGTM 🙌 🚀

pyphilia (approved):
LGTM 🦜

PR #852, opened 2024-03-01
feat: add descriptionPlacement support in item settings

ReidyT:
This PR refer to graasp/graasp-ui#694.

### DIFF
--- a/src/services/item/fluent-schema.ts
+++ b/src/services/item/fluent-schema.ts
@@ -25,7 +25,8 @@ const settings = S.object()
   .prop('isResizable', S.boolean())
   .prop('hasThumbnail', S.boolean())
   .prop('ccLicenseAdaption', S.string())
-  .prop('displayCoEditors', S.boolean());
+  .prop('displayCoEditors', S.boolean())
+  .prop('descriptionPlacement', S.string());
 
 export const partialMember = S.object()
   .additionalProperties(false)
--- a/src/services/item/fluent-schema.ts
+++ b/src/services/item/fluent-schema.ts
@@ -2,6 +2,7 @@
 import S, { JSONSchema, ObjectSchema } from 'fluent-json-schema';
 
 import {
+  DescriptionPlacement,
   ItemType,
   MAX_ITEM_NAME_LENGTH,
   MAX_TARGETS_FOR_MODIFY_REQUEST,
@@ -26,7 +27,7 @@ const settings = S.object()
   .prop('hasThumbnail', S.boolean())
   .prop('ccLicenseAdaption', S.string())
   .prop('displayCoEditors', S.boolean())
-  .prop('descriptionPlacement', S.string());
+  .prop('descriptionPlacement', S.enum(Object.values(DescriptionPlacement)));
 
 export const partialMember = S.object()
   .additionalProperties(false)
--- a/src/services/item/fluent-schema.ts
+++ b/src/services/item/fluent-schema.ts
@@ -18,8 +18,8 @@ import { Ordering, SortBy } from './types';
  * for serialization
  */
 const settings = S.object()
-  // allow additional properties mostly for apps custom settings
-  .additionalProperties(true)
+  // Setting additional properties to false will only filter out invalid properties.
+  .additionalProperties(false)
   .prop('isPinned', S.boolean())
   .prop('tags', S.array())
   .prop('showChatbox', S.boolean())
--- a/src/services/item/fluent-schema.ts
+++ b/src/services/item/fluent-schema.ts
@@ -20,6 +20,8 @@ import { Ordering, SortBy } from './types';
 const settings = S.object()
   // Setting additional properties to false will only filter out invalid properties.
   .additionalProperties(false)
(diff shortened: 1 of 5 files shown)


src/services/item/repository.ts
```diff
@@ -92,13 +94,18 @@ export const ItemRepository = AppDataSource.getRepository(Item).extend({
       parsedExtra = { folder: { childrenOrder: [] } };
     }
 
+    const finalSettings = {
+      ...settings,
+      ...(settings.hasThumbnail === undefined ? DEFAULT_THUMBNAIL_SETTING : {}),
```
pyphilia, line 99: Why not destructure before `...settings` in all cases? Are there cases where we don't want to add the default thumbnail props?
Actually I think if you do 
```
...DEFAULT_SETTINGS,
...settings
```
it should work as well, won't it?

src/services/item/repository.ts
```diff
@@ -31,7 +31,9 @@ import { ItemChildrenParams } from './types';
 import { _fixChildrenOrder, sortChildrenWith } from './utils';
 
 type ItemSettings = DiscriminatedItem['settings'];
-export const DEFAULT_ITEM_SETTINGS: ItemSettings = {
+export const DEFAULT_ITEM_SETTINGS: ItemSettings = {};
```
pyphilia, line 34: Do you still use it? Maybe see my other comment before changing here.

src/services/item/fluent-schema.ts
```diff
@@ -25,7 +25,8 @@ const settings = S.object()
   .prop('isResizable', S.boolean())
   .prop('hasThumbnail', S.boolean())
   .prop('ccLicenseAdaption', S.string())
-  .prop('displayCoEditors', S.boolean());
+  .prop('displayCoEditors', S.boolean())
+  .prop('descriptionPlacement', S.string());
```
pyphilia, line 29: Not sure but maybe you can have an enum here?

src/services/item/test/index.test.ts
```diff
@@ -237,6 +238,23 @@ describe('Item routes tests', () => {
...
+          method: HttpMethod.Post,
+          url: `/items`,
+          payload,
+        });
+
+        const newItem = response.json();
+        expectItem(newItem, payload, actor);
+        expect(response.statusCode).toBe(StatusCodes.OK);
+        expect(newItem.settings.descriptionPlacement).toBe(DescriptionPlacement.ABOVE);
+        expect(newItem.settings.hasThumbnail).toBeFalsy();
```
pyphilia, line 255: Are you checking `false` or `undefined`? `isFalsy` checks both, so it might be worth to use either `isUndefined` or `isEqual(false)`

src/services/item/test/index.test.ts
```diff
@@ -1927,6 +1945,35 @@ describe('Item routes tests', () => {
...
+        const newItem = response.json();
+
+        // this test a bit how we deal with extra: it replaces existing keys
+        expectItem(newItem, {
+          ...item,
+          ...payload,
+        });
+        expect(response.statusCode).toBe(StatusCodes.OK);
+        expect(newItem.settings.descriptionPlacement).toBe(DescriptionPlacement.ABOVE);
+        expect(newItem.settings.hasThumbnail).toBeFalsy();
```
pyphilia, line 1974: Same as above.

pyphilia (changes requested):
Overall changes are good, I just have a question about how you actually deal with thumbnails' properties, thanks for the PR! 🍳

src/services/item/repository.ts
ReidyT: Yes it should, I will update it 👍

src/services/item/test/index.test.ts
ReidyT: Ah okay, I will change that

src/services/item/fluent-schema.ts
ReidyT: Yep, you're right

src/services/item/test/index.test.ts
```diff
@@ -1927,6 +1945,35 @@ describe('Item routes tests', () => {
...
+        // this test a bit how we deal with extra: it replaces existing keys
+        expectItem(newItem, {
+          ...item,
+          ...payload,
+        });
+        expect(response.statusCode).toBe(StatusCodes.OK);
+        expect(newItem.settings.descriptionPlacement).toBe(DescriptionPlacement.ABOVE);
+        expect(newItem.settings.hasThumbnail).toBe(false);
+      });
+
```
pyphilia, line 1976: Add a test that throws for updating a setting that is not allowed? Since you nicely added this check in the schema!

pyphilia (approved):
👍

src/services/item/test/index.test.ts
ReidyT: As mentioned in Slack, the schema will only filter the bad settings without throwing. For now, I written some tests to check this behaviour. If we want to throw an error, we will have to update the code.

PR #855, opened 2024-03-04
fix: reuse item login from parent

### LINKED ISSUE #854: [item-login] item login does not reuse from parent

pyphilia:
fix #854

### DIFF
--- a/src/services/itemLogin/test/index.test.ts
+++ b/src/services/itemLogin/test/index.test.ts
@@ -143,7 +143,7 @@ describe('Item Login Tests', () => {
         expect(result.item.id).toEqual(itemLoginSchema.item.id);
       });
 
-      it('Successfully get item login for child', async () => {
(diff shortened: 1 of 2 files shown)


src/services/itemLogin/test/index.test.ts
```diff
@@ -267,6 +267,24 @@ describe('Item Login Tests', () => {
             expectItemLogin(member, m);
             expect(res.statusCode).toBe(StatusCodes.OK);
           });
+
+          it('Successfully reuse item login with username from child', async () => {
```
spaenleh, line 271:
```suggestion
          it('Successfully reuse item login with username defined in parent when calling from child', async () => {
```

spaenleh (approved):
Looks good, just rename the test to make it clearer that we try to reuse the one defined in the parent from inside the child. 

🍁

PR #856, opened 2024-03-04
feat: add geocoding endpoints

### LINKED ISSUE #850: Use geoapify proxied by the backend

pyphilia:
close #850

todo: 
- [ ] Change secrets

### DIFF
--- a/src/services/item/plugins/geolocation/index.test.ts
+++ b/src/services/item/plugins/geolocation/index.test.ts
@@ -425,17 +425,18 @@ describe('Item Geolocation', () => {
     });
 
     describe('Signed out', () => {
-      it('get adress from reverse', async () => {
+      it('Throw', async () => {
         ({ app } = await build({ member: null }));
         const res = await app.inject({
           method: HttpMethod.Get,
           url: `${ITEMS_ROUTE_PREFIX}/geolocation/reverse`,
           query: {
-            lat: 1,
+            lat: 2,
             lng: 2,
           },
         });
-        expect(res.json()).toMatchObject({ addressLabel: 'address', country: 'country' });
+
+        expect(res.statusCode).toBe(StatusCodes.UNAUTHORIZED);
       });
     });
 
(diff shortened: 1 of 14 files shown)


src/services/item/plugins/geolocation/index.test.ts
```diff
@@ -12,6 +13,8 @@ import { saveMember } from '../../../member/test/fixtures/members';
 import { savePublicItem } from '../../test/fixtures/items';
 import { ItemGeolocation } from './ItemGeolocation';
 
+jest.mock('node-fetch');
```
spaenleh, line 16: We don't use `nock` ?

src/utils/config.ts
```diff
@@ -386,3 +386,11 @@ export const ALLOWED_SEARCH_LANGS = {
   fr: 'french',
   it: 'italian',
 };
+
+// Geolocation API Key
+if (!process.env.GEOLOCATION_API_KEY) {
```
spaenleh, line 391: Add in the env description in the Readme ? Also once stable it should be added to the docs. Will this be required when running the backend ? Can we provide any value if we do not plan to use that feature ?

src/services/item/plugins/geolocation/index.test.ts
```diff
@@ -410,4 +413,162 @@ describe('Item Geolocation', () => {
       });
     });
   });
+
+  describe('GET /geolocation/reverse', () => {
+    beforeEach(() => {
+      (fetch as jest.MockedFunction<typeof fetch>).mockImplementation(async () => {
```
ReidyT, line 419: should this be in fixtures ?

spaenleh (approved):

ReidyT (approved):
LGTM 👍

src/services/item/plugins/geolocation/index.test.ts
pyphilia: I never use nock, only alex did 🤷‍♀️

PR #861, opened 2024-03-05
fix: order descendants by default

### LINKED ISSUE #860: order descendants
Improve ordering of descendants (still flat), set `true` by default.

pyphilia:
fix #860

### DIFF
--- a/src/services/item/utils.ts
+++ b/src/services/item/utils.ts
@@ -59,14 +59,10 @@ export const sortChildrenWith = (idsOrder: string[]) => (stElem: Item, ndElem: I
   return stElem.createdAt.getTime() - ndElem.createdAt.getTime();
 };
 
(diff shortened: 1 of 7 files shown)


src/services/item/utils.ts
```diff
@@ -59,6 +59,28 @@ export const sortChildrenWith = (idsOrder: string[]) => (stElem: Item, ndElem: I
   return stElem.createdAt.getTime() - ndElem.createdAt.getTime();
 };
 
+const getDirectChildren = (descendants: Item[], parentItem: Item): Item[] => {
+  return descendants.filter((child) => isChildOf(child.path, parentItem.path));
+};
```
pyphilia, line 64: Remove and use directly in below function.

spaenleh (approved):
Looks good !

PR #867, opened 2024-03-11
feat: use item id for actions

### LINKED ISSUE #817: Use item_id instead of item_path in action
Changing an item shouldn't trigger a cascade in action because it slows down a lot.

This will imply to change how actions are fetched.

- [x] refactor and test action repository
- [x] test action service
- [x] test item action service
- [ ] test item action controller - already done beforehand
- [ ] check analytics still works
- [ ] refactor to use item id instead of item path in action

pyphilia:
close #817


src/services/action/repositories/action.ts
```diff
@@ -3,6 +3,7 @@ import { EntityManager, Repository } from 'typeorm';
 import { AggregateBy, AggregateFunction, AggregateMetric, CountGroupBy, UUID } from '@graasp/sdk';
 
 import { AppDataSource } from '../../../plugins/datasource';
+import { printFilledSQL } from '../../../utils/debug';
```
spaenleh, line 6: Remove ?

spaenleh (approved):
Looks good to me, I trust you on the data migration :)

Issue #837, opened 2024-02-22
Include highest permission for member in items + tags


pyphilia:
We would like to directly get the best permission for a member on an item from the GET /itemXX endpoints.

For slow integration, we can start with GET /many-items that mainly impacts the table in builder.

Include tags as well.

PR #871, opened 2024-03-12
feat: return permission alongside item

pyphilia:
Would be nice to have a discussion on naming.
For now I've only added the permission to `get` and `getMany` but the idea would be to also have it for all items' GET endpoints.


pyphilia:
ref #837

src/services/item/ItemWrapper.ts
```diff
@@ -0,0 +1,82 @@
...
+  | 'description'
+  | 'extra'
+  | 'createdAt'
+  | 'creator'
+  | 'updatedAt'
+  | 'settings'
+  | 'lang'
+>;
+
+// TODO: export in sdk??
```
spaenleh, line 22: So ? if we took a decision, I think we can remove this.

src/services/item/fluent-schema.ts
```diff
@@ -327,5 +346,29 @@ export default {
...
+          additionalProperties: true,
+        },
+        settings: {
+          type: 'object',
+          additionalProperties: true,
+        },
+        creator: { $ref: 'https://graasp.org/members/#/definitions/member' },
+        createdAt: { type: 'string' },
+        updatedAt: { type: 'string' },
+        permission: { type: ['string', 'null'] },
```
spaenleh, line 370: set as enum ?

src/services/item/plugins/geolocation/index.test.ts
```diff
@@ -522,7 +596,25 @@ describe('Item Geolocation', () => {
         ({ app, actor } = await build());
       });
 
-      it('get adress from search', async () => {
+      it.only('get address from search', async () => {
```
spaenleh, line 599: Remove `.only` ?

spaenleh (approved):
Wow ! looks good, just a few things to cleanup and we are good !

PR #886, opened 2024-03-18
fix: return null when public profile is not visible

spaenleh:


LinaYahya (approved):

src/services/member/plugins/profile/repository.ts
```diff
@@ -33,12 +30,17 @@ const MemberProfileRepository = AppDataSource.getRepository(MemberProfile).exten
...
+    filter?: {
+      visibility?: boolean;
+    },
+  ): Promise<MemberProfile | null> {
     if (!memberId) {
       throw new MemberNotFound();
     }
     const memberProfile = await this.findOne({
-      where: { member: { id: memberId }, ...filter },
+      where: { member: { id: memberId }, visibility: filter?.visibility },
```
pyphilia, line 43: Would this need a default value? I guess it does not consider the prop if it is `undefined`, but not sure?

pyphilia (approved):

src/services/member/plugins/profile/repository.ts
spaenleh: It should safely send undefined when the visibility is not defined.

PR #891, opened 2024-03-22
feat: add display name column

spaenleh:


src/services/item/service.ts
```diff
@@ -154,7 +154,7 @@ export class ItemService {
     const item = await repositories.itemRepository.get(id);
 
     await validatePermission(repositories, permission, actor, item);
-
+    console.log(item);
```
pyphilia, line 157: Remove!

pyphilia (changes requested):

pyphilia:
I've actually deployed both front and back but you never receive the `display_name` from the get endpoints. Is it normal? You can see that if you post a document with a `display_name`, you try to edit but the `display_name` is not the same.

pyphilia (approved):

PR #897, opened 2024-03-22
fix: add CORS on status endpoint

spaenleh:
Add CORS headers on `/status` endpoint to allow querying the status from Auth frontend.


src/plugins/meta.ts
```diff
@@ -1,3 +1,4 @@
+import { rejects } from 'assert';
```
pyphilia, line 1: Remove.

pyphilia (approved):

PR #900, opened 2024-03-25
feat: add webm to allowed mimetypes for h5p import

pyphilia:


spaenleh (commented):
mhmmm looks like `webm` is already mentioned on line 50 ...

PR #903, opened 2024-03-25
fix: do not expose port 1025 to the local machine

spaenleh:
This PR removed the exposed port `1025` and reduces it to be accessible only inside docker network. We do not need this port to be exposed publicly on the machine and it can be seen as a security vulnerability.


pyphilia (approved):

juancarlosfarah (approved):
🚀

PR #911, opened 2024-03-26
fix: do not expose ports that are not necessary

spaenleh:


.devcontainer/docker-compose.yml
```diff
@@ -20,7 +20,10 @@ services:
       # temporary storage for files inside the container
       FILE_STORAGE_ROOT_PATH: /tmp/graasp-file-item-storage
       # the localfile config host points to the static file server defined below
-      FILE_STORAGE_HOST: http://localhost:1081
+      FILE_STORAGE_HOST: http://localfile:80
+      PUBLIC_FILE_STORAGE_HOST: http://localhost:1081
```
pyphilia, line 24: Why not use FILE_STORAGE_HOST with public port directly? In which case does `http://localfile:80` work?

pyphilia (approved):

.devcontainer/docker-compose.yml
spaenleh: @pyphilia I thought it would be interesting to fetch images for validation for example using the local url (inside the container). and use the public url when sending to clients. But I guess it is not that useful. Should I remove it ?

PR #920, opened 2024-04-04
feat: allow user to enable/disable saving actions

ReidyT:
This PR:
- allows new users to enable or disable saving actions
- existing users have enableSaveActions to null
  - This allows to facilitate what to do with existing users like:
     - Stop saving actions until they explicitly accept it
     - Continue saving actions
     - Also allows to ask the user to choose 
  - This wouldn't be possible if default value would be set to false
- allows all users to update this option
- ensure that actions can't be saved if the option is not true (or null if wanted, should be decided)

### DIFF
--- a/src/services/member/repository.ts
+++ b/src/services/member/repository.ts
@@ -75,7 +75,7 @@ export const MemberRepository = AppDataSource.getRepository(Member).extend({
       newData.extra = Object.assign({}, body?.extra, body.extra);
     }
 
-    if (body.enableSaveActions === true || body.enableSaveActions === false) {
+    if (typeof body.enableSaveActions === 'boolean') {
       newData.enableSaveActions = body.enableSaveActions;
     }
 
--- a/src/services/member/repository.ts
+++ b/src/services/member/repository.ts
@@ -95,7 +95,12 @@ export const MemberRepository = AppDataSource.getRepository(Member).extend({
     // The auth frontend only blocks the user to create an account without checking the boxes.
     // The frontend avoids sending agreement data to prevent manipulation of the agreement date.
     // The agreements links are included in the registration email as a reminder.
-    const createdMember = await this.insert({ ...data, email, userAgreementsDate: new Date() });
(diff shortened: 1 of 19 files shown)


src/services/member/repository.ts
```diff
@@ -72,6 +75,10 @@ export const MemberRepository = AppDataSource.getRepository(Member).extend({
       newData.extra = Object.assign({}, body?.extra, body.extra);
     }
 
+    if (typeof body.enableSaveActions === 'boolean') {
```
pyphilia, line 78: Can this really be undefined or null? I suppose you can only send a boolean value from the endpoint. Then I guess the service should provide a valid value.

src/services/member/repository.ts
```diff
@@ -88,7 +95,17 @@ export const MemberRepository = AppDataSource.getRepository(Member).extend({
     // The auth frontend only blocks the user to create an account without checking the boxes.
     // The frontend avoids sending agreement data to prevent manipulation of the agreement date.
     // The agreements links are included in the registration email as a reminder.
-    const createdMember = await this.insert({ ...data, email, userAgreementsDate: new Date() });
+    const userAgreementsDate = new Date();
+    // enableSaveActions defaults to null in the database.
+    // This allows us to differentiate existing users who haven't explicitly agreed to saving actions
+    // from new users who explicitly accept (true) or deny (false).
+    const enableSaveActions = data.enableSaveActions ?? false;
```
pyphilia, line 102: Default to true?

pyphilia (changes requested):
I've sent a message to check whether it's okay to have the saving by default. I'll let you know!

src/services/member/repository.ts
ReidyT: Yes but because it is a patch, the user can do a partial update of the member. That means enableSaveActions could be missing in the request body.

spaenleh (approved):
Looks good to me.

pyphilia (approved):

PR #941, opened 2024-04-12
fix: password login should return `MemberNotSignedUp` if email has no account

### LINKED ISSUE #940: Harmonize error returned on invalid email for `/login` and `/login-password`
When a user tries to login with an email that is not linked to an account the reply is: MEMBER_NOT_SIGNED_UP
When a user tries to login with an email and a password and the email is not linked to an account the reply is: MEMBER_DOES_NOT_EXIST

Was there a reason to have different errors for the seemingly same "problem" ?

spaenleh:
In this PR I fix a bug where calling `/login-password` with an email that is not associated with an account return `Account not found`.
To align with the behaviour of the `/login` endpoint it should return `Member not sign up`.
fix #940


pyphilia (approved):
I guess you could have a test for that ;) 
Thanks for the PR!

PR #959, opened 2024-04-17
fix: optimize get published collections for member

### LINKED ISSUE #958: Optimize get published items for member

spaenleh:
Resolves #958 
In this PR I try to optimize the get published collections for member.
This query is remarkably slow and is executed for the home page of the library in the first section. 

This is the query execution plan for the query before optimization:
Screenshot 2024-04-16 at 18 51 04
The execution time is around ~~9 seconds~~ (2 seconds on a hot database probably because of caching), that is unacceptable

This is the execution plan of the new query:
Screenshot 2024-04-16 at 18 51 35
Execution time is ~~2 seconds~~ (between 2 and 10 miliseconds on a hot DB), better ~~but not as good as it could be with some custom SQL queries~~ actually I compared and the "custom" query below does not run significantly faster than the Typeorm optimized query.

I attache below the query that could be used to reduce even further the execution time:
```sql
SELECT "item"."id" AS "item_id",
  "item"."name" AS "item_name",
  "item"."description" AS "item_description",
  "item"."type" AS "item_type",
  "item"."created_at" AS "item_created_at",
  "item"."updated_at" AS "item_updated_at",
  "item"."deleted_at" AS "item_deleted_at",
  "item"."extra" AS "item_extra",
  "item"."settings" AS "item_settings",
  "item"."path" AS "item_path",
  "item"."creator_id" AS "item_creator_id",
  "member"."id" AS "member_id",
  "member"."name" AS "member_name",
  "member"."email" AS "member_email",
  "member"."type" AS "member_type",
  "member"."extra" AS "member_extra",
  "member"."created_at" AS "member_created_at",
  "member"."updated_at" AS "member_updated_at"
FROM "member" "member"
  INNER JOIN "item_membership" "im" ON "im"."member_id" = "member"."id"
  AND "member"."id" = '049b4f35-f12e-4ec6-8958-586113bbd241'
  RIGHT JOIN "item_published" "pi" ON "im"."item_path" @> "pi"."item_path"
  INNER JOIN "item" "item" ON "item"."path" = "pi"."item_path"
WHERE ("im"."permission" IN ('admin', 'write'))
  AND ("item"."deleted_at" IS NULL);
```
This query can not be used in TypeORM, since we do not have access to right join, and because typeorm needs you to "start" from the entity that you want to get as output. In the case of this query, we start with member as this is what will reduce the size of the joins, but we actually want to get an item back. There is no way to make that query by starting from item or item publish and be as efficient in the size of the joins.

### DIFF
--- a/src/services/item/plugins/published/repositories/itemPublished.ts
+++ b/src/services/item/plugins/published/repositories/itemPublished.ts
@@ -1,14 +1,30 @@
+import { EntityManager, Repository } from 'typeorm';
+import { QueryDeepPartialEntity } from 'typeorm/query-builder/QueryPartialEntity';
+
+import { PermissionLevel } from '@graasp/sdk';
+
 import { AppDataSource } from '../../../../../plugins/datasource';
-import { Actor } from '../../../../member/entities/member';
+import { Actor, Member } from '../../../../member/entities/member';
 import { mapById } from '../../../../utils';
 import { Item } from '../../../entities/Item';
 import { ItemPublished } from '../entities/itemPublished';
 import { ItemPublishedNotFound } from '../errors';
 
-export const ItemPublishedRepository = AppDataSource.getRepository(ItemPublished).extend({
+export class ItemPublishedRepository {
+  private repository: Repository<ItemPublished>;
+
+  constructor(manager?: EntityManager) {
+    if (manager) {
+      this.repository = manager.getRepository(ItemPublished);
+    } else {
+      this.repository = AppDataSource.getRepository(ItemPublished);
+    }
+  }
+
   async getForItem(item: Item): Promise<ItemPublished> {
     // this returns the root published item when querying a child item
-    const entry = await this.createQueryBuilder('pi')
+    const entry = await this.repository
+      .createQueryBuilder('pi')
       .innerJoinAndSelect('pi.item', 'item', 'pi.item @> :itemPath', { itemPath: item.path })
       .innerJoinAndSelect('pi.creator', 'member')
       // Order isn't guaranteed so we must force it to avoid flaky results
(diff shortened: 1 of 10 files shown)


ReidyT (approved):
Really nice PR  🙌. LGTM.
It's a good improvement comparing to before the optimisation.
I just have few questions concerning the SQL query:
1. Did you analyse it to see if it is effectively faster comparing to your last optimisation ?
2. If yes, isn't it possible to just use the raw SQL in TypeORM ?

spaenleh:
After some more analysis, the optimized version runs 3 orders of magnitude faster than the current version. However the "custom sql" version does not run significantly faster. I think we can keep the optimized version with typeorm. 

The bigges win I think comes from adding the memberId in the join condition to reduce the size of the joined ensemble. We should investigate other queries to see if such wins can be made.

src/services/item/plugins/published/repositories/itemPublished.ts
```diff
@@ -38,64 +55,91 @@ export const ItemPublishedRepository = AppDataSource.getRepository(ItemPublished
...
-  },
+  }
+
+  async getForMember(memberId: Member['id']): Promise<Item[]> {
+    const itemPublished = await this.repository
+      .createQueryBuilder('pi')
+      .innerJoin('item_membership', 'im', 'im.item_path @> pi.item_path')
+      .innerJoin('member', 'm', 'im.member_id = m.id and m.id = :memberId', {
+        memberId,
+      })
```
pyphilia, line 66: If it's what we discussed, you don't use a `where` clause because it is only computed at the end of the query, right? Maybe add a comment.

src/services/item/plugins/published/repositories/itemPublished.ts
```diff
@@ -38,64 +55,91 @@ export const ItemPublishedRepository = AppDataSource.getRepository(ItemPublished
...
+      .where('im.permission IN (:...permissions)', {
+        permissions: [PermissionLevel.Admin, PermissionLevel.Write],
+      })
+      .getMany();
+
+    return itemPublished.map(({ item, creator }) => {
+      // use creator of publish entry as the creator of the item
+      item.creator = creator;
+      return item;
+    });
```
pyphilia, line 78: Why? Why not just return `itemPublished` in that case?

pyphilia (changes requested):
I left 2 comments, I think the main one is the return value that seems a bit hacky.
Thanks for the PR!

src/services/item/plugins/published/repositories/itemPublished.ts
spaenleh: Yes exactly, I add a condition in the join to minimize the number of rows that will be joined. I added the same for the memberships.

pyphilia (approved):

PR #978, opened 2024-04-22
fix: use () instead of [] in spaces regex

ReidyT:
- This allow spaces only if there is a character after

The Regex I used last weeks allows having multiple space and to end with a space. I changed the regex to not allow that.


spaenleh (approved):
Nice ! Would you consider adding a few test cases to cover the usage of these regexes ?

PR #979, opened 2024-04-23
fix: upgrade gpt versions

### LINKED ISSUE #767: Change default value for GPT Model version to 3.5-turbo
This model il faster and a bit less precise.
But it also costs less.

spaenleh:
In this PR:
- move `GPTVersion` enum to sdk
- set the default version of the model to be used directly in the controller
- update openai package
close #767


ReidyT (approved):
Really nice, looks good 🚀

PR #984, opened 2024-04-23
fix: return hidden only for writer on export

### LINKED ISSUE #957: [Bug] should not export hidden item

pyphilia:
fix #957


src/services/item/plugins/importExport/service.ts
```diff
@@ -321,6 +321,7 @@ export class ImportExportService {
       archiveRootPath: rootPath,
       archive,
     }).catch((error) => {
+      console.log(error);
```
spaenleh, line 324: Keep this ? Maybe use fastify logger ?

spaenleh (approved):
Yeah !

PR #985, opened 2024-04-24
feat: change mobile link

pyphilia:


pyphilia:
Were there another place we need to change? I tested on dev you can still login (with pwd).

spaenleh (approved):
👍

PR #987, opened 2024-04-25
feat: create an api to get user actions for dashboard view

### LINKED ISSUE #986: Track and revert user actions for user dashboard
Create an endpoint to get user dashboard (personal) actions, such as
- [ ] times the user logged in
- [ ] items created
- [ ] items updated
- [ ] items deleted

All these data by default should be retrieved over the last month, with an ability to identify start/end dates

LinaYahya:
closes #986


src/services/member/plugins/action/service.ts
```diff
@@ -1,15 +1,54 @@
...
+    const { startDate, endDate } = filters;
+    const start = startDate ? new Date(startDate) : getMonthBeforeNow();
+    const end = endDate ? new Date(endDate) : new Date();
+
+    const allowedTypes = [
+      ActionTriggers.MemberLogin,
+      ItemActionType.Create,
+      ItemActionType.Update,
+      ItemActionType.Delete,
+    ];
```
pyphilia, line 38: Why not get everything, or a subset given by the user? It feels a bit random which types you're fetching.

src/services/member/plugins/action/service.ts
```diff
@@ -1,15 +1,54 @@
...
+
+    if (!actor) {
+      throw new UnauthorizedMember(actor);
+    }
+    const data = await actionRepository.getForMember(actor.id, {
```
pyphilia, line 43: I've realised that you might have actions for item you don't have access to anymore. I think it's safer if you run `validatePermissionMany` over all the items (a set of them, since they will exist multiple times in the actions). Also add a comment because that might not be the most efficient way to get the right set of actions... if you can think of another way don't hesitate. You can find an example of what I have in mind in `getParents` on the item service.

pyphilia (changes requested):
Thanks for the PR! I've left some comments let me know if you have questions.

pyphilia:
@LinaYahya thanks for updating the PR, could you actually write tests? for the controller and the service? Thanks!

src/services/member/plugins/action/service.ts
```diff
@@ -1,15 +1,70 @@
...
+import { PermissionLevel } from '@graasp/sdk';
+
+import { CannotModifyOtherMembers, UnauthorizedMember } from '../../../../utils/errors';
 import { Repositories } from '../../../../utils/repositories';
 import { ActionService } from '../../../action/services/action';
+import { validatePermissionMany } from '../../../authorization';
+import { Item, ItemExtraMap } from '../../../item/entities/Item';
 import { Actor } from '../../entities/member';
 
+export const getMonthBeforeNow = () => {
```
pyphilia, line 10:
```suggestion
export const getPreviousMonthFromNow = () => {
```

src/services/member/plugins/action/test/index.test.ts
```diff
@@ -0,0 +1,61 @@
...
+      const response = await app.inject({
+        method: HttpMethod.Get,
+        url: GET_URL,
+      });
+
+      expect(response.statusCode).toEqual(StatusCodes.OK);
+      expect(response.json()).toHaveLength(2);
+    });
+
+    it('Get actions when user signed in specifing start and end dates that doesnt ', async () => {
```
pyphilia, line 47: I think it's missing a word?

src/services/member/plugins/action/test/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+import { AppDataSource } from '../../../../../plugins/datasource';
+import { Action } from '../../../../action/entities/action';
+import { saveActions } from '../../../../action/test/fixtures/actions';
+import { ItemTestUtils } from '../../../../item/test/fixtures/items';
+import { saveMember } from '../../../test/fixtures/members';
+
+const testUtils = new ItemTestUtils();
+const rawRepository = AppDataSource.getRepository(Action);
+
+export const getDateBeforeOrAfterNow = (dateDiff) => {
```
pyphilia, line 12: How can this be before? You use a negative `dateDiff`? Not sure about this, I didn't see where you use it.

src/services/member/plugins/action/test/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const saveActionsWithItems = async (member, { saveActionForNotOwnedItem = false } = {}) => {
+  const item = await testUtils.saveItem({ actor: member });
+  const actions = [
+    {
+      member,
+      createdAt: new Date().toISOString(),
+      type: ActionTriggers.Update,
+      item: item as unknown as DiscriminatedItem,
+    },
```
pyphilia, line 26: What about creating a factory in sdk so you don't need to specify the created date every time?

src/services/member/plugins/action/service.ts
```diff
@@ -1,15 +1,70 @@
...
+      }),
+      actionRepository.getMemberActionsNoNeedForPesrmission(actor.id, {
+        startDate: start,
+        endDate: end,
+      }),
+    ]);
+
+    const setOfItemsToCheckPermission = Array.from(
+      new Map(actionsNeedsPermission.map(({ item }) => [item?.id, item])).values(),
+    );
```
pyphilia, line 50:
```suggestion
    const setOfItemsToCheckPermission = [
      ...new Set(actionsNeedsPermission.map(({ item }) => [item?.id, item])]
    
```
or similar

src/services/member/plugins/action/service.ts
```diff
@@ -1,15 +1,70 @@
...
+    const [actionsNeedsPermission, actionsWithoutPermssion] = await Promise.all([
+      actionRepository.getMemberActionsNeedPesrmission(actor.id, {
+        startDate: start,
+        endDate: end,
+      }),
+      actionRepository.getMemberActionsNoNeedForPesrmission(actor.id, {
+        startDate: start,
+        endDate: end,
+      }),
+    ]);
```
pyphilia, line 46: I think it's faster to get all the data, and then get all the items from the actions. What do you think? It can be expensive to fetch from the db.

src/services/member/plugins/action/service.ts
```diff
@@ -1,15 +1,70 @@
...
+
+    const setOfItemsToCheckPermission = Array.from(
+      new Map(actionsNeedsPermission.map(({ item }) => [item?.id, item])).values(),
+    );
+
+    const memberships = await validatePermissionMany(
+      repositories,
+      PermissionLevel.Read,
+      actor,
+      setOfItemsToCheckPermission as Item<keyof ItemExtraMap>[],
```
pyphilia, line 56: And with above change you don't need the cast anymore.

src/services/member/plugins/action/service.ts
```diff
@@ -1,15 +1,70 @@
...
+    );
+
+    const filteredActionsWithAccessPermission = actionsNeedsPermission.filter((g) => {
+      if (g.item && g?.item?.id in memberships.data) {
+        return true;
+      }
+      return null;
```
pyphilia, line 63: Here you can return `true` if the action does not have an item or if it passes your condition. Otherwise `false`.

src/services/member/plugins/action/schemas.ts
```diff
@@ -4,3 +4,14 @@ import { idParam } from '../../../../schemas/fluent-schema';
 export const deleteAllById = {
   params: idParam,
 };
+
+export const getMemberFilteredActions = {
+  querystring: {
+    type: 'object',
+    properties: {
+      startDate: { type: 'string' },
+      endDate: { type: 'string' },
```
pyphilia, line 13: Not sure it is possible, but do you think you can provide some kind of format? since it's supposed to be dates.

pyphilia (changes requested):
Thanks for the changes! See my comments, let me know if you have questions!

src/services/member/plugins/action/service.ts
spaenleh: I think the initial idea was that there were actions that we would retrieve in any case, and others that we would need to filter based on user permissions. 

So I proposed to have 2 calls to the db that would reduce the logic we would have to do in JS. This may not been the best solution however, I would be open to seeing some tests on bottlenecks and where to place the divide for optimal performance, I also think that the code should read as clearly as possible. Even if we encoure a slight perf penalty.

src/services/member/plugins/action/test/utils.ts
LinaYahya: basically I'm using it to specify start and end date for testing purposes, either the date or days difference can be a positive value to get days after or a negative value to get days before now.

src/services/member/plugins/action/test/utils.ts
LinaYahya: YES, sure!

src/services/member/plugins/action/service.ts
LinaYahya: set will not work with objects as every single object referring to a different reference, or a need to stringify item to use a set, Which I think doesn't really make difference than using `Map`

src/services/member/plugins/action/service.ts
LinaYahya: using this way is more readable somehow, and has less of work with js, as retrieving all data would require classifying actions based on what requires permission and what does not, then validating permission for those that require permission.

pyphilia: From the where conditions I'm quite sure we don't have an index on 4 props, so it will basically be 2 linear scan of the db. I think it will be less hard the db if we do only one call, but I might be wrong. If we do a partition https://www.geeksforgeeks.org/lodash-_-partition-method/ I think it's more or less the same code.

pyphilia (changes requested):

src/services/member/plugins/action/service.ts
```diff
@@ -1,15 +1,82 @@
...
+  date.setMonth(date.getMonth() - 1); // Set the date to one month ago
+  return date;
+};
+
+export const actionTypesWithoutNeedOfPermission: string[] = [
+  ActionTriggers.CollectionView,
+  ActionTriggers.ItemLike,
+  ActionTriggers.ItemUnlike,
+  ActionTriggers.ItemDownload,
+  ActionTriggers.ItemEmbed,
```
pyphilia, line 23: Actually all of these you need to check the permissions. At least you should be able to access it. I think with the current way you check the permissions it is okay.

src/services/member/plugins/action/service.ts
```diff
@@ -1,15 +1,82 @@
...
+  ) {
+    const { actionRepository } = repositories;
+
+    const { startDate, endDate } = filters;
+    const start = startDate ? new Date(startDate) : getPreviousMonthFromNow();
+    const end = endDate ? new Date(endDate) : new Date();
+
+    if (!actor) {
+      throw new UnauthorizedMember(actor);
+    }
```
pyphilia, line 48: Have this at the beginning of the function.

src/services/member/plugins/action/service.ts
```diff
@@ -1,15 +1,82 @@
...
+
+    const actions = await actionRepository.getMemberActions(actor.id, {
+      startDate: start,
+      endDate: end,
+    });
+
+    // filter actions based on permission validity
+    const [actionsWithoutPermssion, actionsNeedsPermission] = partition(actions, (action) => {
+      return actionTypesWithoutNeedOfPermission.indexOf(action.type) > -1;
```
pyphilia, line 57:
```suggestion
      return actionTypesWithoutNeedOfPermission.includes(action.type);
```

src/services/member/plugins/action/service.ts
```diff
@@ -1,15 +1,82 @@
...
+      PermissionLevel.Read,
+      actor,
+      setOfItemsToCheckPermission as Item<keyof ItemExtraMap>[],
+    );
+
+    const filteredActionsWithAccessPermission = actionsNeedsPermission.filter((g) => {
+      if (g.item && g?.item?.id in itemMemberships.data) {
+        return true;
+      }
+      return false;
```
pyphilia, line 75:
```suggestion
      return (g.item && g?.item?.id in itemMemberships.data) 
```

src/services/member/plugins/action/service.ts
```diff
@@ -1,15 +1,82 @@
...
+      throw new UnauthorizedMember(actor);
+    }
+
+    const actions = await actionRepository.getMemberActions(actor.id, {
+      startDate: start,
+      endDate: end,
+    });
+
+    // filter actions based on permission validity
+    const [actionsWithoutPermssion, actionsNeedsPermission] = partition(actions, (action) => {
```
pyphilia, line 56:
```suggestion
    const [actionsWithoutPermission, actionsNeedPermission] = partition(actions, (action) => {
```

pyphilia (changes requested):
Thanks for the changes, let me know if you have questions!

pyphilia (approved):

Issue #823, opened 2024-02-15
Create a seed script that can populate the database with useful and mock data


spaenleh:
Useful data:
- app data
- publisher data
- categories data
- some users
- published items
- user items
- invitations
- others ...

spaenleh:
This could also be combined with our plan to have a way of seeding the database with data for testing.

PR #989, opened 2024-04-25
feat: implements seed function with sample data

Defferrard:
#823 

It is a draft version. We're able to seed the database from command line or by calling the seed() function.
We could now use a lite syntax to generate the parameters needed to call the seed() function


pyphilia (changes requested):
Thank you for this PR!
Maybe let's not stack all the properties on one item. Let's have an item that is hidden, another item that is public, another item that has a geoloc. And reflect on the name as well so it's easier to understand the items.

pyphilia (approved):

PR #992, opened 2024-04-25
feat: add endpoint for most-used apps per user

### LINKED ISSUE #988: Add endpoint to get mostly used apps by user

ztlee042:
In this PR:

an endpoint is added to get the most used apps from the user

fix #988

### DIFF
--- a/src/services/item/plugins/app/fluent-schema.ts
+++ b/src/services/item/plugins/app/fluent-schema.ts
@@ -8,12 +8,23 @@ export const appSchema = S.object()
   .prop('url', S.string())
   .prop('extra', S.object().additionalProperties(true));
 
+export const mostUsedAppSchema = S.object()
+  .additionalProperties(false)
+  .prop('name', S.string())
+  .prop('url', S.string())
+  .prop('nbr', S.number());
+
 export const getMany = {
   response: {
     200: S.array().items(appSchema),
   },
 };
 
+export const getMostUsed = {
+  response: {
+    200: S.array().items(mostUsedAppSchema),
+  },
+};
 /**
  * Fluent schema definitions to extend core schemas
  */
--- a/src/services/item/plugins/app/fluent-schema.ts
+++ b/src/services/item/plugins/app/fluent-schema.ts
@@ -12,7 +12,7 @@ export const mostUsedAppSchema = S.object()
   .additionalProperties(false)
   .prop('name', S.string())
   .prop('url', S.string())
-  .prop('nbr', S.number());
+  .prop('count', S.number());
 
 export const getMany = {
   response: {
--- a/src/services/item/plugins/app/index.ts
+++ b/src/services/item/plugins/app/index.ts
@@ -13,7 +13,7 @@ import appDataPlugin from './appData';
 import appSettingPlugin from './appSetting';
 import chatBotPlugin from './chatBot';
 import { DEFAULT_JWT_EXPIRATION } from './constants';
-import { createSchema, getMany, updateSchema } from './fluent-schema';
+import { createSchema, getMany, getMostUsed, updateSchema } from './fluent-schema';
 import common, { generateToken, getContext } from './schemas';
 import { AppService } from './service';
 import { AppsPluginOptions } from './types';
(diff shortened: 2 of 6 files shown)


src/services/item/plugins/app/repository.ts
```diff
@@ -12,6 +12,18 @@ export const AppRepository = AppDataSource.getRepository(App).extend({
     return this.findBy({ publisher: { id: publisherId } });
   },
 
+  async getMostUsedApps(memberId: string) {
+    return this.createQueryBuilder('app')
+      .leftJoin('item', "item.extra::json->>'app'->>'url' = app.url")
+      .select('item.creator_id', memberId)
+      .addselect('app.url', 'url')
```
spaenleh, line 19:
```suggestion
      .addSelect('app.url', 'url')
```

src/services/item/plugins/app/repository.ts
```diff
@@ -12,6 +12,18 @@ export const AppRepository = AppDataSource.getRepository(App).extend({
     return this.findBy({ publisher: { id: publisherId } });
   },
 
+  async getMostUsedApps(memberId: string) {
```
spaenleh, line 15:
```suggestion
  async getMostUsedApps(memberId: string): Promise<{url: string; name: string; nbr: number}[]> {
```

src/services/item/plugins/app/repository.ts
```diff
@@ -12,6 +12,18 @@ export const AppRepository = AppDataSource.getRepository(App).extend({
...
+    return this.createQueryBuilder('app')
+      .leftJoin('item', "item.extra::json->>'app'->>'url' = app.url")
+      .select('item.creator_id', memberId)
+      .addselect('app.url', 'url')
+      .addSelect('app.name', 'name')
+      .addSelect('COUNT(item.id)', 'nbr')
+      .groupBy('app.url')
+      .addGroupBy('app.name')
+      .orderBy('nbr', 'DESC');
```
spaenleh, line 24: Does this work ? Maybe we need to add `.getMany()` or `.getRawMany()`.

src/services/item/plugins/app/index.ts
```diff
@@ -117,6 +117,15 @@ const plugin: FastifyPluginAsync<AppsPluginOptions> = async (fastify, options) =
         return aS.getAllApps(member, buildRepositories(), publisherId);
       });
 
+      fastify.get(
+        '/most-used',
+        { schema: getMany },
+        async ({ member, authTokenSubject: requestDetails }) => {
+          const memberId = member ? member.id : requestDetails?.memberId;
+          return aS.getMostUsedApps(member, buildRepositories(), memberId as string);
```
spaenleh, line 125: I don't think the casting should be done here, instead check for `memberId` and if it is not available throw an error as this endpoint does not work if users are not authenticated.

src/services/item/plugins/app/service.ts
```diff
@@ -32,6 +32,10 @@ export class AppService {
     return repositories.appRepository.getAll(publisherId);
   }
 
+  async getMostUsedApps(actor: Actor, repositories: Repositories, memberId: string) {
```
spaenleh, line 35: You do not need to have a `memberId` parameter since you can get the current user of the request with the `actor` parameter. You can also change the type from `Actor` to `Member` since you know it should be defined.

spaenleh (commented):
Looks like a good start, did you test it ? Does it work ?

src/services/item/plugins/app/fluent-schema.ts
```diff
@@ -8,12 +8,23 @@ export const appSchema = S.object()
   .prop('url', S.string())
   .prop('extra', S.object().additionalProperties(true));
 
+export const mostUsedAppSchema = S.object()
+  .additionalProperties(false)
+  .prop('name', S.string())
+  .prop('url', S.string())
+  .prop('nbr', S.number());
```
pyphilia, line 15: As I suggested in builder:
```suggestion
  .prop('count', S.number());
```

src/services/item/plugins/app/index.ts
```diff
@@ -117,6 +117,17 @@ const plugin: FastifyPluginAsync<AppsPluginOptions> = async (fastify, options) =
         return aS.getAllApps(member, buildRepositories(), publisherId);
       });
 
+      fastify.get(
+        '/most-used',
+        { schema: getMostUsed, preHandler: fastify.attemptVerifyAuthentication },
+        async ({ member }) => {
+          if (!member) {
+            throw new Error('This user is not authenticated.');
```
pyphilia, line 125:
```suggestion
           throw new UnauthorizedMember(member);
```

src/services/item/plugins/app/index.ts
```diff
@@ -117,6 +117,17 @@ const plugin: FastifyPluginAsync<AppsPluginOptions> = async (fastify, options) =
         return aS.getAllApps(member, buildRepositories(), publisherId);
       });
 
+      fastify.get(
+        '/most-used',
+        { schema: getMostUsed, preHandler: fastify.attemptVerifyAuthentication },
```
pyphilia, line 122:
```suggestion
        { schema: getMostUsed, preHandler: fastify.verifyAuthentication },
```

pyphilia (changes requested):
LGTM, could you write tests at least for the controller? Thanks! 🏋️

PR #999, opened 2024-04-29
fix: set config.ts file import as 1st priority

Defferrard:


src/server.ts
```diff
@@ -1,10 +1,12 @@
+// sort-imports-ignore
```
spaenleh, line 1: Maybe simply add a comment explaining that in order for dotenv to initialise the variables they need to be imported first before anything else.

spaenleh (approved):
Looks good, thank you for the update, just a small comment and you can merge ! 🚀

PR #1004, opened 2024-05-01
fix: update tests to use fastify instance type for the app var

spaenleh:
In this PR I update the type setting in typescript to allow tests to benefit from the `decorators.d.ts` file. 
I had to fix a few tests.

Mainly: 
- query params should always be passed as strings

### DIFF
--- a/src/services/item/plugins/geolocation/index.test.ts
+++ b/src/services/item/plugins/geolocation/index.test.ts
@@ -2,6 +2,8 @@ import { StatusCodes } from 'http-status-codes';
 import nock from 'nock';
 import { v4 } from 'uuid';
 
+import { FastifyInstance } from 'fastify';
+
 import { HttpMethod } from '@graasp/sdk';
 
 import build, { clearDatabase } from '../../../../../test/app';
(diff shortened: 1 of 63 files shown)


src/services/item/plugins/geolocation/index.test.ts
```diff
@@ -493,8 +495,8 @@ describe('Item Geolocation', () => {
           method: HttpMethod.Get,
           url: `${ITEMS_ROUTE_PREFIX}/geolocation/reverse`,
           query: {
-            lat: 2,
-            lng: 2,
+            lat: '2',
+            lng: '2',
```
pyphilia, line 499: You cannot use number here? Maybe it becomes worth to use json stringify... or build the url manually, not sure.

pyphilia (approved):
Thank you typescript master! 💯  
I just have a concern about number->string. Were the schema wrong?

src/services/item/plugins/geolocation/index.test.ts
spaenleh: @pyphilia This has nothing to do with our schema. This is because the app.inject options takes only strings as inputs for the query parameters. They will be validated and parsed and when needed cast to numbers as usual when received. So I think it is fine.

ReidyT (approved):
Nice 🤩

PR #1008, opened 2024-05-01
fix(build): use alpine version of node docker image

Defferrard:


.dockerignore
```diff
@@ -3,8 +3,11 @@ coverage
 dist
 docs
 node_modules
-test
+**/test/**
```
spaenleh, line 6: Does this also include `fixtures` folders ?

docker/Dockerfile
```diff
@@ -1,49 +1,56 @@
...
+        build-base
+
+COPY ./package.json ./
+
+# Install the application
+RUN yarn install --production --ignore-scripts
+COPY --from=builder ./app/dist ./dist
+
+USER node
+CMD yarn start
```
spaenleh, line 56: Is it better practice to do `yarn start` than to use `node` on the built server.js ?

spaenleh (changes requested):
Nice work ! 🥡 🐳 

I have a small question about the command used to run the app.

docker/Dockerfile
spaenleh: Ah I see that `yarn start` is executing `yarn serve` which does exactly `NODE_PATH=./dist node dist/server.js`

.dockerignore
Defferrard: Yes it does ! Should we keep theses folders in production ? It doesn't seems to be used in the application to me

spaenleh: Okay then it is all good ! No they should not be used in production.

spaenleh (approved):

Defferrard:
`bcrypt` use `node-pre-gyp` to build the dependency. So we need to omit "--ignore-scripts" and add a condition for our `postinstall` script. Otherwise we can prebuild `bcrypt` somewhere else, but it means that we need to do the same for every packages that needs to be prebuild.

PR #1009, opened 2024-05-01
feat: add route to extract link's metadata

ReidyT:


src/services/item/plugins/embeddedLink/service.ts
```diff
@@ -0,0 +1,78 @@
...
+      description: description?.trim(),
+      html,
+      thumbnails: links.filter(({ rel }) => hasThumbnailRel(rel)).map(({ href }) => href),
+      icons: links.filter(({ rel }: { rel: string[] }) => hasIconRel(rel)).map(({ href }) => href),
+    };
+  }
+
+  async checkEmbeddingAllowed(url: string, logger?: FastifyBaseLogger): Promise<boolean> {
+    try {
+      const { headers } = await fetch(url);
```
spaenleh, line 46: Does it work if you fetch with a `HEAD` request instead ? This would reduce the bandwidth of the page.

ReidyT: I tried but it doesn't seems to work with the HEAD request.

spaenleh: okay !

src/services/item/plugins/embeddedLink/index.ts
```diff
@@ -26,6 +29,23 @@ const plugin: FastifyPluginAsync<GraaspEmbeddedLinkItemOptions> = async (fastify
   // "install" custom schema for validating embedded link items creation
   extendCreateSchema(createSchema);
 
+  // TODO: should we protect this route ?
```
spaenleh, line 32: I think it should be accessible only to authenticated members. As it will be used only for creating content. So members will need to be authenticated to create.

spaenleh (approved):
Looks good !

PR #1021, opened 2024-05-08
feat: reset password

Defferrard:


Defferrard:
It seems that I have some issues with environnement variables.

I've added 2 new environnement variables, one of them is mandatory to make the application works. It is a JWT secret. We have to define it somewhere. I've tried to push an edit of `test.yml`but I've been refused...

src/services/auth/plugins/password/index.test.ts
```diff
@@ -0,0 +1,315 @@
...
+      });
+      expect(response.statusCode).toBe(StatusCodes.NO_CONTENT);
+
+      // Wait for the mail to be sent
+      await new Promise((resolve) => setTimeout(resolve, 100));
+
+      expect(mockSendEmail).toHaveBeenCalledTimes(1);
+      expect(mockSendEmail.mock.calls[0][1]).toBe(entities[0].email);
+    });
+    it('Create a password request to an non-existing email', async () => {
```
spaenleh, line 119:
```suggestion
    it('Create a password request to a non-existing email', async () => {
```

src/services/auth/plugins/password/index.ts
```diff
@@ -1,20 +1,51 @@
...
+    PASSPORT_STATEGY_ID,
+    new Strategy(
+      {
+        jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
+        secretOrKey: PASSWORD_RESET_JWT_SECRET,
+        passReqToCallback: true,
+      },
+      async (req, _payload, done) => {
+        const token = ExtractJwt.fromAuthHeaderAsBearerToken()(req);
+        if (token && (await memberPasswordService.validateResetPasswordToken(token))) {
```
spaenleh, line 39: Do we maybe want to perform this before checking to improve readability ? Also add some comments about what it means to return `done(..., false)` or `done(..., true)`

src/services/auth/plugins/password/index.ts
```diff
@@ -58,6 +89,83 @@ const plugin: FastifyPluginAsync = async (fastify) => {
       });
     },
   );
+
+  /**
+   * Create a reset password request.
+   * This will send an email to the member in his langage with a link to reset the password.
+   * The link target a frontend route endpoint.
```
spaenleh, line 96:
```suggestion
   * The link targets a frontend route endpoint.
```

src/services/auth/plugins/password/index.ts
```diff
@@ -58,6 +89,83 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+   * @returns 204 No Content if the request was successful.
+   */
+  fastify.patch<{ Body: { password: string } }>(
+    '/reset-password-request',
+    {
+      schema: patchResetPasswordRequest,
+      preValidation: fastifyPassport.authenticate(PASSPORT_STATEGY_ID, { session: false }), // Session is not required.
+    },
+    async (request, reply) => {
+      const token: string = ExtractJwt.fromAuthHeaderAsBearerToken()(request)!;
```
spaenleh, line 152: It looks like you do this already in the callback of the passport strategy, is there a reason why you do it again here ? is it not safe/possible to pass information needed down to the request handler function from the passport strategy callback ?

src/services/auth/plugins/password/service.ts
```diff
@@ -17,9 +33,19 @@ const promisifiedJwtSign = promisify<
...
+  redis: Redis;
 
-  constructor(log) {
+  constructor(mailer, log) {
+    this.mailer = mailer;
     this.log = log;
+
+    this.redis = new Redis({
+      host: REDIS_HOST,
+      port: parseInt(REDIS_PORT ?? '6379'),
```
spaenleh, line 45: Maybe the default port could be set as a constant in some file ?

src/services/auth/plugins/password/service.ts
```diff
@@ -40,6 +66,104 @@ export class MemberPasswordService {
...
+   * Modify the password of a member. Force the change without checking the current password.
+   * Check if the token is registered in the redis database. If it is, delete it and change the password of the associated member.
+   * @param repositories Object with the repositories needed to interact with the database. Must contain a memberPasswordRepository.
+   * @param password New password.
+   * @param token The JSON Web Token associated to the member that wants to reset the password.
+   * @returns void
+   */
+  async forcePatch(repositories: Repositories, password: string, token: string): Promise<void> {
+    const id = await this.redis.get(`${REDIS_PREFIX}${token}`);
+    if (!id) return;
```
spaenleh, line 79: Avoid inline conditionals, use braces even for one-line statements. 
```suggestion
    if (!id) {
      return;
    }
```

src/services/auth/plugins/password/service.ts
```diff
@@ -40,6 +66,104 @@ export class MemberPasswordService {
...
+   */
+  async createResetPasswordRequest(
+    repositories: Repositories,
+    email: string,
+  ): Promise<{ token: string; member: Member } | undefined> {
+    const { memberRepository } = repositories;
+    const member: Member = await memberRepository.getByEmail(email);
+
+    if (!member) return;
+    if (!(await MemberPasswordRepository.getForMemberId(member.id))) return;
```
spaenleh, line 103: Same as above, always use braces for one-line conditionals. This improves clarity and code refactoring.

src/services/auth/plugins/password/service.ts
```diff
@@ -40,6 +66,104 @@ export class MemberPasswordService {
...
+      `${REDIS_PREFIX}${token}`,
+      PASSWORD_RESET_JWT_EXPIRATION_IN_MINUTES * 60,
+      member.id,
+    );
+    return { token, member };
+  }
+
+  /**
+   * Send an email to the member with a link to reset their password.
+   * The link target to a frontend endpoint that will handle the password reset.
```
spaenleh, line 116:
```suggestion
   * The link targets a frontend endpoint that will handle the password reset.
```

src/services/auth/plugins/password/service.ts
```diff
@@ -40,6 +66,104 @@ export class MemberPasswordService {
...
+   * @param email The email of destination
+   * @param token The JSON Web Token to reset the password.
+   * @param memberLang The language to use for the email.
```
spaenleh, line 119:
```suggestion
   * @param lang The language to use for the email.
```

spaenleh (changes requested):
Great work ! Let me know if you want to discuss some comments. Let's define what we need to implement in the frontend !

Is it correct that any session the user has currently set before requesting a password reset will not be terminated by reseting the password ? This means that, if someone acquired a session before the reset there is no way to "quick them out" ? Maybe this is something we would like to tackle in another issue ?

src/services/auth/plugins/password/index.ts
```diff
@@ -58,6 +89,83 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+  /**
+   * Solve the reset password request.
+   * This will force the password to be updated.
+   * A special token is required to perform this action. This token is sent by email to the member after creating a reset password request.
+   * If the password is not strong enough, the request will fail with an error 400 Bad Request.
+   * @param password - New password.
+   * @returns 204 No Content if the request was successful.
+   */
+  fastify.patch<{ Body: { password: string } }>(
+    '/reset-password-request',
```
pyphilia, line 146: Is there a reason for having the same url? Otherwise I would suggest `/password/reset/apply` (not ideal suggestion)

src/services/auth/plugins/password/schemas.ts
```diff
@@ -43,3 +47,37 @@ export const updatePassword = {
     },
   },
 };
+
+export const postResetPasswordRequest: FastifySchema = {
+  body: {
+    type: 'object',
+    properties: {
+      email: { type: 'string' },
```
pyphilia, line 55: I think you can precise the format.

src/services/auth/plugins/password/service.ts
```diff
@@ -17,9 +33,19 @@ const promisifiedJwtSign = promisify<
...
   log: FastifyBaseLogger;
+  mailer: MailerDecoration;
+  redis: Redis;
 
-  constructor(log) {
+  constructor(mailer, log) {
+    this.mailer = mailer;
     this.log = log;
+
+    this.redis = new Redis({
```
pyphilia, line 43: Should we define this higher in the code? In decorators?

src/services/auth/plugins/password/service.ts
```diff
@@ -40,6 +66,104 @@ export class MemberPasswordService {
...
 
+  /**
+   * Modify the password of a member. Force the change without checking the current password.
+   * Check if the token is registered in the redis database. If it is, delete it and change the password of the associated member.
+   * @param repositories Object with the repositories needed to interact with the database. Must contain a memberPasswordRepository.
+   * @param password New password.
+   * @param token The JSON Web Token associated to the member that wants to reset the password.
+   * @returns void
+   */
+  async forcePatch(repositories: Repositories, password: string, token: string): Promise<void> {
```
pyphilia, line 77: Does this need to be called `forcePatch`? Could this be `reset`?  I understand the need for `force`, but it does not explain really what it does.

src/services/auth/plugins/password/service.ts
```diff
@@ -40,6 +66,104 @@ export class MemberPasswordService {
...
+
+  /**
+   * Get the member associated to the JSON Web Token registered in the redis database.
+   * @param repositories Object with the repositories needed to interact with the database. Must contain a memberRepository.
+   * @param token The JSON Web Token to identify the member.
+   * @returns The member associated to the token. Otherwise, undefined if we couldn't find the member.
+   */
+  async getMemberByToken(repositories: Repositories, token: string): Promise<Member | undefined> {
+    const id = await this.redis.get(`${REDIS_PREFIX}${token}`);
+    if (!id) return;
```
pyphilia, line 161:
```suggestion
    if (!id) {
     return;
    }
```

src/services/auth/plugins/password/spec.md
```diff
@@ -0,0 +1,71 @@
...
+
+                service-)+repoPass: patch(member.id, password)
+            end
+        service--)-ctrl: void
+    ctrl--)-usr: 204 No Content
+```
+
+## Question
+
+- Actions
```
pyphilia, line 71: Still a question? Maybe write a clear question aha

pyphilia (changes requested):
Thanks for the nice diagrams! I'd like some texts as well though, it doesn't have to be as detailed, but maybe a rough description of the workflow, and what happens if the user does not have a password.

spaenleh (approved):
👍

Defferrard:
> Great work ! Let me know if you want to discuss some comments. Let's define what we need to implement in the frontend !
> 
> Is it correct that any session the user has currently set before requesting a password reset will not be terminated by reseting the password ? This means that, if someone acquired a session before the reset there is no way to "quick them out" ? Maybe this is something we would like to tackle in another issue ?

Currently, if the Frontend Application wants to implement the Reset Password Feature, it needs to :
1. Have a form to create a `reset password request`. It needs
   - An email Input
   - A Captcha `RecaptchaAction.ResetPassword`
   - A button that'll send a `POST` request to `{{graasp-backend}}/password/reset` with a body :
```json
{
    "email":"{{user_email}}",
    "captcha":"{{captcha}}"
}
```
   - The API will reply with
      - 204 : The request has been handled successfully 
      - 401 : The captcha has not been validated
2. Have a page `{{AUTH_CLIENT_HOST}}/reset-password?t=<token>`, that contains  a form with
   - A password input (With `confirm password` validation input ? )
   - A button that'll send a `PATCH` request to `{{graasp-backend}}/password/reset` with
      - An authorization header : `Bearer {{token}}`
      - A body :
```json
{
   "password":"{{password}}"
}
```
   - The API will reply with
      - 204 : The request has been handled successfully 
      - 401 : The token is invalid, has not been specified or is expired.

src/services/auth/plugins/password/index.ts
spaenleh: @pyphilia I think it is ok to use the same route and simply change the Http verbe.

src/services/auth/plugins/password/index.test.ts
```diff
@@ -0,0 +1,314 @@
...
+  });
+
+  beforeEach(async () => {
+    jest.clearAllMocks();
+
+    // Seed the database with members and passwords
+    entities = [
+      {
+        id: faker.string.uuid(),
+        pass: faker.internet.password({ prefix: '!1Aa' }),
```
pyphilia, line 68:
```suggestion
        password: faker.internet.password({ prefix: '!1Aa' }),
```

pyphilia (approved):

Issue #998, opened 2024-04-29
env variables are used before initialised by dotenv


Defferrard:
The environnement variables are used on many place (b.e : PostgreSQL Database Connection), but they're not initialised by dotenv because of the import order.

Defferrard:
This issue has been fixed in a bad way.

We should use `config.ts`variables instead of `process.env` in database configuration context. We shouldn't use `process.env` without a reason.

PR #1023, opened 2024-05-13
998 env variables are used before initialised by dotenv

Defferrard:
The first implementation of #998 didn't solved the issue. Because environnement variables where still used before initialisation during test context.

I've moved every `process.env`declarations in `config.ts` file, so every time we need to fetch an environnement variable, we have to import the configuration file.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,13 +11,56 @@ on:
 concurrency:
   group: ${{ github.head_ref || github.ref }}
 
+env:
+  CI: true
+  LOG_LEVEL: fatal # avoid unnecessary logging
+  # random keys
+  APPS_JWT_SECRET: ${{ secrets.MOCK_JWT_SECRET }}
+  APPS_PUBLISHER_ID: 9c9cea73-f3b7-48a3-aa6e-ead82c0685e7 # mock uuid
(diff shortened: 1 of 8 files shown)


spaenleh (approved):
LGTM ! thank you !

spaenleh:
@Defferrard You seem to have an error with an undefined `COOKIE_DOMAIN` env.

Defferrard:
> @Defferrard You seem to have an error with an undefined `COOKIE_DOMAIN` env.

It seems like the `build-node` only define the needed environnement variable to run the `src/plugins/datasource.ts` file. Which originaly use `process.env`to fetch the environnement variable.

Now with my modification, it imports `config.ts`, and needs to define **all** environnement variable needed to run the whole application.

Which means I can either :
1. **Set a default value to every constants in `config.ts``** I'd use development variable, so we wouldn't need a `.env.development` with 130 line. But it wouldn't throw an error anymore if we don't define an mandatory variable.
2. **Define all environnement variable in `build-node` job.** Which is already done in `test-node` job.

Any preference, or suggestions @spaenleh , @pyphilia  ?

spaenleh (approved):
🚀

pyphilia (approved):
So cool, LGTM 💯

PR #1033, opened 2024-05-14
fix: creator serialisation issue

### LINKED ISSUE #1031: Creator is not serialized correctly because of incorrect return code in schema
A miss-match between the effective return codes and the return code specified in the schema makes the serialisation not work on item create.

The effective return code is 200 while the schema expected response code is 201, the schema is not used for the response.

spaenleh:
fix #1031


src/services/item/service.ts
```diff
@@ -322,6 +322,7 @@ export class ItemService {
       memberships.map(({ item }) => item),
       resultOfMembership,
     );
+    this.log.debug(packedItems);
```
pyphilia, line 325: Remove?

pyphilia (approved):
I guess we should also decide on what format we want to use, we still have this strange mix of fluent-schema and ajv..

PR #1053, opened 2024-05-21
fix: getIn geoloc returns public only inside item

pyphilia:
The previous implementation was returning public items of everyone when navigating the map on root.

I removed the public items when requesting on root, but they are returned when scoped in a parent item.


src/services/item/plugins/geolocation/index.test.ts
```diff
@@ -189,6 +189,43 @@ describe('Item Geolocation', () => {
...
+
+        const item1 = await testUtils.saveItem({ actor: member, parentItem });
+        const { packed: geoloc1 } = await saveGeolocation({
+          item: new ItemWrapper(item1, null, [publicTag]).packed(),
+          lat: 1,
+          lng: 2,
+          country: 'de',
+        });
+        const item2 = await testUtils.saveItem({ actor: member, parentItem });
+        console.log(item2);
```
ReidyT, line 209: Still wanted ? I know it's in the tests 😅

src/services/item/plugins/geolocation/service.ts
```diff
@@ -78,14 +78,17 @@ export class ItemGeolocationService {
...
     return geoloc
       .map((g) => {
-        if (g.item.id in itemMemberships.data) {
+        const itemId = g.item.id;
+        if (
+          // accessible items
+          itemId in itemMemberships.data &&
+          // accept public items within parent item
+          // otherwise the actor should have at least read permission on root
+          (query.parentItemId || itemMemberships.data[g.item.id])
```
ReidyT, line 87: Is `itemMemberships.data[g.item.id]` different to `itemId in itemMemberships.data` ? I don't think so, you may have to merge the two conditions ?

If not, you could reuse the `itemId`:

```suggestion
          (query.parentItemId || itemMemberships.data[itemId])
```

src/services/item/plugins/geolocation/service.ts
```diff
@@ -78,14 +78,17 @@ export class ItemGeolocationService {
     // TODO optimize?
     return geoloc
       .map((g) => {
-        if (g.item.id in itemMemberships.data) {
+        const itemId = g.item.id;
+        if (
+          // accessible items
```
ReidyT, line 83: Maybe extracting the conditions in well named variables allow you to avoid comments ?

`const itemIsAccessible = itemId in itemMemberships.data;` or a better name ;)

src/services/item/plugins/geolocation/service.ts
```diff
@@ -78,14 +78,17 @@ export class ItemGeolocationService {
...
+          (query.parentItemId || itemMemberships.data[g.item.id])
+        ) {
           return {
             ...g,
-            item: new ItemWrapper(
-              g.item,
-              itemMemberships.data[g.item.id],
-              tags.data[g.item.id],
-            ).packed(),
+            item: new ItemWrapper(g.item, itemMemberships.data[itemId], tags.data[itemId]).packed(),
```
ReidyT, line 91: Maybe extracting the new item in a const at the begining of the if's body and using it here could increase the readability ? 

```suggestion
            item: newItem.packed(), // I think my example doesn't use a good name ;)
```

ReidyT (approved):
LGTM, I juste let some comments 😆

src/services/item/plugins/geolocation/service.ts
pyphilia: The difference is that `itemId in itemMemberships.data` is `true` even if `itemMemberships.data[itemId]` is `null`.  This happens if the item is public. I added a comment because it's not that clear indeed!

spaenleh (approved):

PR #1058, opened 2024-05-22
fix: allow to update link url and disallow editing html

### LINKED ISSUE #1057: Update item link creation flow to fit new modal and infos
Todo:
- [x] provided item description should take precedence over injected iframely description
- [x] html integration code should be added by the backend and stripped from the post payload if provided
- [ ] (optional) thumbnails should be generated from the icons to remove the need to ask CDNs and leak user informations

spaenleh:
In this PR:
- move automatic link description to the `extra.description` field
- allow to update a link `url`
- do not allow to update more that the `extra.url` field in updates to prevent injecting into the unsafe `extra.html` prop

fix #1057

### DIFF
--- a/src/services/item/plugins/embeddedLink/schemas.ts
+++ b/src/services/item/plugins/embeddedLink/schemas.ts
@@ -2,10 +2,13 @@ import S from 'fluent-json-schema';
 
 import { ItemType } from '@graasp/sdk';
 
+import { error } from '../../../../schemas/fluent-schema';
+
 const embeddedLinkItemExtraCreate = S.object()
   .additionalProperties(false)
   .prop(
     ItemType.LINK,
+    // only allow to set url in extra on create
     S.object().additionalProperties(false).prop('url', S.string().format('url')).required(['url']),
   )
   .required([ItemType.LINK]);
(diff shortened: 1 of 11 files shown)


src/services/item/plugins/embeddedLink/schemas.ts
```diff
@@ -15,25 +18,27 @@ export const createSchema = S.object()
...
-//       required: ['embeddedLink']
-//     }
-//   }
-// }
+// schema defining the allowed properties for the link item
+export const updateExtraSchema = S.object()
+  .additionalProperties(false)
+  .prop(
+    ItemType.LINK,
+    // only allow to set url in extra on create
```
ReidyT, line 26: I'm not sure to understand where this constraint is reflected in this schema. It is a description of the next line or an indication that we should allow to set the url in extra on create only ? I think it is the same for the createExtraSchema.

src/services/item/plugins/embeddedLink/schemas.ts
```diff
@@ -15,25 +18,27 @@ export const createSchema = S.object()
...
+    S.object().additionalProperties(false).prop('url', S.string().format('url')).required(['url']),
+  )
+  .required([ItemType.LINK]);
+
+export const getLinkMetadata = {
+  querystring: S.object().additionalProperties(false).prop('link', S.string().format('url')),
```
ReidyT, line 32: Could it be interesting to extract the object schema in a const and reuse it in the three schema ?

src/services/item/plugins/embeddedLink/test/embeddedLink.test.ts
```diff
@@ -189,7 +190,7 @@ describe('Link Item tests', () => {
         ({ app, actor } = await build());
       });
 
-      it('Bad Request for link', async () => {
+      it('Allow to edit link url', async () => {
```
ReidyT, line 193: So we can update the url ? It is not the opposite of what the comments said ? 😅

ReidyT (approved):
LGTM 😄 I just let some comments because there is something not clear to me 😅

spaenleh:
@ReidyT Thank you for your review, I tried to make the comments clearer. My intent is to allow only the `extra.url` on creation and update. any other properties are not allowed in the extra payload.

ReidyT:
@spaenleh thanks for this update 👍 ! it is clearer for me now 😄

src/services/item/plugins/embeddedLink/test/embeddedLink.test.ts
```diff
@@ -215,6 +216,34 @@ describe('Link Item tests', () => {
           payload,
         });
 
+        expect(response.statusMessage).toEqual(ReasonPhrases.OK);
+        expect(response.statusCode).toBe(StatusCodes.OK);
+        // TODO: check that the title and description have been updated for the new link
```
pyphilia, line 221: Todo?

src/services/item/plugins/embeddedLink/test/embeddedLink.test.ts
```diff
@@ -274,8 +303,9 @@ describe('Link Item tests', () => {
           payload,
         });
 
-        expect(response.statusMessage).toEqual(ReasonPhrases.BAD_REQUEST);
-        expect(response.statusCode).toBe(StatusCodes.BAD_REQUEST);
+        expect(response.statusMessage).toEqual(ReasonPhrases.ACCEPTED);
+        expect(response.statusCode).toBe(StatusCodes.ACCEPTED);
+        // TODO: test that the link info has been updated
```
pyphilia, line 308: Todo

src/services/item/service.ts
```diff
@@ -434,6 +434,8 @@ export class ItemService {
     const item = await itemRepository.get(itemId);
     await validatePermission(repositories, PermissionLevel.Write, actor, item);
 
+    // TODO: if updating a link item, fetch the new informations
```
pyphilia, line 437: Todo

pyphilia (approved):
LTGM, you left some todos I've highlighted them 🥟

src/services/item/plugins/embeddedLink/test/embeddedLink.test.ts
spaenleh: referenced in #1066

src/services/item/service.ts
spaenleh: referenced in #1066

PR #1061, opened 2024-05-23
fix: use yarn4 in dockerfile

Defferrard:


Defferrard:
There is an issue on AWS

```
23 May 2024 at 09:23 (UTC+2:00) | Error: connect ECONNREFUSED 127.0.0.1:5432 | graasp
23 May 2024 at 09:23 (UTC+2:00) | at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1606:16) | graasp
23 May 2024 at 09:23 (UTC+2:00) | This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason: | graasp
```

pyphilia (approved):
💯

spaenleh (approved):
All Good ! Very good bug finding ! 🎉

PR #1063, opened 2024-05-23
feat: use full text search in accessible

pyphilia:
Added `keywords` to qs to allow searching broadly in the item.
`name` still work, the same way as `keywords` now, but is deprecated. (so no immediate changes needed from the frontend)


src/services/itemMembership/repository.ts
```diff
@@ -143,9 +145,29 @@ export const ItemMembershipRepository = AppDataSource.getRepository(ItemMembersh
         return 'item.path =' + subQuery.getQuery();
       });
 
-    if (name) {
-      query.andWhere("LOWER(item.name) LIKE '%' || :name || '%'", {
-        name: name.toLowerCase().trim(),
+    // todo: remove name
```
Defferrard, line 148: You already have deprecated name, I don't think it is necessary to add a todo here. It'll be naturally deleted.
So we don't get drowned by too much todo.

With this comment, we're risking someone to actually remove name here, without removing the whole deprecated feature.

Defferrard (commented):
I'm really not a big fan of theses query builder and string queries annotations (and which is not related to theses file changes : the big flat query string from line `597` to `617`)

It is a pain to understand what is happening and why we're doing it that way instead of using standard `typeorm` find options. The ItemMembershipRepository is 665 lines long, we should split the big files into multiple smaller files.

Theses query string annotations makes me feel inconfortable about security. I've tried some injections and it looks safe. But we should avoid them if possible. I don't see any good reason to use them to be honest...

The rest looks good to me !

spaenleh:
@Defferrard I agree, it makes me uneasy using unchecked queries like the long one you mention. Actually the git blame indicates this one is 4 years old. I think it is one of these "don't understand it, don't touch it, as long as it works". 

I am not sure to which annotations you are reffering, the `:name` ones ? I think for these they are passed as sql parameters and should thus be safe from injections.

pyphilia:
Indeed this big long query is legacy from long ago. Even when refactoring the code to typeorm I didn't have the confidence to fully translate it. To remove it, we should probably proceed by steps:
- move membership repository to typed repository
- add tests for each function of this repository with few changes to the functions themselves
- rewrite/improve some long queries that were not translated yet. We have confidence at this step with the previously written tests.

Mainly what I did for member some days ago.

Defferrard:
> @Defferrard I agree, it makes me uneasy using unchecked queries like the long one you mention. Actually the git blame indicates this one is 4 years old. I think it is one of these "don't understand it, don't touch it, as long as it works".
> 
> I am not sure to which annotations you are reffering, the `:name` ones ? I think for these they are passed as sql parameters and should thus be safe from injections.

@spaenleh Yes, I'm referring to the `:name` ones. They are safe from injections because they're using SQL parameters. But giving code instruction from string gives me insecurity feelings. I believe we could have a cleaner version by using options on TypeORM.

src/services/itemMembership/repository.ts
```diff
@@ -143,9 +145,29 @@ export const ItemMembershipRepository = AppDataSource.getRepository(ItemMembersh
...
+    const allKeywords = [name, ...keywords].filter((s) => s && s.length);
```
ReidyT, line 149: If you want to keep defined none empty string, this should work:

```suggestion
    const allKeywords = [name, ...keywords].filter((s) => Boolean(s));
```

But of course, your code is fine too !

ReidyT (approved):
Thanks for your PR ! I agree with the other comments, but in any case, we should do that in another issue.... Concerning this PR, it looks good to me 🙌🚀.

PR #1072, opened 2024-05-29
fix: copy suffix can't increment past 3

spaenleh:
Using the `global` flag on a regex makes it stateful, which is not what we want in our case. 

I added a test and fixed the issue by removing the global flag.


Defferrard (commented):
LGTM!

Defferrard (approved):

PR #1077, opened 2024-06-03
feat: publish automatically when item is validated

ReidyT:
Publish the item after the validation if it is valid.
If the item is private, it will be set to public after the validation succeeded (the frontend has to ask to the user).

### DIFF
--- a/src/services/item/plugins/published/service.ts
+++ b/src/services/item/plugins/published/service.ts
@@ -118,7 +118,7 @@ export class ItemPublishedService {
     actor: Actor,
     repositories: Repositories,
     itemId: string,
-    { shouldBePublic = true }: { shouldBePublic: boolean } = { shouldBePublic: true },
+    { canBePrivate }: { canBePrivate?: boolean } = {},
   ) {
     if (!actor) {
       throw new UnauthorizedMember(actor);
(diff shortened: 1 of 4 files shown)


src/services/item/plugins/published/service.ts
```diff
@@ -123,7 +128,16 @@ export class ItemPublishedService {
...
-    await itemTagRepository.getType(item, ItemTagType.Public, { shouldThrow: true });
+    const tag = await itemTagRepository.getType(item, ItemTagType.Public, {
+      shouldThrow: shouldBePublic,
+    });
+
+    // if the item can be private and be published, set it to public automatically.
+    // it's usefull for the publishing automatically after the validation.
+    // the user is asked to set the item to public in the frontend.
+    if (!tag && !shouldBePublic) {
+      await itemTagRepository.post(actor, item, ItemTagType.Public);
```
ReidyT, line 139: It is the responsibility of the publish service or the validation service to set the item to public if it is private ?

src/services/item/plugins/validation/service.ts
```diff
@@ -31,12 +32,19 @@ import { stripHtml } from './utils';
...
   itemService: ItemService;
+  private publishService: ItemPublishedService;
   fileService: FileService;
   imageClassifierApi?: string;
 
-  constructor(itemService: ItemService, fileService: FileService, imageClassifierApi?: string) {
+  constructor(
+    itemService: ItemService,
+    fileService: FileService,
+    publishService: ItemPublishedService,
```
ReidyT, line 42: For now I inject the dependency manually, I will check if it is a small change to use a DI library for that.

src/services/item/plugins/published/service.ts
```diff
@@ -114,7 +114,12 @@ export class ItemPublishedService {
     };
   }
 
-  async post(actor: Actor, repositories: Repositories, itemId: string) {
+  async post(
+    actor: Actor,
+    repositories: Repositories,
+    itemId: string,
+    { shouldBePublic = true }: { shouldBePublic: boolean } = { shouldBePublic: true },
```
spaenleh, line 121: Maybe handling the missing option inside the function will be less cumbersome ?

spaenleh (approved):
This looks good to me.

PR #1079, opened 2024-06-04
fix: check that array is empty before getting many tags

### LINKED ISSUE #1078: [BUG] Geolocation may request validate permission on empty item list causing join on full DB
### Context

A recent addition of a geolocation button to the player interface caused some severe perfrmance issues in the database.

After investigation, the queries slowing the DB down were related to querying the tags.

Tags are queried whenever we need to validate permissions as we need to check if an item is hidden or public.
A particularly heavy query is the descendants query which get a full tree of items sometimes returning 300+ elements.

During investigation, one itemTag query that was surprising contained a `WHERE 1=1` condition.
This condition was not written by a human but generated by TypeOrm when we itteraed over an empty list of items to add to the where condition.

We should check if the requested list of items for which we want the tags is not empty, in which case it does not make sens to query the tags, as it would join on the full item and tags tables returning a massive amount of data, that would anyway not be used. 

### Problem

The `validatePermissionMany` function is sometimes called on an emtpy list of items, causing a large useless join between item and tags.

### Solution

Check the size of the array of items requested to be checked for permissions and return early if array is empty.

spaenleh:
fix #1078

### DIFF
--- a/src/services/item/plugins/geolocation/service.ts
+++ b/src/services/item/plugins/geolocation/service.ts
@@ -74,7 +74,7 @@ export class ItemGeolocationService {
 
     // check if there are any items with a geolocation, if not return early
     const itemsWithGeoloc = geoloc.map(({ item }) => item);
-    if (itemsWithGeoloc.length === 0) {
+    if (!itemsWithGeoloc.length) {
       return null;
     }
 
--- a/src/services/item/plugins/geolocation/service.ts
+++ b/src/services/item/plugins/geolocation/service.ts
@@ -62,7 +62,7 @@ export class ItemGeolocationService {
       lng2?: ItemGeolocation['lng'];
       keywords?: string[];
     },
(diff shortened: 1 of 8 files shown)


src/services/item/plugins/geolocation/service.ts
```diff
@@ -71,6 +71,13 @@ export class ItemGeolocationService {
     }
 
     const geoloc = await itemGeolocationRepository.getItemsIn(actor, query, parentItem);
+
+    // check if there are any items with a geolocation, if not return early
+    const itemsWithGeoloc = geoloc.map(({ item }) => item);
+    if (itemsWithGeoloc.length === 0) {
```
pyphilia, line 77:
```suggestion
    if (!itemsWithGeoloc.length) {
```

pyphilia (approved):

pyphilia (approved):
LGTM

PR #1097, opened 2024-06-07
refactor: remove item hooks - in progress

pyphilia:
> [!WARNING]  
> This PR is here to show an idea, but won't be merged, because it's missing the dependencies injection.

Here is my idea to remove hooks related to items:
- structure the item services so that they extends a generic `itemService`. 
- then we can use an common implemented function depending on the item type (eg. copy)
- I haven't tried really hard for now, but how I did the switching of services given the type is to have a `ItemServiceManager`. Perhaps there's some way to link a string to a service. The other problem here is also that we often don't have the item immediately.  

Notes:
- this does not apply for websockets, or related entities such as tags, geolocation, etc.
- the current code does not compile, it is a proof of concept that a common function can be used. The functions themselves are not 100% correct.

```mermaid
classDiagram
    ItemService : copy()
    ItemService --|>  HTMLService
    ItemService --|>  FolderService
    FolderService : copy()
    ItemService --|>  AppService
    AppService : copy()
    HTMLService  --|>  H5PService
    H5PService : copy()
    
    ItemController : copy(req, reply) { \n service = getServiceForItemType() \n service.copy() \n  }
    ItemController --> ItemService
    class ItemServiceManager {
        static getServiceForItemType()
    }

    ItemServiceManager o-- H5PService
    ItemServiceManager o-- AppService
    ItemServiceManager o-- FolderService
    
```

### DIFF
--- a/src/services/item/classcontroller.ts
+++ b/src/services/item/classcontroller.ts
@@ -1,0 +1,46 @@
+import { StatusCodes } from 'http-status-codes';
+import { DataSource } from 'typeorm';
+
+import { buildRepositories } from '../../utils/repositories';
+import { ItemServiceManager } from './itemServiceManager';
+import { AppService } from './plugins/app/service';
+import { H5PService } from './plugins/html/h5p/service';
+
+export class ItemController {
+  // get from injection
+  private db: DataSource;
+  private h5pService: H5PService;
+  private appService: AppService;
+
+  async copy(request, reply) {
+    const {
+      member,
+      query: { id: ids },
+      body: { parentId },
+    } = request;
+
+    // todo: replace with authenticate
+    if (!member) {
+      throw 'not authenticated';
+    }
+
+    this.db.transaction(async (manager) => {
+      const repositories = buildRepositories(manager);
+
+      const results = await Promise.all(
+        ids.map(async (id) => {
+          const service = await ItemServiceManager.getServiceForTypeFromId(repositories, id);
+
+          // TODO: check permission here
+
+          return service.copy(member, repositories, id, { parentId });
+        }),
+      );
+      return { items: results.map(({ item }) => item), copies: results.map(({ copy }) => copy) };
+    });
+    // TODO: set back websocket
+
+    reply.status(StatusCodes.ACCEPTED);
+    return ids;
+  }
+}
--- a/src/services/item/folderService.ts
+++ b/src/services/item/folderService.ts
@@ -1,0 +1,32 @@
+import { UUID } from '@graasp/sdk';
+
+import { Repositories } from '../../utils/repositories';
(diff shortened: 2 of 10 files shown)


src/services/item/classcontroller.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (!member) {
+      throw 'not authenticated';
+    }
+
+    this.db.transaction(async (manager) => {
+      const repositories = buildRepositories(manager);
+
+      const results = await Promise.all(
+        ids.map(async (id) => {
+          const service = await ItemServiceManager.getServiceForTypeFromId(repositories, id);
```
ReidyT, line 32: What do you think if instead of asking one service per item, we ask in one request the services for the items ? this could improve the performances and avoid to do multiple requests when it is possible to do a single request.

src/services/item/folderService.ts
```diff
@@ -0,0 +1,32 @@
...
+import { ItemService } from './service';
+
+export class FolderService extends ItemService {
+  async copy(actor: Actor, repositories: Repositories, itemId: UUID, args: { parentId?: UUID }) {
+    const item = (await this.get(actor, repositories, itemId)) as FolderItem;
+
+    const descendants = await repositories.itemRepository.getDescendants(item);
+    const items = [...descendants, item];
+
+    const res = await super.copy(actor, repositories, itemId, args);
```
ReidyT, line 16: Maybe it is clearer to do the parent's copy first, and then defining the const concerning the subclass's copy logic.

src/services/item/classcontroller.ts
```diff
@@ -0,0 +1,46 @@
...
+
+          // TODO: check permission here
+
+          return service.copy(member, repositories, id, { parentId });
```
ReidyT, line 36: Also, maybe if it is the same service used by multiple items, we could try to do one request copy for multiple items. I don't know if it is possible and if it could improve the performances.

src/services/item/itemServiceManager.ts
```diff
@@ -0,0 +1,35 @@
...
+  private folderService: FolderService;
+
+  static async getServiceForTypeFromId(repositories, id: string): Promise<ItemService> {
+    // TODO: check permission
+
+    // not really cool
+    const item = await repositories.itemRepository.get(id);
+
+    // question: would be best to get the service given the item
+    // like item.getService() but not sure it's feasible
```
ReidyT, line 24: I'm wondering if it is the responsibility of the item to know which service to use, I'm not sure.

pyphilia:
Because of inheritance, it would be best to have a different service, maybe `ManyItemsService`?? that can `copyMany` , and use `FolderService`, `FileItemService`, etc

PR #1103, opened 2024-06-11
feat: filters item type and hidden item

### LINKED ISSUE #1080: Optimization: validatePermissionDescendants
This should be faster to check descendants rather than separated items.

Defferrard:
fix #1080

### DIFF
--- a/src/services/authorization.ts
+++ b/src/services/authorization.ts
@@ -190,9 +190,14 @@ export const validatePermission = async (
 /**
  * Internal filtering function that takes out limited items (eg. hidden children)
  *  */
-const _filterOutItems = async (actor: Actor, repositories: Repositories, items: Item[]) => {
+const _filterOutItems = async (
+  actor: Actor,
+  repositories: Repositories,
+  items: Item[],
+  options?: { showHidden?: boolean },
+) => {
   const { itemMembershipRepository } = repositories;
-
+  const showHidden = options?.showHidden ?? true;
   if (!items.length) {
     return { items: [], memberships: [] };
   }
@@ -210,6 +215,9 @@ const _filterOutItems = async (actor: Actor, repositories: Repositories, items: 
   ]);
   const filteredItems = items.filter((item) => {
     const isHidden = tags.data[item.id].find((t) => t.type === ItemTagType.Hidden);
+    if (isHidden && !showHidden) {
+      return false;
+    }
     const permission = PermissionLevelCompare.getHighest(
       memberships[item.id]?.map(({ permission }) => permission),
     );
@@ -240,12 +248,13 @@ export const filterOutPackedItems = async (
   actor: Actor,
   repositories,
   items: Item[],
+  options?: { showHidden?: boolean },
 ): Promise<PackedItem[]> => {
   const {
     items: filteredItems,
     memberships,
     tags,
-  } = await _filterOutItems(actor, repositories, items);
+  } = await _filterOutItems(actor, repositories, items, options);
   return filteredItems.map((item) => {
     const permission = PermissionLevelCompare.getHighest(
       memberships[item.id]?.map(({ permission }) => permission),
--- a/src/services/item/controller.test.ts
+++ b/src/services/item/controller.test.ts
@@ -1,0 +1,131 @@
+import { StatusCodes } from 'http-status-codes';
+import { v4 as uuid } from 'uuid';
+
+import { FastifyInstance } from 'fastify';
+
+import {
+  AppItemFactory,
+  FolderItemFactory,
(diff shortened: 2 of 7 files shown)


src/services/authorization.ts
```diff
@@ -190,9 +190,14 @@ export const validatePermission = async (
 /**
  * Internal filtering function that takes out limited items (eg. hidden children)
  *  */
-const _filterOutItems = async (actor: Actor, repositories: Repositories, items: Item[]) => {
+const _filterOutItems = async (
```
ReidyT, line 193: If this function should be private, we should enforce it.

```suggestion
private const filterOutItems = async (
```

src/services/item/controller.test.ts
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  describe('GET /:id/descendants filters', () => {
+    let uuids;
+    beforeEach(async () => {
+      // uuids[0] is the root
+      // uuids[1] is folder
+      // uuids[2] is hidden
+      // uuids[3] is public
+      uuids = [uuid(), uuid(), uuid(), uuid()];
```
ReidyT, line 43: What about using distinct constants or a map to increase readability and avoid the comments ?

```suggestion
      const rootUUID = uuid();
      const folderUUID = uuid();
      const hiddenUUID = uuid();
      const publicUUID = uuid();
```

src/services/item/service.ts
```diff
@@ -386,15 +386,20 @@ export class ItemService {
     return filterOutPackedItems(actor, repositories, children);
   }
 
-  async _getDescendants(actor: Actor, repositories: Repositories, itemId: UUID) {
+  async _getDescendants(
```
ReidyT, line 389:
```suggestion
  private async getDescendants(
```

ReidyT (approved):
Nice PR ! LGTM, I just let some recommandations 😄

src/services/authorization.ts
Defferrard: Because of this function is not part of a class, I won't be able to add the private clause. We can encapsulate the whole file into a static class, but it would need a lot of changes.

ReidyT: Ah I didn't saw correctly, my bad ^^. But maybe we don't need the underscore because the function is not exported ? What do you think ?

src/services/item/service.ts
Defferrard: I've set it to private, but since there is already a public function getDescendants, I can't rename it `getDescendants`. I think we can rename it into a more accurate name. Any suggestions ?

What about renaming the publics ones to getFilteredDescendants ?

src/services/authorization.ts
Defferrard: It looks like there's already a `filterOutItems` function  in the same file which basically call the private function. So we would have to rename something. I agree that the `_` could not be necessary. Any suggestions ?

What about `filterOutItemsByMembershipAndTags` ? It seems too verbose to me. Maybe the function is doing too much things ? What do you think @spaenleh & @pyphilia ?

src/services/item/repository.ts
```diff
@@ -208,16 +208,23 @@ export class ItemRepository {
...
-    const { ordered = true } = options ?? {};
+    const { ordered = true, types } = options ?? {};
 
     const query = this.repository
       .createQueryBuilder('item')
       .leftJoinAndSelect('item.creator', 'creator')
       .where('item.path <@ :path', { path: item.path })
       .andWhere('item.id != :id', { id: item.id });
 
+    if (types) {
```
spaenleh, line 224: Since `Boolean([]) == true` You should also check the length.

src/services/item/service.ts
spaenleh: Yes, we can rename them, or maybe rename the private ones ?

spaenleh (approved):
Looks good to me ! We should add these changes to the query-client to test it with the frontend.

src/services/item/service.ts
ReidyT: Yes I think it is a good idea to rename it `getDescendants` for the private one and `getFilteredDescendants` for the public one, because it is explain what the functions really do.

src/services/authorization.ts
Defferrard: As we discussed with Thibault this morning. Since we don't export the function we won't be able to use it outside the file. But I understand the use of `_` here, because even if the function is not exported in its declaration, it still can be exported at the end of the file.

If there's no better solution, I won't modify the name since it doesn't concern this PR. If needed we can open another PR to refactor the `authorization.ts` structure ?

PR #1106, opened 2024-06-11
feat: restrict publication items to folder only

ReidyT:
- feat: restrict publication items to folder
- fix: try to publish only if not already published
- feat: getForItem of publish repository doesn't throw anymore
- chore(dep): update SDK to 4.13.0

### DIFF
--- a/src/services/item/plugins/published/service.ts
+++ b/src/services/item/plugins/published/service.ts
@@ -1,6 +1,6 @@
 import { FastifyBaseLogger } from 'fastify';
 
-import { ItemTagType, PermissionLevel, UUID } from '@graasp/sdk';
+import { ItemTagType, PermissionLevel, PublishableItemTypeChecker, UUID } from '@graasp/sdk';
 import { DEFAULT_LANG } from '@graasp/translations';
 
 import type { MailerDecoration } from '../../../../plugins/mailer';
@@ -15,7 +15,7 @@ import { ItemWrapper } from '../../ItemWrapper';
 import { Item } from '../../entities/Item';
 import { ItemService } from '../../service';
 import { buildPublishedItemLink } from './constants';
-import { ItemPublishedNotFound } from './errors';
+import { ItemTypeNotAllowedToPublish } from './errors';
 
 interface ActionCount {
   actionCount: number;
@@ -77,6 +77,10 @@ export class ItemPublishedService {
     try {
       // get item published entry
       const publishedItem = await itemPublishedRepository.getForItem(item);
+
+      if (!publishedItem) {
+        return null;
+      }
       // get views from the actions table
       const totalViews = await actionRepository.getAggregationForItem(item.path, {
         view: 'library',
@@ -84,10 +88,6 @@ export class ItemPublishedService {
       });
       return { totalViews: (totalViews?.[0] as ActionCount)?.actionCount, ...publishedItem };
     } catch (err) {
-      // when the item is found but it is not published we simply return `null`
-      if (err instanceof ItemPublishedNotFound) {
-        return null;
-      }
       // if the error was not expecte we throw it back
       throw err;
     }
@@ -114,18 +114,31 @@ export class ItemPublishedService {
     };
   }
 
+  async publishIfNotExist(member: Member, repositories: Repositories, itemId: string) {
+    const { itemPublishedRepository } = repositories;
+
+    const item = await this.itemService.get(member, repositories, itemId, PermissionLevel.Admin);
+
+    const itemPublished = await itemPublishedRepository.getForItem(item);
+
+    if (itemPublished) {
+      return itemPublished;
+    }
+
+    return await this.post(member, repositories, item, { canBePrivate: true });
+  }
+
   async post(
-    actor: Actor,
+    member: Member,
     repositories: Repositories,
-    itemId: string,
+    item: Item,
     { canBePrivate }: { canBePrivate?: boolean } = {},
   ) {
-    if (!actor) {
-      throw new UnauthorizedMember(actor);
-    }
     const { itemPublishedRepository, itemTagRepository } = repositories;
 
-    const item = await this.itemService.get(actor, repositories, itemId, PermissionLevel.Admin);
+    if (!PublishableItemTypeChecker.isItemTypeAllowedToBePublished(item.type)) {
+      throw new ItemTypeNotAllowedToPublish(item.id, item.type);
+    }
 
     // item should be public first
     const tag = await itemTagRepository.getType(item, ItemTagType.Public, {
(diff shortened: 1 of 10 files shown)


src/services/item/plugins/published/service.ts
```diff
@@ -114,6 +115,36 @@ export class ItemPublishedService {
     };
   }
 
+  private async getPublishedIfExist(item: Item, itemPublishedRepository: ItemPublishedRepository) {
+    try {
+      return await itemPublishedRepository.getForItem(item);
```
pyphilia, line 120:
```suggestion
      return itemPublishedRepository.getForItem(item);
```

src/services/item/plugins/published/service.ts
```diff
@@ -114,6 +115,36 @@ export class ItemPublishedService {
...
+
+    const item = await this.itemService.get(actor, repositories, itemId, PermissionLevel.Admin);
+
+    const itemPublished = await this.getPublishedIfExist(item, itemPublishedRepository);
+
+    if (itemPublished) {
+      return itemPublished;
+    }
+
+    return await this.post(actor, repositories, itemId, { canBePrivate });
```
pyphilia, line 145:
```suggestion
    return this.post(actor, repositories, itemId, { canBePrivate });
```

src/services/item/plugins/published/test/index.test.ts
```diff
@@ -415,6 +415,25 @@ describe('Item Published', () => {
         expect(res.json()).toMatchObject(new MemberCannotAdminItem(expect.anything()));
       });
 
+      it('Cannot publish none folder item', async () => {
```
pyphilia, line 418:
```suggestion
      it('Cannot publish non-folder item', async () => {
```

src/services/item/plugins/published/service.ts
```diff
@@ -114,6 +115,36 @@ export class ItemPublishedService {
...
+    } catch (_e) {
+      return undefined;
+    }
+  }
+
+  async publishIfNotExist(
+    actor: Actor,
+    repositories: Repositories,
+    itemId: string,
+    { canBePrivate }: { canBePrivate?: boolean } = {},
```
pyphilia, line 130: What does that do? In which case it is ok to have the item being private?

src/utils/errors.ts
```diff
@@ -540,3 +540,20 @@ export class ShortLinkLimitExceed extends CoreError {
...
+        message: `The type "${itemType}" of item "${itemId}" is not allowed to be published. Only theses types are allowed "${ALLOWED_PUBLISHED_TYPES}".`,
+      },
+      {
+        itemId,
+        itemType,
+      },
+    );
+    this.origin = 'published';
+  }
+}
```
pyphilia, line 559: Actually there's a file `errors.ts` in published where you can save this new error.

src/services/item/plugins/published/service.ts
```diff
@@ -127,6 +158,10 @@ export class ItemPublishedService {
 
     const item = await this.itemService.get(actor, repositories, itemId, PermissionLevel.Admin);
 
+    if (itemTypeNotAllowedToBePublished(item.type)) {
```
pyphilia, line 161: I wonder whether it would be clearer to export an array from sdk and check if the type is included here. It's a bit "black boxed" to import a function.

pyphilia (changes requested):
Thanks for the changes! Let me know if you have questions^^

src/services/item/plugins/published/service.ts
```diff
@@ -114,6 +115,36 @@ export class ItemPublishedService {
...
+  }
+
+  async publishIfNotExist(
+    actor: Actor,
+    repositories: Repositories,
+    itemId: string,
+    { canBePrivate }: { canBePrivate?: boolean } = {},
+  ) {
+    if (!actor) {
+      throw new UnauthorizedMember(actor);
```
Defferrard, line 133: Can we move this check into to controller ? So instead of having an Actor Object as a parameter, we have a member

Defferrard (approved):
Well done, LGTM !

src/services/item/plugins/published/service.ts
ReidyT: good catch ! thanks

src/services/item/plugins/published/service.ts
ReidyT: Yes it's better and it was already the case ^^.

src/services/item/plugins/published/service.ts
ReidyT: I'm not sure to understand why I have to remove the `await` here. Especially in this case, if I don't wait on `getForItem`, I'll not catch the exception, am I wrong ?

src/services/item/plugins/published/service.ts
ReidyT: Here too, I'm not sure to understand why I should not await @pyphilia

src/services/item/plugins/published/service.ts
ReidyT: Yeah I copied the post method but I forgot to remove the canBePrivate params ^^

src/utils/errors.ts
ReidyT: Nice, thanks !

src/services/item/plugins/published/service.ts
ReidyT: I think it is preferable to use a function to voluntary hide the implementation to the service. In this way, the service delegate the check to another class (it is now a class, as @spaenleh suggested) and only ask if the given type is valid. It also allow us to reuse it in the frontend too. Finally, I don't think that its a problem to hide the valid types in the backend, if we really need to know, we can check in the SDK.

What do you think @pyphilia, @spaenleh, @Defferrard ? I'm open to argumentation ;)

src/services/item/plugins/validation/index.ts
```diff
@@ -92,7 +92,9 @@ const plugin: FastifyPluginAsync<GraaspPluginValidationOptions> = async (fastify
         if (hasValidationSucceeded) {
           // publish automatically the item if it is valid.
           // private item will be set to public automatically (should ask the user on the frontend).
-          await publishService.post(member, repositories, itemId, { canBePrivate: true });
+          await publishService.publishIfNotExist(member, repositories, itemId, {
```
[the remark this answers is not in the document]
ReidyT: Yes I though too ! Good idea, I just means that the publish controller have to get the item before calling the post, I don't think it's an issue. What do you think. @pyphilia  ?

src/services/item/plugins/published/service.ts
spaenleh: I am ok with this black box.

pyphilia (approved):

PR #1108, opened 2024-06-11
fix: many tags items length

### LINKED ISSUE #1107: Get many tags does not check array size

spaenleh:
There were instances where we fetch data when the items array is empty.

fix #1107

### DIFF
--- a/src/services/item/plugins/itemTag/repository.ts
+++ b/src/services/item/plugins/itemTag/repository.ts
@@ -7,7 +7,12 @@ import { Member } from '../../../member/entities/member';
 import { mapById } from '../../../utils';
 import { Item } from '../../entities/Item';
 import { ItemTag } from './ItemTag';
-import { CannotModifyParentTag, ConflictingTagsInTheHierarchy, ItemTagNotFound } from './errors';
+import {
+  CannotModifyParentTag,
+  ConflictingTagsInTheHierarchy,
+  InvalidUseOfItemTagsRepository,
+  ItemTagNotFound,
+} from './errors';
 
 /**
  * Database's first layer of abstraction for Item Tags and (exceptionally) for Tags (at the bottom)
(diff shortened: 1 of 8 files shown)


src/services/item/plugins/itemTag/repository.ts
```diff
@@ -225,6 +225,11 @@ export class ItemTagRepository {
    * @param  {Item[]} items
    */
   async getForManyItems(items: Item[], { withDeleted = false }: { withDeleted?: boolean } = {}) {
+    // should not query when items array is empty
+    if (!items.length) {
+      throw new Error('empty items list joins on full table');
+    }
+
```
ReidyT, line 232: It is more a general question, but should we throw in the Repository when something is not found (or here because the items are empty) or should we return null (when single value) or an empty array ? What do you think @pyphilia @Defferrard ?

ReidyT (approved):
Nice, LGTM ! I just let a more general question 😆

src/services/item/plugins/itemTag/repository.ts
pyphilia: I would have returned an empty result here?

pyphilia (changes requested):

src/services/item/plugins/itemTag/repository.ts
spaenleh: @pyphilia @ReidyT I want to throw an error, because I do not think that calling this function with an empty array will give the expected result. I would consider it a violation of the usage of the function. So given this I think it makes sens to throw instead of silently ignoring. I think if we silently ignore, it will make it harder to debug bad uses of this function, as it makes it harder to know iff you got nothing because there was nothing or because you supplied an empty array. 

I am however open to argumentation.

Defferrard: I fell like it is a multiplication of 0, which results to 0. I don't know if the comparaison is explicit, but I'd not expect the function to crash if I give an empty array.
Having an exception for this case means that we would probably have to add a check before each call of this function. If we want to respect explicit exceptions, on this situation you'd add a `@throw` comment on the function's JSDoc, on all other functions that calls it and also on the AJV Schema.
Moreover, I don't think we should define our interface by how we need it to be debugged.

If we decide to keep the exception here, we'd have to throw a more explicit error instead of `new Error(...)`

Defferrard (approved):
Nice PR ! LGTM !

pyphilia (approved):

Issue #604, opened 2023-09-12
[member] edit email


pyphilia:

Defferrard:
If the user has defined a password. He has to send it to the API to change his email.

```mermaid
sequenceDiagram
    participant oldmail as Old Mail Box
    participant newmail as New Mail Box
    actor usr as User
    participant g as Graasp
    
    usr ->> +g: Hey, I want to change my mail to <...> !
    g -) newmail: Click here to validate your new mail
    g -->> -usr: Check your mailbox to validate your mail

    usr ->> +newmail: Get validation link
    newmail -->> -usr: Validation link

    usr ->> +g: Validate my email
    g -) oldmail: You've changed your email !
    g -->> -usr: OK
```

PR #1113, opened 2024-06-14
feat: restrict members that didn't validate their email

### LINKED ISSUE #994: don't create member immediately on register
handle pending members that actually click on auth for the first time to create the member.
So we can delete pending members that never logged in.

Defferrard:
fix #994
> Note that by default all test are using validated members
# POST-PATCH-DELETE Routes **that have not been restricted**
- **PATCH /members/update-password**
- **PATCH /members/:id**
- **DELETE /members/:id**
- **DELETE /members/:id/delete**
- **POST /members/export-data**


Defferrard:
Wait for #604 before merging

Defferrard:
We have to wait for Frontend Implementation before merging

src/services/auth/error.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const GraaspAuthError = ErrorFactory(PLUGIN_NAME);
+
+export class InsufficientPermission extends GraaspAuthError {
+  constructor(data?: unknown) {
+    super(
+      {
+        code: 'GAUTHERR001',
+        statusCode: StatusCodes.FORBIDDEN,
+        message: 'User has insufficient permission to perform this action',
```
pyphilia, line 15: Translate

src/services/auth/error.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+}
+
+export class InvalidatedMember extends GraaspAuthError {
+  constructor(data?: unknown) {
+    super(
+      {
+        code: 'GAUTHERR001',
+        statusCode: StatusCodes.FORBIDDEN,
+        message: 'Member has not been validated by email',
```
pyphilia, line 28: Translate

pyphilia:
```suggestion
        message: 'Member has not validated its account by email',
```

src/services/auth/error.ts
```diff
@@ -0,0 +1,33 @@
...
+        code: 'GAUTHERR001',
+        statusCode: StatusCodes.FORBIDDEN,
+        message: 'User has insufficient permission to perform this action',
+      },
+      data,
+    );
+  }
+}
+
+export class InvalidatedMember extends GraaspAuthError {
```
pyphilia, line 22:
```suggestion
export class NotValidatedMember extends GraaspAuthError {
```
Pour moi même en français il y a une nuance entre "invalidé" il a été mécaniquement été invalidé (eg banni) alors que "pas validé" c'est qu'il n'a pas encore été validé.

src/services/auth/plugins/magicLink/index.ts
```diff
@@ -120,16 +116,27 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+        authInfo,
         query: { url },
         log,
       } = request;
+      const member = notUndefined(user?.member);
       const redirectionUrl = getRedirectionUrl(log, url ? decodeURIComponent(url) : undefined);
+      await db.transaction(async (manager) => {
+        const repositories = buildRepositories(manager);
+        memberService.refreshLastAuthenticatedAt(member.id, repositories);
+        if (authInfo?.emailValidation && !member.isValidated) {
```
pyphilia, line 136: Add comment here
```
on auth, if the user used the email sign in, its account gets validated
```

src/services/auth/plugins/mobile/index.ts
```diff
@@ -123,8 +124,15 @@ const plugin: FastifyPluginAsync = async (fastify) => {
       schema: mauth,
       preHandler: authenticateJWTChallengeVerifier,
     },
-    async ({ user }) => {
+    async ({ user, authInfo }) => {
       const member = notUndefined(user?.member);
+      await db.transaction(async (manager) => {
+        const repositories = buildRepositories(manager);
+        memberService.refreshLastAuthenticatedAt(member.id, repositories);
+        if (authInfo?.emailValidation && !member.isValidated) {
```
pyphilia, line 132: Same comment as above.

src/services/auth/plugins/roles/preHandlers.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export function whitelistRoles(...strategies: RoleStrategy[]): RouteHandlerMethod {
+  return async ({ user }: FastifyRequest<{ Body: { captcha: string } }>, _reply: FastifyReply) => {
+    if (!strategies.some((strategy) => strategy.test(user))) {
+      // If none of the strategies pass, throw an error.
+      if (strategies.length === 1 && strategies[0].error) {
+        throw new strategies[0].error();
+      } else {
+        throw new InsufficientPermission();
+      }
```
pyphilia, line 20: not sure of the logic here?

pyphilia (changes requested):
Thanks for this PR!

Reminder for frontend developers: ADD BANNER IN BUILDER TO FORCE USERS TO VALIDATE THEIR ACCOUNT !! INCLUDING MOBILE !!

Defferrard:
Apparently, log in by a magic link doesn't validate the member.

src/services/auth/plugins/roles/preHandlers.ts
Defferrard: I've added a comment `If only one strategy is provided, throw that error. Otherwise, throw a generic error.` I feel like the implementation is what we want (So we can throw a custom error). Tell me if you need more information !

src/services/auth/plugins/roles/preHandlers.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { InsufficientPermission } from '../../error';
+import { RoleStrategy } from './types';
+
+/**
+ * preHandler to check if user validate one of the preconditions to access a route.
+ * @param strategies One or more role strategies to check if user validate the preconditions.
+ * @throws {InsufficientPermission} If user does not validate any of the preconditions.
+ * @throws {GraaspAuthError} If only one role strategy is provided and it failed with a provided error.
+ */
```
ReidyT, line 11: It's really a suggestion, you can skip it if you want ;)

```suggestion
/**
 * Pre-handler function that checks if the user meets at least one of the specified access preconditions.
 * @param strategies The array of role strategies to check for access.
 * @throws {InsufficientPermission} If user does not satisfy any of the preconditions.
 * @throws {GraaspAuthError} If only one role strategy is provided and it failed with a provided error.
 */
```

src/services/auth/plugins/roles/preHandlers.ts
```diff
@@ -0,0 +1,25 @@
...
+export function whitelistRoles(...strategies: RoleStrategy[]): RouteHandlerMethod {
```
ReidyT, line 12: I'm not sure if this name really help to understand what the function does, I don't found a better name for now but it could be something like that `authorizeUserWithAtLeastOnePassingStrategy`. It's too long I know ! But it adds clarity... I'll try to suggest something shorter ^^ What do you think ? @Defferrard , @spaenleh and @pyphilia ^^ If you think this name is good, just skip my comment :)

ReidyT (approved):
Really nice PR, good job ! LGTM, I just let some suggestions, be free to ignore them if you think it's not necessary ;)

spaenleh:
The frontend is ready at https://github.com/graasp/graasp-builder/pull/1350

spaenleh (approved):

src/services/auth/plugins/magicLink/index.ts
```diff
@@ -120,16 +117,28 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
         log,
       } = request;
+      const member = notUndefined(user?.member);
       const redirectionUrl = getRedirectionUrl(log, url ? decodeURIComponent(url) : undefined);
+      await db.transaction(async (manager) => {
+        const repositories = buildRepositories(manager);
+        memberService.refreshLastAuthenticatedAt(member.id, repositories);
+        // on auth, if the user used the email sign in, its account gets validated
+        if (authInfo?.emailValidation && !member.isValidated) {
+          memberService.validate(member.id, repositories);
```
pyphilia, line 139: Missing an await?

pyphilia (changes requested):

pyphilia (approved):

PR #1117, opened 2024-06-18
build: migrate to NodeNext

### LINKED ISSUE #1116: Can't update ESM-only @graasp/translations package
The `@graasp/translations` package is configured to export an ESM only build created with `tsc`. 

The backend currently compiles down to CJS, so it is not possible to use newer version of the translations package.

There are multiple solutions to this.

### Solution 1: Convert the backend to use ESM

This is the most "future-proof" solution but also the most time-consuming and un-certain.
Having tried to do it quite un-successfully in #1117 the following issues have surfaced:
- Need to use fully-qualified paths in imports -> this is easily handled by a code-mod and the small number of manual fixes is OK.
- the tests need to be able to run on the ESM source, but Jest (our current testing framework) has ESM support under the experimental frag since at least 4 years, and trying to make tests work with the migrated ESM source code proved unsuccessful. 
  - We could use a different testing framework, like `vitest` which support ESM natively, but the size/number of tests to refactor is quite high. A more progressive transition to vitest is preferable.

### Solution 2: What if the issue is not the backend but actually `@graasp/translations`

While trying out Solution 1 and it miserably failing after 2 days. The question arrose whether we could maybe make it so that translations could be used inside a commonJS project.

The main blocker for this was: 
- since translations was using the `module=nodenext` config, the json file imports had to be asserted with the new `with { type: 'json' }` syntax
- this new syntax is not convertible as is in commonJS when using default tsc compilation to `module=commonJS`... 

So we are done ? no solution ? 

No. We can use a :star: bundler ✨  ! 

And indded, it worked, using a bundler allows us to produce a commonJS package that conforms to the requirements of the backend.

spaenleh:
- **fix: set module resolution to NodeNext and add type=module in package**
- **fix: update imports to use new nodenext syntax with fully qulified imports**
- **fix: eslint errors**
- **fix: ts errors**
- **fix: pesky require in esm files**

fix #1116


src/services/item/plugins/action/test/ws.test.ts
```diff
@@ -1,20 +1,22 @@
...
 import {
   ItemOpFeedbackErrorEvent,
   ItemOpFeedbackEvent,
   memberItemsTopic,
-} from '../../../ws/events';
-import { ActionRequestExportRepository } from '../requestExport/repository';
+} from '../../../ws/events.js';
+import { ActionRequestExportRepository } from '../requestExport/repository.js';
+
+const waitForExpect = waitForExpectDefault.default;
```
pyphilia, line 19: Would having this in a util easier? And you export it and use this proxy file instead of setting the default each time?

src/services/item/plugins/embeddedLink/schemas.ts
```diff
@@ -1,8 +1,10 @@
-import S from 'fluent-json-schema';
+import Ss from 'fluent-json-schema';
 
 import { ItemType } from '@graasp/sdk';
 
-import { error } from '../../../../schemas/fluent-schema';
+import { error } from '../../../../schemas/fluent-schema.js';
+
+const S = Ss.default;
```
pyphilia, line 7: Same comment as for `waitFor`.

src/services/auth/plugins/password/service.ts
```diff
@@ -1,23 +1,23 @@
-import Redis from 'ioredis';
+import { Redis } from 'ioredis';
```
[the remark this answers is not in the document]
spaenleh: Default imports from CommonJS packages seem like they don't behave like they should. So I moved as much of them to named imports.

src/services/item/plugins/action/test/ws.test.ts
spaenleh: Actually I just copied the `wait-for-export` package code in a utils file. It works better and we do not need to worry about those import issues.

src/services/item/plugins/embeddedLink/schemas.ts
spaenleh: This one was also fixed by using the named import instead of the default.

PR #1120, opened 2024-06-20
fix: update translations

spaenleh:
In this PR I try to make it possible to update the translations package.

For this I had to:
- make the translations package a dual package that exports ESM and CJS libraries
- update the tsconfig.json `moduleResolution` from `node` to `NodeNext` which meant update a few of the imports to be better qualified.
- make some fixes in sdk (similar to translations)


Defferrard (approved):
LGTM!

I think you've miswritten the PR title

pyphilia (approved):
LGTM 🧇

PR #1126, opened 2024-06-26
fix: eslint warnings

spaenleh:
In this PR:
- add the jest extension to devcontainer
- fix eslint warnings when using the default import of packages
- upgrade yarn


Defferrard (approved):
LGTM!

By the way, there is a query property for the `app.inject` in our test file context. So we shouldn't use these `qs.stringify` function

pyphilia (approved):
LGTM 💯

PR #1129, opened 2024-06-27
fix: email regex for `.swiss` emails

### LINKED ISSUE #1128: Email regex used in `/member/search` does not allow to reply with emails ending in `.swiss`
Email adresses using TLD (top level domain) names with less than 2 and more than 4 characters are removed before the reply is sent to the client.

for example: `bob@hello.swiss` will never be sent because `.swiss` is 5 chars long.

See the line where the regex is defined:
https://github.com/graasp/graasp/blob/79cadc3ddd560744a0d0a96181e9cb0ab01d4a34/src/services/member/schemas.ts#L9

And where it is used inside the schema definition for the response to the getManyBy email:
https://github.com/graasp/graasp/blob/79cadc3ddd560744a0d0a96181e9cb0ab01d4a34/src/services/member/schemas.ts#L166

spaenleh:
fix #1128 

- Increase the tolerance for TLD names up to 63 chars long (according to https://www.rfc-editor.org/rfc/rfc1034)
- Add unit test for the email regex
- App regression test for the `/members/search` endpoint with offending `.swiss` email

### DIFF
--- a/src/services/member/schemas.ts
+++ b/src/services/member/schemas.ts
@@ -6,7 +6,25 @@ import {
 
 import { NAME_REGEX, UUID_REGEX } from '../../schemas/global';
 
-const EMAIL_REGEX = '^[\\w-\\.]+@([\\w-]+\\.)+[\\w-]{2,4}$';
+/**
+ * This allows email adresses that are structured as follows:
+ * - at least one of
+ *   - word character
+ *   - dash
+ *   - dot
+ * - the @ symbol
+ * - multiple times
+ *   - at least one of
+ *     - word cahracter
+ *     - dash
+ *   - a dot
+ * - a tld extension that is between 2 and 63 chars long
+ *
+ * The maximum length of the TLD is defined by https://www.rfc-editor.org/rfc/rfc1034
+ * and is 63 octets (chars)
+ *
+ */
+export const EMAIL_REGEX = '^[.-\\w]+@[.-\\w]+.[-\\w]{2,63}$';
 
 export default {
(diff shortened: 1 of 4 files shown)


src/services/member/schemas.ts
```diff
@@ -6,7 +6,25 @@ import {
...
+ *     - word cahracter
+ *     - dash
+ *   - a dot
+ * - a tld extension that is between 2 and 63 chars long
+ *
+ * The maximum length of the TLD is defined by https://www.rfc-editor.org/rfc/rfc1034
+ * and is 63 octets (chars)
+ *
+ */
+export const EMAIL_REGEX = '^[.-\\w]+@[.-\\w]+.[-\\w]{2,63}$';
```
pyphilia, line 27: Thanks for the changes, but do we really need this regex? I let you decide if it's better to keep the regex, or if we simplify the schema.

pyphilia (approved):

src/services/member/schemas.ts
spaenleh: @pyphilia I see, well, this regex might be a bit strict, so we would be better off in the future to remove it. I do not think this is too urgent, and I am proud of what I did for the tests ahah.

Defferrard (commented):
I don't get why we're not using the email format of AJV ?
If we need it outside of controller, it should still be possible.

Is AJV too/not enough restrictive ? By the way, we can define our own AJV format as I did for the `Strong Password`. It would be easier to use in the schemas.

PR #1132, opened 2024-06-28
fix: already taken email test result with a conflict

Defferrard:
I've missed a change in the tests. Here the fix so it matches with the implementation.


spaenleh (approved):
💯

PR #1137, opened 2024-07-01
chore: add missing env var error, for better DX

spaenleh:
In this PR I propose to send a more meaningful error when and env var is missing. 

I spent a good amount of time figuring out the origin of the issue. 
The server would not start, because a new env variable was missing, but the error made it look like there was an error somewhere else. 

I try to create a new error that gives more information and indicates that the issue comes from a missing env variable.


Defferrard (approved):
LGTM!

pyphilia (approved):

PR #1142, opened 2024-07-01
fix: remove memberId authentication from item login

Defferrard:
Graasp Builder & Graasp Player should remove the combobox, that allow the memberId authentication


spaenleh (approved):

pyphilia (approved):
🙏

PR #1158, opened 2024-07-03
fix: logout member on delete

### LINKED ISSUE #1157: Logout member on delete
On Delete member we should directly logout the member by deleting the session.
This will simplify the client code, so it does not need to call the logout endpoint on cascade.

With the current implementation, when deleting the member, the client then calls `/logout`, but the call fails with a "Member not found" since the member serialisation is applied, but the member does not exist anymore.

spaenleh:
fix #1157 

In this PR:
- add `request.session.delete()` on /logout to ensure cookie is removed from the browser
- add types on schemas
- fix an AJV warning in the h5p schema by adding `allowUnionTypes: true` to the options
- deprecate `DELETE /members/:id` and add a new `DELETE `/members/current` instead 
- add the logout directly on these delete routes
- refactor the members profile plugin and routes to make the responses more consistent
  - NO_CONTENT when there is nothing
  - OK when there is data
  - UNAUTHORIZED when not logged in
- move the memberProfile repository to a custom repository class
- experiment with the querybuilder for efficient queries, but these are not type-safe and lack transformations ... sad

### DIFF
--- a/src/services/item/controller.ts
+++ b/src/services/item/controller.ts
@@ -132,7 +132,9 @@ const plugin: FastifyPluginAsync = async (fastify) => {
       preHandler: optionalIsAuthenticated,
     },
     async ({ user, params: { id } }) => {
-      return itemService.getPacked(user?.member, buildRepositories(), id);
+      const item = await itemService.getPacked(user?.member, buildRepositories(), id);
+      console.log(item);
+      return item;
     },
   );
 
--- a/src/services/item/controller.ts
+++ b/src/services/item/controller.ts
@@ -133,7 +133,6 @@ const plugin: FastifyPluginAsync = async (fastify) => {
     },
     async ({ user, params: { id } }) => {
       const item = await itemService.getPacked(user?.member, buildRepositories(), id);
-      console.log(item);
       return item;
     },
   );
--- a/src/services/member/controller.ts
+++ b/src/services/member/controller.ts
@@ -107,10 +107,7 @@ const controller: FastifyPluginAsync = async (fastify) => {
 
   // delete member
   /**
-   * @deprecated
-   * TODO: Fix this endpoint as it should not need the member ID.
-   * We only want to delete the currently authenticated member
-   * and this information is already provided by the session
+   * @deprecated use the delete member function without the id param
    */
   fastify.delete<{ Params: IdParam }>(
     '/:id',
@@ -139,7 +136,7 @@ const controller: FastifyPluginAsync = async (fastify) => {
       const { user } = request;
       const member = notUndefined(user?.member);
       return db.transaction(async (manager) => {
-        await memberService.deleteCurrent(member, buildRepositories(manager));
+        await memberService.deleteCurrent(member.id, buildRepositories(manager));
         // logout member
         request.logOut();
         // remove session from browser
--- a/src/services/member/controller.ts
+++ b/src/services/member/controller.ts
@@ -17,6 +17,7 @@ import { Member } from './entities/member';
 import { EmailAlreadyTaken } from './error';
 import { StorageService } from './plugins/storage/service';
 import {
+  deleteCurrent,
   deleteOne,
   getCurrent,
   getMany,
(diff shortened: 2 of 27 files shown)


src/services/item/controller.ts
```diff
@@ -129,7 +129,9 @@ const plugin: FastifyPluginAsync = async (fastify) => {
       preHandler: optionalIsAuthenticated,
     },
     async ({ user, params: { id } }) => {
-      return itemService.getPacked(user?.member, buildRepositories(), id);
+      const item = await itemService.getPacked(user?.member, buildRepositories(), id);
+      console.log(item);
```
spaenleh, line 133: Remove

src/services/member/controller.ts
```diff
@@ -103,12 +104,44 @@ const controller: FastifyPluginAsync = async (fastify) => {
   );
 
   // delete member
+  /**
+   * @deprecated
+   * TODO: Fix this endpoint as it should not need the member ID.
```
Defferrard, line 109: I'm just highlighting this `TODO`. I think it should be converted as an issue ?

src/services/member/service.ts
```diff
@@ -94,6 +94,10 @@ export class MemberService {
     });
   }
 
+  async deleteCurrent(member: Member, { memberRepository }: Repositories) {
```
Defferrard, line 97: Instead of asking the whole member, you can ask for only an Id.

It makes this function easier to call. We wouldn't need to `SELECT` the whole `Member` on the `database` before calling this function.
It would be easier to refactor (B.e : Converting all members into Account)
It reduces the possibilities of a bad implementation of this function for the future.

src/services/member/plugins/profile/repository.ts
```diff
@@ -39,17 +50,35 @@ const MemberProfileRepository = AppDataSource.getRepository(MemberProfile).exten
...
 
     return memberProfile;
-  },
+  }
+
+  async patch(memberId: string, data: Partial<IMemberProfile>): Promise<MemberProfile | null> {
+    /**
+     * this function is a test to try to make the queries more efficient while still
+     * getting the return values in one go.
+     * but this is really type unsafe, sinc ewe have to use the raw sql output provided by typeorm
```
Defferrard, line 65:
```suggestion
     * but this is really type unsafe, since we have to use the raw sql output provided by typeorm
```

src/services/member/plugins/profile/service.ts
```diff
@@ -1,28 +1,24 @@
...
-      throw new UnauthorizedMember();
-    }
+  async post(member: Member, { memberProfileRepository }: Repositories, data: IMemberProfile) {
     const profile = await memberProfileRepository.createOne({ ...data, member });
     return profile;
   }
 
-  async get(memberId: string, repositories: Repositories) {
-    const { memberProfileRepository } = repositories;
+  async get(_actor: Actor, { memberProfileRepository }: Repositories, memberId: UUID) {
```
Defferrard, line 21: Since we're not using the `_actor` parameter, should we simply remove it from the function definition ?

Defferrard (approved):
Well done ! LGTM! I just have a few question which I think is related to the global implementation of our backend.

src/services/member/controller.ts
spaenleh: Yes done !

src/services/member/service.ts
spaenleh: I see, do you prefer:
```ts
async deleteCurrent(member: { id: string }, { memberRepository }: Repositories) {
```

or 

```ts
async deleteCurrent(memberId: string, { memberRepository }: Repositories) {
```

Defferrard: I'd say 
```typescript
async deleteCurrent(memberId: string, { memberRepository }: Repositories) {
```

But I'm open to the first proposal which is interesting !

spaenleh: okay ! Yes, in the end I used the second approach as I think it is less ambiguous that a string is required. The first one is "cool" because at the call site you can pass anything that contains an is prop that is a string, so it is very flexible. But the risk is that you spread the object, and then you might get un-expected consequences, mainly it might contain more than what you think.

src/services/member/plugins/profile/errors.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const GraaspMemberProfileError = ErrorFactory('graasp-member-profile');
+
+export class MemberProfileNotFound extends GraaspMemberProfileError {
+  constructor(data?: unknown) {
+    super(
+      {
+        code: 'GMPERR001',
+        statusCode: StatusCodes.NOT_FOUND,
+        message: 'Member Profile not found',
```
pyphilia, line 13: Translate.

src/services/member/plugins/profile/repository.ts
```diff
@@ -7,7 +8,17 @@ import { MemberNotFound } from '../../../../utils/errors';
...
+class MemberProfileRepository {
+  private repository: Repository<MemberProfile>;
+
+  constructor(manager?: EntityManager) {
+    if (manager) {
+      this.repository = manager.getRepository(MemberProfile);
+    } else {
+      this.repository = AppDataSource.getRepository(MemberProfile);
+    }
+  }
```
pyphilia, line 20: I think it can be replaced by `extends AbstractRepository<TYPE>`, check `MemberRepository`.

src/services/member/plugins/profile/repository.ts
```diff
@@ -39,17 +50,35 @@ const MemberProfileRepository = AppDataSource.getRepository(MemberProfile).exten
...
+      .returning('*')
+      .execute();
 
-  async patch(memberId: string, data: Partial<IMemberProfile>): Promise<void> {
-    await this.update({ member: { id: memberId } }, data);
-  },
-});
+    return profile;
+  }
+}
```
pyphilia, line 82: Do we really want to push that? If you want to keep it somewhere, it can be added to an issue or technical debt board.

src/services/member/plugins/profile/schemas.ts
```diff
@@ -1,90 +1,66 @@
...
 export const createProfile = {
-  response: {
-    201: profileMember,
-    '4xx': error,
-  },
   body: S.object()
-    .prop('visibility', S.boolean())
     .additionalProperties(false)
+    .prop('visibility', S.boolean())
+    // QUESTION: why is the bio always required ? Can't we only update a social link ?
```
pyphilia, line 31: I think it can be optional, but maybe we want to enforce users to fill this field for a better look in library (I made it up, might not be consistent with all endpoints). I'm good with any decision.

pyphilia (changes requested):

pyphilia (approved):

PR #1162, opened 2024-07-03
fix: filter out smartly given descendants

pyphilia:
Prevent getting lots of items when we know the structure (descendants)

This avoids running enormous sql queries: eg. get OR path...


src/services/authorization.ts
```diff
@@ -268,6 +268,59 @@ export const filterOutPackedItems = async (
...
+  descendants: Item[],
+  options?: { showHidden?: boolean },
+): Promise<PackedItem[]> => {
+  const { itemMembershipRepository } = repositories;
+  const showHidden = options?.showHidden ?? true;
+  if (!descendants.length) {
+    return [];
+  }
+
+  // TODO: optimize with on query
```
spaenleh, line 287: ?
```suggestion
  // TODO: optimize with one query
```

spaenleh (approved):
Looks good !

PR #1180, opened 2024-07-08
feat: determine publication state backend

ReidyT:
- Move Published and Validation folders to publication
- Implement the publication state in the backend


src/services/item/plugins/publication/publicationState/fixtures.ts
Defferrard, line 1: If this file is only for test case, it should either have `*.test.ts` suffix, or being in a `test/` folder. Otherwise, it would be builded in production.

There is also the root test folder that you can use.

src/services/item/plugins/publication/publicationState/fixtures.ts
```diff
@@ -0,0 +1,55 @@
...
+  ItemType,
+  ItemValidationProcess,
+  ItemValidationStatus,
+  buildPathFromIds,
+} from '@graasp/sdk';
+
+import { ItemTag } from '../../itemTag/ItemTag';
+import { ItemMetadata, ItemValidationGroupStatus } from './types';
+
+export const ItemEntityFactory = (
```
Defferrard, line 14: I'm not sure if it is what you're expecting by naming it as `...EntityFactory`, but it doesn't create a TypeORM Entity.
Otherwise, should we have this Factory in SDK ?

src/services/item/plugins/publication/publicationState/index.test.ts
```diff
@@ -0,0 +1,83 @@
...
+import { ItemTestUtils } from '../../../test/fixtures/items';
+
+const testUtils = new ItemTestUtils();
+
+// mock datasource
+jest.mock('../../../../../plugins/datasource');
+
+describe('Publication Controller', () => {
+  let app: FastifyInstance;
+  let actor;
```
Defferrard, line 20: This variable is `any`, it can create unexpected behavior.

```suggestion
  let actor: Actor;
```

src/services/item/plugins/publication/publicationState/index.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  afterEach(async () => {
+    jest.clearAllMocks();
+    await clearDatabase(app.db);
+    actor = null;
+    app.close();
+  });
+
+  describe('GET /publication', () => {
+    describe('Signed Out', () => {
+      let member;
```
Defferrard, line 31: This is any as well
```suggestion
      let member: Member;
```

src/services/item/plugins/publication/publicationState/types.ts
```diff
@@ -0,0 +1,14 @@
+import { PackedItem } from '../../../ItemWrapper';
+import { ItemValidation } from '../validation/entities/ItemValidation';
+import { ItemValidationGroup } from '../validation/entities/ItemValidationGroup';
+
+type ValidationStatus = Pick<ItemValidation, 'status'>;
+
+export type ItemMetadata = Pick<PackedItem, 'updatedAt' | 'path' | 'type'> &
+  Partial<Pick<PackedItem, 'public'>>;
```
Defferrard, line 8: Is it the same as :
```suggestion
export type ItemMetadata = Partial<Pick<PackedItem, 'updatedAt' | 'path' | 'type' | 'public'>>;
```

Defferrard (approved):
Seems pretty good to be, well done !

src/services/item/plugins/publication/publicationState/fixtures.ts
ReidyT: Good catch !

src/services/item/plugins/publication/publicationState/fixtures.ts
ReidyT: Yeah you right ! It was initially for that, but it is not the case anymore... I'll rename it as `ItemMetadataFactory`. I don't think it's useful to put it into SDK, because it is only used to simplify the backend test...

src/services/item/plugins/publication/publicationState/index.test.ts
ReidyT: Good catch !

src/services/item/plugins/publication/publicationState/types.ts
ReidyT: Not exactly ! Yours set all the attributes to nullable while mine construct a type like `{ updatedAt, path, type, public? }` where the only nullable property is public. 

If you know a simpler way to achieve that, I'm open to suggestions :).

Defferrard: Oh, you're right, I've missed that. I think it is alright then ! Thank you !

pyphilia (approved):

src/services/item/plugins/publication/publicationState/test/fixtures.ts
```diff
@@ -0,0 +1,55 @@
...
+): ItemMetadata => {
+  const path = item?.path ?? v4();
+
+  return {
+    updatedAt: item?.updatedAt ?? new Date(),
+    path: item?.parentItem ? buildPathFromIds[(item.parentItem.path, path)] : path,
+    type: item?.type ?? ItemType.FOLDER,
+    public: isPublic ? ({ type: ItemTagType.Public } as ItemTag) : undefined,
+  };
+};
```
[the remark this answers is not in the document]
ReidyT: It's because I was unable to use the `PackedFolderItemFactory`. When I used it I had issues because the service wanted a "backend" `PackedItem` and I provided an "SDK" `PackedItem`. I spent a lot of time trying to work with these two types and I finally decided to use a specific type.

PR #1186, opened 2024-07-11
fix: add email in link for email update

### LINKED ISSUE #1184: Add new email in link as query string on email udpate request
Adding the new email as a query parameter in the link that is sent by email when requesting an update of the email used for the account would allow us to display it in the frontend adding more feedback for the user. 

It would be safer to implement an endpoint that allows to get the email when sending the request token. This might be done later.

spaenleh:
fix #1184


Defferrard (approved):
LGTM!

ReidyT (approved):
LGTM 😶‍🌫️

PR #1196, opened 2024-07-15
feat(refactor): update MobileService to allow injection using DI

ReidyT:
- use authService instead of fastify instance in mobile service
- inject mobile service using DI


spaenleh (approved):
🎉

PR #1198, opened 2024-07-15
feat(refactor): update magiclink service to allow injection using DI

ReidyT:
- use authService instead of fastify instance in the service
- inject the service using DI


spaenleh (approved):
LGTM ! 🚈

PR #1204, opened 2024-07-16
fix: remove leaked information and remove unused errors

spaenleh:
In this PR:
- remove leaked information from errors
- specify data type expected by the errors instead of using `unknown`
- remove unused errors
- make heavier use of the actor type in services that are used inside authenticated controllers, so we do not need to throw if the member does not exist as it will always be defined.

### DIFF
--- a/src/services/auth/plugins/password/service.ts
+++ b/src/services/auth/plugins/password/service.ts
@@ -193,7 +193,7 @@ export class MemberPasswordService {
     // Fetch the member's password
     const memberPassword = await memberPasswordRepository.getForMemberId(member.id);
     if (!memberPassword) {
-      throw new MemberWithoutPassword({ email });
+      throw new MemberWithoutPassword();
     }
     // Validate credentials to build token
     if (await comparePasswords(password, memberPassword.password)) {
--- a/src/services/item/repository.ts
+++ b/src/services/item/repository.ts
@@ -103,7 +103,7 @@ export class ItemRepository {
     } = args;
 
     if (parent && !isItemType(parent, ItemType.FOLDER)) {
-      throw new ItemNotFolder(parent);
+      throw new ItemNotFolder(parent.id);
     }
 
     // TODO: extra
@@ -185,7 +185,7 @@ export class ItemRepository {
     options: { withOrder?: boolean } = {},
   ): Promise<Item[]> {
     if (!isItemType(parent, ItemType.FOLDER)) {
-      throw new ItemNotFolder(parent);
+      throw new ItemNotFolder(parent.id);
     }
 
     const query = this.repository
--- a/src/services/item/repository.ts
+++ b/src/services/item/repository.ts
@@ -103,7 +103,7 @@ export class ItemRepository {
     } = args;
 
     if (parent && !isItemType(parent, ItemType.FOLDER)) {
-      throw new ItemNotFolder(parent.id);
+      throw new ItemNotFolder({ id: parent.id });
     }
 
     // TODO: extra
@@ -185,7 +185,7 @@ export class ItemRepository {
     options: { withOrder?: boolean } = {},
   ): Promise<Item[]> {
     if (!isItemType(parent, ItemType.FOLDER)) {
-      throw new ItemNotFolder(parent.id);
+      throw new ItemNotFolder({ id: parent.id });
     }
 
     const query = this.repository
@@ -358,7 +358,7 @@ export class ItemRepository {
 
       // cannot move inside non folder item
       if (!isItemType(parentItem, ItemType.FOLDER)) {
-        throw new ItemNotFolder(parentItemId);
+        throw new ItemNotFolder({ id: parentItemId });
       }
 
       // fail if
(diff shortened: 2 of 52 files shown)


src/services/item/repository.ts
```diff
@@ -103,7 +103,7 @@ export class ItemRepository {
     } = args;
 
     if (parent && !isItemType(parent, ItemType.FOLDER)) {
-      throw new ItemNotFolder(parent);
+      throw new ItemNotFolder(parent.id);
```
Defferrard, line 106: I think it is inconsistent with your other throws (B.e MemberNotFound)
```suggestion
      throw new ItemNotFolder({ id: parent.id });
```

src/services/item/repository.ts
```diff
@@ -185,7 +185,7 @@ export class ItemRepository {
     options: { withOrder?: boolean } = {},
   ): Promise<Item[]> {
     if (!isItemType(parent, ItemType.FOLDER)) {
-      throw new ItemNotFolder(parent);
+      throw new ItemNotFolder(parent.id);
```
Defferrard, line 188:
```suggestion
      throw new ItemNotFolder({ id: parent.id });
```

src/services/itemMembership/service.ts
```diff
@@ -211,15 +195,15 @@ export class ItemMembershipService {
...
+      throw new ItemMembershipNotFound({ id: itemMembershipId });
     }
 
     const memberships = itemIdToMemberships[item.id];
     const otherAdminMemberships = memberships.filter(
       (m) => m.id !== itemMembershipId && m.permission === PermissionLevel.Admin,
     );
     if (otherAdminMemberships.length === 0) {
-      throw new CannotDeleteOnlyAdmin(item);
+      throw new CannotDeleteOnlyAdmin(item.id);
```
Defferrard, line 206:
```suggestion
      throw new CannotDeleteOnlyAdmin({ id: item.id });
```

src/services/member/plugins/action/test/service.test.ts
```diff
@@ -37,19 +38,22 @@ describe('Action member service', () => {
...
+    //   ({ app } = await build({ member: null }));
 
-      await expect(
-        getActionMemberService().getFilteredActions(actor, buildRepositories(), {}),
-      ).rejects.toBeInstanceOf(UnauthorizedMember);
-    });
+    //   await expect(
+    //     getActionMemberService().getFilteredActions(actor, buildRepositories(), {}),
+    //   ).rejects.toBeInstanceOf(UnauthorizedMember);
+    // });
```
Defferrard, line 49: I think we can delete this test, right ? No needs to keep it

src/services/auth/plugins/password/service.ts
```diff
@@ -193,7 +193,7 @@ export class MemberPasswordService {
     // Fetch the member's password
     const memberPassword = await memberPasswordRepository.getForMemberId(member.id);
     if (!memberPassword) {
-      throw new MemberWithoutPassword({ email });
+      throw new MemberWithoutPassword();
```
Defferrard, line 196: On the function `getManyByEmail` from the `MemberRepository`, we're still throwing the `email`. Is it intended to have this difference ?

Defferrard (approved):
LGTM! Well done! The code will looks more friendly after this PR :) I've just seen some inconsistent and wondering if it is intented.

src/services/item/repository.ts
spaenleh: Yes you are right, my issue is that with the object signature, it is possible to pass the full object and it will not complain. This could potentially lead to the same issue that created the need for this PR. 

I can use the object signature in all errors but then I need to explicitly strip out any additional properties that could be passed alongside. 

@Defferrard Do you see a better way of doing it ?

src/services/member/plugins/action/test/service.test.ts
spaenleh: Yes will delete it !

src/services/auth/plugins/password/service.ts
spaenleh: I would say they are different, since in the `getManyByEmail` we are adding these errors to the `errors` key of the result object. But it is true we could omit it.

src/services/item/repository.ts
Defferrard: From https://stackoverflow.com/questions/49580725/is-it-possible-to-restrict-typescript-object-to-contain-only-properties-defined someone suggest to create theses types :
```typescript
type Impossible<K extends keyof any> = {
  [P in K]: never;
};
type NoExtraProperties<T, U extends T = T> = U & Impossible<Exclude<keyof U, keyof T>>;
```

So we would to use it that way : 
```typescript
function a(s: NoExtraProperties<{a:string}>){
  return s
}
a({a:""}) // OK
a({}) // NOK: Property 'a' is missing in type '{}' but required in type '{ a: string; }'
a({a:"", b:""}) // NOK: Object literal may only specify known properties, and 'b' does not exist in type 
```

Fun fact, an issue is opened since 2016 https://github.com/microsoft/TypeScript/issues/12936, but there is still no native solution.

PR #1207, opened 2024-07-16
feat: refactor etherpad service

ReidyT:
- Inject Etherpad service using DI
- Use EtherpadConfig class to store the configuration
- Validate the url and api key env in the const file using Validators
- Refactor the wrapErrors


src/services/item/plugins/etherpad/serviceConfig.ts
```diff
@@ -0,0 +1,49 @@
...
+    return this._publicUrl;
+  }
+
+  public get cookieDomain() {
+    return this._cookieDomain;
+  }
+
+  public get apiVersion() {
+    return this._apiVersion;
+  }
```
Defferrard, line 48: I don't remember what we've talked about exactly yesterday, but you could simply have `public readonly attributes`

src/utils/validators/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+function readEnv(envName: string): Variable {
+  return {
+    name: envName,
+    value: process.env[envName],
+  };
+}
+
+export function validateEnv(envName: string, validator: Validator): string {
+  return validator.validate(readEnv(envName));
+}
```
Defferrard, line 14: If you still want to have multiple validatior, is it possible to just have an array here ?

```suggestion
export function validateEnv(envName: string, ...validators: Validator[]): string {
    const result = readEnv(envName);
    for(const validator in validators){
        result = validator.validate(result);
    }
  return result;
}
```

src/di/container.ts
```diff
@@ -126,4 +132,20 @@ export const registerDependencies = (instance: FastifyInstance) => {
...
+  registerValue(
+    Etherpad,
+    wrapErrors(
+      new Etherpad({
+        url: etherPadConfig.url,
+        apiKey: etherPadConfig.apiKey,
+        apiVersion: etherPadConfig.apiVersion,
+      }),
+    ),
+  );
```
Defferrard, line 148: I think we need to discuss about it, because I probably don't fully understand the use of this `wrapErrors` function. But I think we should handle it with custom `decorators` instead (Or any other suggestion). So we can get rid of this call.

Defferrard (approved):
Good work ! LGTM ! I just have some suggestions that can be discussed !

src/services/item/plugins/etherpad/service.ts
```diff
@@ -13,31 +15,39 @@ import { Item, isItemType } from '../../entities/Item';
...
-  private readonly log: FastifyBaseLogger;
+  private readonly log: BaseLogger;
 
   constructor(
     etherpad: Etherpad,
-    padNameFactory: () => string,
-    publicUrl: string,
-    cookieDomain: string,
+    @inject(ETHERPAD_NAME_FACTORY_DI_KEY) padNameFactory: PadNameFactory,
+    etherPadConfig: EtherPadServiceConfig,
```
spaenleh, line 43:
```suggestion
    etherpadConfig: EtherpadServiceConfig,
```

spaenleh (approved):
Looks good to me, I let you discuss raised issues. I am open to discussion too.

src/services/item/plugins/etherpad/serviceConfig.ts
ReidyT: Yeah you're right !

src/utils/validators/utils.ts
ReidyT: Yes it is, I had finally simplify to only one validation but we could keep it to offer the possibility to add multiple validators, thanks for your suggestion !

src/di/container.ts
ReidyT: Here I just moved the existing code, so I didn't refactor the wrapErrors. I'll try to improve it if I can

PR #1221, opened 2024-07-18
fix: sort descendants as number

pyphilia:
Descendants does not return the correct order. In real context, typeorm actually returns strings, and they are not sorted correctly when representing numbers (1001 < 20).


spaenleh (approved):

Defferrard (approved):
LGTM!

PR #1222, opened 2024-07-19
feat: copy suffix is determined by siblings

Defferrard:


src/services/item/repository.ts
```diff
@@ -555,6 +557,33 @@ export class ItemRepository {
     }
   }
 
+  /**
+   * Return a copy with a suffix of the string given in parameter. The suffix is determined by the siblings of the item.
+   * If there is no sibling with the same name, the copied name will be the same as the original name.
+   * The suffix respect the format " (0)". "0" is a succint positive number starting at 2.
+   * If the string given in parameter already have a valid suffix, increment the number.
```
spaenleh, line 564:
```suggestion
   * If the string given in parameter already has a valid suffix, increment the number.
```

src/services/item/repository.ts
```diff
@@ -555,6 +557,33 @@ export class ItemRepository {
     }
   }
 
+  /**
+   * Return a copy with a suffix of the string given in parameter. The suffix is determined by the siblings of the item.
+   * If there is no sibling with the same name, the copied name will be the same as the original name.
+   * The suffix respect the format " (0)". "0" is a succint positive number starting at 2.
```
spaenleh, line 563:
```suggestion
   * The suffix respects the format " (0)". "0" is a succinct positive number starting at 2.
```

src/services/item/repository.ts
```diff
@@ -555,6 +557,33 @@ export class ItemRepository {
...
+   * If the string given in parameter already have a valid suffix, increment the number.
+   * If the copied name exceed the maximum characters allowed, the original name will be shorten from the end,
```
spaenleh, line 565:
```suggestion
   * If the copied name exceeds the maximum characters allowed, the original name will be shorten from the end,
```

src/services/item/repository.ts
```diff
@@ -555,6 +557,33 @@ export class ItemRepository {
...
+   * the copied name will be equals to the maximum allowed.
```
spaenleh, line 566:
```suggestion
   * the copied name will have the maximum allowed length.
```

src/services/item/service.ts
```diff
@@ -682,8 +682,16 @@ export class ItemService {
...
 
+    const siblings: Item[] = [];
+    if (parentItem) {
+      siblings.push(...(await itemRepository.getChildren(parentItem)));
+    } else {
+      const itemMemberships = await itemMembershipRepository.getAccessibleItems(actor, {}, null);
+      siblings.push(...itemMemberships.data.map((m) => m.item));
+    }
+
     // TODO: args?
```
spaenleh, line 693: Should we remove this todo ?

spaenleh (approved):
Looks good !

src/services/item/repository.ts
```diff
@@ -555,6 +557,33 @@ export class ItemRepository {
...
+   */
+  private addCopySuffix(name: string, siblings: Item[]): string {
+    const maxIterations = siblings.length;
+    let result = name;
+    for (let i = 0; i < maxIterations; i++) {
+      const twinIndex = siblings.findIndex((sibling) => sibling.name === result);
+      if (twinIndex >= 0) {
+        result = this.incrementCopySuffix(result);
+        // By definition, the result string will never be the same as the current twin, so we can remove it from the list.
+        siblings.splice(twinIndex, 1);
```
ReidyT, line 579: This is a side effect of the `addCopySuffix` and its not specified in the JSDoc. If possible, we should avoid to update the state of an input parameter, because it can lead to unexpected behaviour. If you have to update the array of siblings (I didn't check the whole code now), it could be good to indicate in the JSDoc that the siblings will be mutated.

src/services/item/plugins/file/test/index.test.ts
```diff
@@ -600,16 +600,11 @@ describe('File Item routes tests', () => {
           setTimeout(async () => {
             await expect(copyObjectMock).toHaveBeenCalled();
 
-            const items1 = await testUtils.rawItemRepository.find({ where: { name: item.name } });
-            expect(items1).toHaveLength(1);
+            const items = await testUtils.rawItemRepository.find({ where: { name: item.name } });
+            expect(items).toHaveLength(2);
```
ReidyT, line 604: Just to be sure here, when you copy an item, the copy has exactly the same name ?

ReidyT (approved):
LGTM, nice job! I just let a comment about a side effect of one function 😆

src/services/item/service.ts
Defferrard: I have no idea what it is about. I removed it !

src/services/item/repository.ts
Defferrard: Good catch, I'm not expecting this function to have an impact on the parameter. I'll just copy the array to an another one.

src/services/item/plugins/file/test/index.test.ts
Defferrard: It depends on the destination of the copy
From this diagram :
```mermaid
flowchart TD
    A-->B
    A-->C
```

1. If we copy `C` in `B`. The copy will have the exact same name
```mermaid
flowchart TD
    A-->B
    A-->C((C))
    B-->copy((C))
```

2. If we copy `C` in `A`. The copy will have a suffix, because there is already an Item named `C` in `A`
```mermaid
flowchart TD
    A-->B
    A-->C((C))
    A-->copy(("C (2)"))
```

Probably we need to confirm this is the behavior we want before merging ? @spaenleh  @pyphilia

ReidyT: Ah ok thanks for your explanations and diagrams, it clearer to me ^^

ReidyT: it is important to know that a copied item is a copy ? If it is, should we include "copy" in the new name ?

src/services/item/service.ts
```diff
@@ -682,8 +682,15 @@ export class ItemService {
       await this.hooks.runPreHooks('copy', actor, repositories, { original });
     }
 
-    // TODO: args?
-    const { copyRoot, treeCopyMap } = await itemRepository.copy(item, actor, parentItem);
+    const siblings: Item[] = [];
+    if (parentItem) {
+      siblings.push(...(await itemRepository.getChildren(parentItem)));
+    } else {
+      const itemMemberships = await itemMembershipRepository.getAccessibleItems(actor, {}, null);
```
pyphilia, line 689: I wonder if we really want to do that in the end.. this call is quite heavy (but it might be heavy on other ends of the call, but getting accessible is not light)

pyphilia (approved):

src/services/item/service.ts
Defferrard: We should implement a getAccessibleItemsName which is a light similar function.

We will set a limit of 999 copy, after that, we're having multiple items (1000)

PR #1225, opened 2024-07-19
feat: implement fastify swagger

Defferrard:
The documentation is accessible on `/docs`

We should be able to seperate our route by service with titles
Apparently for some reason models at the end of the page are buggy.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -8,6 +8,7 @@ import fp from 'fastify-plugin';
 import { registerDependencies } from './di/container';
 import databasePlugin from './plugins/database';
 import metaPlugin from './plugins/meta';
+import swaggerPlugin from './plugins/swagger';
 import shared from './schemas/fluent-schema';
 import authPlugin from './services/auth';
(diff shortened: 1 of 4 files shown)


src/app.ts
```diff
@@ -24,6 +26,15 @@ import {
 } from './utils/config';
 
 export default async function (instance: FastifyInstance): Promise<void> {
+  await instance.register(swaggerPlugin, { openapi: { openapi: '3.0.0' } });
+
+  await instance.register(swaggerUiPlugin, {
```
spaenleh, line 31: I this something we want to deploy in production ? I guess it is ok and this would provide some open documentation which is nice.

spaenleh (commented):
Very nice ! I tried it and it works. Lets see how we can improve the documentation for each endpoint in code.

src/app.ts
Defferrard: I think so too !

ReidyT (approved):
Really nice ! LGTM 🎃

pyphilia (approved):

PR #1229, opened 2024-07-19
refactor(repositories): extends abstract repositories

Defferrard:


spaenleh (approved):
LGTM ! Just maybe some tests that still fail ?

ReidyT (approved):
Thanks for the PR! LGTM, good job 😆.

I have a few questions about the repository refactor to ensure I understand the long-term direction:

1. I noticed that some repositories haven't been refactored to AbstractRepository yet. Is this intentional or an omission?
2. Currently, AbstractRepository accepts the manager in the constructor. Are we planning to pass the manager as a function parameter in future PRs (alongside a default repo from AppData)? In this way, we could use the default repo if no manager is passed to the function.
3. Will we introduce interfaces or a similar mechanism to standardize method signatures for find, create, update, and delete operations across repositories?

I know that these questions are unrelated to the current PR but it's to help me have a better understand on the overall repository refactor strategy.

pyphilia:
@ReidyT  I believe he started with repository that were already classes or those that had tests. Moving non-class repositories is a bit difficult because they might have broken types. 

What do you mean by "pass the manager as function args"? Like repository.get(manager)?

I would like interfaces to uniformize recurrent functions!

ReidyT:
@pyphilia thanks for your answers ! I understand for the none class repo, but it is something we will try to do ? 

Concerning the manager, it is to allow to inject the repositories and a possible solution could be the following (I think it is something you already mention some time ago):

```
class ActionRepository {
  const repository;

  constructor () {
    this.repository = AppDataSource.getRepository(Action);
  }

  private getRepository(manager?) {
    return manager ? this.repository;
  }

  public find(manager?) {
    return this.getRepository(manager)...
  }
}

```

spaenleh:
@ReidyT @Defferrard I let you decide how you want to proceed for the next steps with this PR. I simply tried to adapt the code to make the tests pass. If you feel I changed things I should not have, feel free to revert my changes.

ReidyT:
Thanks for your changes @spaenleh ! I'll check that and I'll check with @Defferrard what are the next steps and how we organise us :)

pyphilia:
> @pyphilia thanks for your answers ! I understand for the none class repo, but it is something we will try to do ?
> 
> Concerning the manager, it is to allow to inject the repositories and a possible solution could be the following (I think it is something you already mention some time ago):
> 
> ```
> class ActionRepository {
>   const repository;
> 
>   constructor () {
>     this.repository = AppDataSource.getRepository(Action);
>   }
> 
>   private getRepository(manager?) {
>     return manager ? this.repository;
>   }
> 
>   public find(manager?) {
>     return this.getRepository(manager)...
>   }
> }
> ```

not sure about this, we will need to provide the manager every time, and pass it from controller to service to repository each time. But it's also true creating repository from top to bottom is not so good as well. Not sure what's the best solution.

ReidyT:
Yeah I can understand your point of view, I didn't find another solution ^^ The advantage with this one is that we can easily inject the repositories and it simplify the usage when no manager is required. But it's true that when you need the manager, you have to pass it down...

ReidyT:
Otherwise, maybe we should investigate if it is possible to automatically propagate the transaction inside all the repositories. I know that Sequelize have a solution like that using [cls-hooked](https://github.com/Jeff-Lewis/cls-hooked) (CLS) available on (https://sequelize.org/docs/v6/other-topics/transactions/#automatically-pass-transactions-to-all-queries)[automatically pass transactions to all queries]. 

I also found an interesting [issue on github](https://github.com/typeorm/typeorm/issues/1895) talking about our problem in TypeORM. I'm not a big fan of the proposed solution, as well as what squelize propose. Indeed, the cls-hooked was not updated since the last five years. However, I think this allows us to think about another solution which is to try to automatically propagate the transactions...

I don't know what do you all think about that.

Defferrard:
> Otherwise, maybe we should investigate if it is possible to automatically propagate the transaction inside all the repositories. I know that Sequelize have a solution like that using [cls-hooked](https://github.com/Jeff-Lewis/cls-hooked) (CLS) available on [automatically pass transactions to all queries](https://sequelize.org/docs/v6/other-topics/transactions/#automatically-pass-transactions-to-all-queries).
> 
> I also found an interesting [issue on github](https://github.com/typeorm/typeorm/issues/1895) talking about our problem in TypeORM. I'm not a big fan of the proposed solution, as well as what squelize propose. Indeed, the cls-hooked was not updated since the last five years. However, I think this allows us to think about another solution which is to try to automatically propagate the transactions, in addition to pass the transaction from the controller (I'd say our current issue ^^).
> 
> This [medium article](https://medium.com/@dev.muhammet.ozen/advanced-transaction-management-with-nestjs-typeorm-43a839363491) present another possible solution (at least in the idea).
> 
> I don't know what do you all think about that.


In Sequelize v7, they started using [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage) which is native to NodeJS 22.
An equivalent for TypeORM would be to use : https://github.com/Aliheym/typeorm-transactional
The big question would be : Do we still create our transaction in controller ?

A nice feature could be to handle the transaction from the specification of the controller (As a middleware?)

In order to use this AsyncLocalStorage, we have to
1. Migrate all repository to abstract repository
2. Migrate from NodeJS 20 to 22
  2.1 To migrate, we need to remove Mmmmagic
  2.2 To remove Mmmmagic, we need to migrate to ESM
  2.3 To migrate to ESM we need to migrate from Jest to Vitest


Or we can simply use cls-hooked, but we still need to migrate every repository to our abstract repository.

Anyway, I think theses solutions are the way we should implements transactions and it probably need to be discussed later.

ReidyT:
> > Otherwise, maybe we should investigate if it is possible to automatically propagate the transaction inside all the repositories. I know that Sequelize have a solution like that using [cls-hooked](https://github.com/Jeff-Lewis/cls-hooked) (CLS) available on [automatically pass transactions to all queries](https://sequelize.org/docs/v6/other-topics/transactions/#automatically-pass-transactions-to-all-queries).
> > I also found an interesting [issue on github](https://github.com/typeorm/typeorm/issues/1895) talking about our problem in TypeORM. I'm not a big fan of the proposed solution, as well as what squelize propose. Indeed, the cls-hooked was not updated since the last five years. However, I think this allows us to think about another solution which is to try to automatically propagate the transactions, in addition to pass the transaction from the controller (I'd say our current issue ^^).
> > This [medium article](https://medium.com/@dev.muhammet.ozen/advanced-transaction-management-with-nestjs-typeorm-43a839363491) present another possible solution (at least in the idea).
> > I don't know what do you all think about that.
> 
> In Sequelize v7, they started using [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage) which is native to NodeJS 22. An equivalent for TypeORM would be to use : https://github.com/Aliheym/typeorm-transactional The big question would be : Do we still create our transaction in controller ?
> 
> A nice feature could be to handle the transaction from the specification of the controller (As a middleware?)
> 
> In order to use this AsyncLocalStorage, we have to
> 
> 1. Migrate all repository to abstract repository
> 2. Migrate from NodeJS 20 to 22
>    2.1 To migrate, we need to remove Mmmmagic
>    2.2 To remove Mmmmagic, we need to migrate to ESM
>    2.3 To migrate to ESM we need to migrate from Jest to Vitest
> 
> Or we can simply use cls-hooked, but we still need to migrate every repository to our abstract repository.
> 
> Anyway, I think theses solutions are the way we should implements transactions and it probably need to be discussed later.

I've just realised that AsyncLocalStorage has been available for a long time (introduced in node 12.17 and stable since v16.4.0). This means we could use it without having to migrate to node 22.

We should just be aware of the potential performance degradation and (very rare) (context loss)[https://nodejs.org/docs/latest-v20.x/api/async_context.html#troubleshooting-context-loss]. Maybe it could be interesting to benchmark the usage of AsyncLocalStorage and to decide if the degradation is acceptable or not.

I agree with @Defferrard that propagating the transaction "automatically" is the most interesting solution. Indeed, without its usage, I don't see any other solution except passing the manager in each repository's methods. If anyone knows other potential solutions, I'd be happy to discover their 😆.

src/services/item/plugins/publication/validation/repositories/ItemValidationGroup.ts
```diff
@@ -22,42 +24,33 @@ export const ItemValidationGroupRepository = AppDataSource.getRepository(
...
-  },
+  }
 
   /**
    * Get last item validation group for given item id
    * @param {string} itemId id of the item being checked
    */
-  async getLastForItem(itemId: string): Promise<ItemValidationGroup> {
+  async getLastForItem(itemId: string): Promise<ItemValidationGroup | null> {
     // return this.createQueryBuilder('iVG')
```
pyphilia, line 34: Can we check that `itemId` is not `undefined`? Otherwise typeorm could return everything... (I know the type is string but well...)

src/services/item/plugins/publication/validation/repositories/itemValidation.ts
```diff
@@ -1,22 +1,32 @@
...
+    if (!result) {
+      throw new ItemValidationNotFound(id);
+    }
+    return result;
+  }
 
   async getForGroup(groupId: string): Promise<ItemValidation[]> {
-    return this.findBy({ group: { id: groupId } });
-  },
+    return this.repository.findBy({ itemValidationGroup: { id: groupId } });
```
pyphilia, line 28: Same, can we check `groupId` is defined?

pyphilia (approved):
I swear we will merge it this week 😄

PR #1230, opened 2024-07-19
feat: allow publishing only for validated items

ReidyT:


src/services/item/plugins/publication/published/service.ts
```diff
@@ -122,20 +131,39 @@ export class ItemPublishedService {
...
+      case PublicationStatus.ItemTypeNotAllowed:
+        throw new ItemTypeNotAllowedToPublish(id, type);
+      case PublicationStatus.Published:
+      case PublicationStatus.PublishedChildren:
+        throw new ItemPublicationAlreadyExists(id);
+      case PublicationStatus.Unpublished:
+      case PublicationStatus.Pending:
+      case PublicationStatus.Invalid:
+      case PublicationStatus.Outdated:
+      case PublicationStatus.NotPublic:
```
spaenleh, line 150: Is this removed with the update of SDK ?

spaenleh (approved):
Looks good to me !

src/services/item/plugins/publication/published/service.ts
ReidyT: Yes, I have to remove it in this PR too ^^

PR #1232, opened 2024-07-23
feat: change last migration to set legacy data validation to true

pyphilia:
set legacy members as validated members. New members will be not validated.

WARNING: we need to redo the migration in dev


spaenleh (approved):

ReidyT (approved):
LGTM!

PR #1234, opened 2024-07-23
fix: remove mock for datasource

spaenleh:
In this PR I:
- remove the mock datasource 
- adapt the datasource file to be used in dev/prod and in the tests
- comment out all the calls to `jest.mock('../datasource')` 

This should help with keeping our maintenance low when adding entities we only have a single file to change. 
Also the differences between the two files were very small.

I added the `app.db.synchronize()` call to the `test/app.ts` so when running in test the DB is synchonized with schema drop. 

I am not sure where to put the `app.db.runMigrations()`. Maybe after the server has started ? or use a 2-stage startup where we only open the status route, run the migrations and once done, we open the trafic ? Maybe we need to discuss this a bit more.


src/plugins/datasource.ts
```diff
@@ -63,8 +66,9 @@ export const AppDataSource = new DataSource({
   poolSize: DB_CONNECTION_POOL_SIZE, // *2 because of the number of tasks
   // log queries that take more than 2s to execute
   maxQueryExecutionTime: 2000,
+  // needs to change based on where it runs
   logging: ['migration', 'error'],
-  migrationsRun: true,
+  // migrationsRun: true,
```
spaenleh, line 71: WARNING: Currently migrations will not be run automatically on start.

src/plugins/datasource.ts
```diff
@@ -63,8 +66,9 @@ export const AppDataSource = new DataSource({
   poolSize: DB_CONNECTION_POOL_SIZE, // *2 because of the number of tasks
   // log queries that take more than 2s to execute
   maxQueryExecutionTime: 2000,
+  // needs to change based on where it runs
   logging: ['migration', 'error'],
```
ReidyT, line 70: Should be make the same here ?

```js
export const DB_LOGGING = CI ? ['migration', 'error'] : [];
// ...
logging: DB_LOGGING
```

ReidyT (approved):
Nice changes, LGTM 🙌! Should we change something in the configuration files (for local or deployment) ?

src/plugins/datasource.ts
pyphilia: As long as we don't have a solution for migration, I would let it be, so in case of easy migrations it can run smoothly.

pyphilia (changes requested):

pyphilia (approved):

PR #1239, opened 2024-07-24
feat: modify get actions and get actions aggregate to have a start and end date

### LINKED ISSUE #1237: Update GET `/actions/aggregate`, `/actions` to receive start and end dates

LinaYahya:
closes #1237


src/services/action/repositories/action.ts
```diff
@@ -167,12 +191,25 @@ export class ActionRepository {
...
+    }
     // Second stage aggregation.
     const query = AppDataSource.createQueryBuilder()
       .from(`(${subquery.getQuery()})`, 'subquery')
       .setParameter('path', itemPath)
       .setParameter('view', view)
-      .setParameter('types', types);
+      .setParameter('types', types)
+      .setParameter('startDate', filters?.startDate)
+      .setParameter('endDate', filters?.endDate);
```
spaenleh, line 212: I am not sure you need to set these parameters ? Looks like you pass them already in the `andWhere`.

spaenleh (dismissed):
Thank you for the work. I think a small thing can be improved, otherwise it looks good to me. I let also @pyphilia weigh in.

src/services/action/repositories/action.ts
LinaYahya: It didn't work unless I passed them

src/services/item/plugins/action/test/index.test.ts
```diff
@@ -459,6 +459,33 @@ describe('Action Plugin Tests', () => {
       expect(parseFloat(response.json()[1]['aggregateResult'])).toBeCloseTo(4);
     });
```
pyphilia, line 461: Add or edit a test to check the default dates are used.

src/services/action/repositories/action.ts
```diff
@@ -99,7 +99,13 @@ export class ActionRepository {
    */
   async getForItem(
     itemPath: UUID,
-    filters?: { sampleSize?: number; view?: string; memberId?: UUID },
+    filters?: {
+      sampleSize?: number;
+      view?: string;
+      memberId?: UUID;
+      startDate?: string;
+      endDate?: string;
```
pyphilia, line 107: I think it needs default values. That would be nice to have one month to now as defaults.

src/services/action/repositories/action.ts
```diff
@@ -132,7 +150,13 @@ export class ActionRepository {
    */
   async getAggregationForItem(
     itemPath: UUID,
-    filters?: { sampleSize?: number; view?: string; types?: string[] },
+    filters?: {
+      sampleSize?: number;
+      view?: string;
+      types?: string[];
+      startDate?: string;
+      endDate?: string;
```
pyphilia, line 158: Set default values as mentioned above.

src/services/action/repositories/action.ts
```diff
@@ -167,12 +191,25 @@ export class ActionRepository {
...
+      subquery.andWhere('action.created_at >= :startDate', {
+        startDate: filters.startDate,
+      });
+    }
+
+    if (filters?.endDate) {
+      subquery.andWhere('action.created_at <= :endDate', {
+        endDate: filters.endDate,
+      });
+    }
```
pyphilia, line 204: I think we should check the validity of the range, eg startDate cannot be "higher" than `endDate`. Let's throw an error if it's not valid. And let's do it for both endpoints.

pyphilia (changes requested):
Thanks for the PR! I've added some comments let me know if you have questions! 🐩

pyphilia (approved):

PR #1244, opened 2024-07-25
feat: improve search for accessible and children

pyphilia:
Given keywords, search in many fields thanks to `search_document`.

### DIFF
--- a/src/services/item/repository.ts
+++ b/src/services/item/repository.ts
@@ -15,6 +15,7 @@ import {
 } from '@graasp/sdk';
 
 import { AppDataSource } from '../../plugins/datasource';
+import { ALLOWED_SEARCH_LANGS } from '../../utils/config';
 import {
   HierarchyTooDeep,
   InvalidMoveTarget,
@@ -24,7 +25,7 @@ import {
   UnexpectedError,
 } from '../../utils/errors';
 import { MemberIdentifierNotFound } from '../itemLogin/errors';
-import { Member } from '../member/entities/member';
+import { Actor, Member } from '../member/entities/member';
 import { itemSchema } from '../member/plugins/export-data/schemas/schemas';
 import { schemaToSelectMapper } from '../member/plugins/export-data/utils/selection.utils';
 import { mapById } from '../utils';
@@ -180,6 +181,7 @@ export class ItemRepository {
   }
 
   async getChildren(
+    actor: Actor,
     parent: Item,
     params?: ItemChildrenParams,
     options: { withOrder?: boolean } = {},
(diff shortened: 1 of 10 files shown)


src/services/item/repository.ts
```diff
@@ -198,6 +200,38 @@ export class ItemRepository {
...
+        });
+
+        // search by member lang
+        const memberLang = actor?.lang;
+        if (memberLang && memberLang != 'en' && ALLOWED_SEARCH_LANGS[memberLang]) {
+          q.orWhere('item.search_document @@ plainto_tsquery(:lang, :keywords)', {
+            keywords: keywordsString,
+            lang: ALLOWED_SEARCH_LANGS[memberLang],
+          });
+        }
```
ReidyT, line 231: I'm wondering if you really need the member for that. What about directly passing the language in parameter ?

src/services/item/repository.ts
```diff
@@ -198,6 +200,38 @@ export class ItemRepository {
       query.andWhere('item.type IN (:...types)', { types });
     }
 
+    const allKeywords = params?.keywords?.filter((s) => s && s.length);
+    if (allKeywords?.length) {
+      const keywordsString = allKeywords.join(' ');
+      query.andWhere((q) => {
+        // search in english by default
+        q.where("item.search_document @@ plainto_tsquery('english', :keywords)", {
```
ReidyT, line 208: I didn't remember how the search document works, but why do we need to search in english, using simple and using the member's language? Isn't it possible to only use the item's language for the search ? Also, do you know if adding multiple search conditions impact the performance

src/services/itemMembership/repository.ts
```diff
@@ -161,9 +162,35 @@ export const ItemMembershipRepository = AppDataSource.getRepository(ItemMembersh
...
+        });
+
+        // search by member lang
+        const memberLang = actor.lang;
+        if (memberLang != 'en' && ALLOWED_SEARCH_LANGS[memberLang]) {
+          q.orWhere('item.search_document @@ plainto_tsquery(:lang, :keywords)', {
+            keywords: keywordsString,
+            lang: ALLOWED_SEARCH_LANGS[memberLang],
+          });
+        }
```
ReidyT, line 193: same questions here ^^

ReidyT (approved):
Nice PR, LGTM! 👍 I just let some comments 😶‍🌫️

src/services/item/repository.ts
pyphilia: It works with stemmed vectors:
- english: if I search for "dogs", it will stem to "dog" and can search "dog" in items. Search for "chiens" will look for "chiens" as is.
- member lang: (let's say fr) if I search for "chiens", it will stem to "chien" and search for "chien" in items. Looking for "dogs" will look for "dogs" as is
- simple: if I search for "i am" it will look for "i am". However looking for "i am" in english will remove everything because they are stop words.
- raw ILIKE:  if I search for "www" it will look for "www". However looking for "www" in simple will not find "wwwffff" because vectors match on full words only.

So definitely adding lots of cases will impact performance. But as you see I cannot only look for english, or only for member's language. We can later tune if it impacts the performance too much.

I will add that to the specs.

src/services/item/repository.ts
pyphilia: We don't need the full member, but I prefer passing the full entity so it's similar to `accessible` + if one day we need an additional data, it's already there.

src/services/item/repository.ts
ReidyT: Okay I better understand thanks for your explanation ! I'm just not sure to understand why should we keep english and user's lang instead of only using item's lang (english if not set?). In this way if the item is written in English and set in English (in item settings), if the user:
- Search for "dogs", it will be stem to "dog" and search it in the item.
- Search for "chiens" it will search as it is in the item.

I think it's a good idea to keep simple and raw to handle the other cases.

Sorry if I misunderstood something, I think you're right, we can change it later if there is some performance issues.

PR #1245, opened 2024-07-26
fix: fix where scope on search

pyphilia:
The search scope lacks brackets, so it's overriding all other conditions. I've added the brackets to
- accessible
- children
- get from geolocation

with regression tests.


ReidyT (approved):
LGTM!

Issue #433, opened 2023-06-23
[import-export] `.graasp` extension confuses users


pyphilia:
Some users have reported being confused by the `.graasp` extension given to documents, which they believe to be proprietary (by default, it will not open with a web browser for instance).

Since the documents are serialized into HTML, we may consider using a `.html` extension, or a `.graaspdocument.html` so that we can recognize it as such when parsing the zip.

PR #1246, opened 2024-07-26
fix: allow to download raw file

### LINKED ISSUE #1130: [ZIP-Export] When downloading a single file, do not wrap it in a .zip
When downloading a single file it is wrapped in a .zip. This is good when downloading a folder, but for single files it is a hassle.

Detect when user asks for download of a single file and simply return the file as is.

pyphilia:
Allow to download single files without wrapping in a zip.

note: getFilenameFromItem is back in the backend because it is not necessary in the frontend anymore.

close #1130


src/services/item/plugins/importExport/index.ts
```diff
@@ -110,6 +104,33 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+
+        return stream || buffer;
+      }
+
+      // generate archive stream
+      const archiveStream = await importExportService.export(member, repositories, {
+        item,
+        reply,
+      });
+      console.log('gwrfsd');
```
spaenleh, line 132: remove ?

src/services/item/plugins/importExport/index.ts
```diff
@@ -81,7 +81,7 @@ const plugin: FastifyPluginAsync = async (fastify) => {
     },
   );
 
-  // download item as zip
+  // download item
   fastify.route<{ Params: { itemId: string } }>({
     method: 'GET',
     url: '/zip-export/:itemId',
```
ReidyT, line 87: Maybe we should change the path here ? Because it is not a zip anymore

src/services/item/plugins/importExport/index.ts
ReidyT, line 132:
```suggestion
```

src/services/item/plugins/importExport/service.ts
```diff
@@ -295,8 +310,19 @@ export class ImportExportService {
...
+    }
+
+    // save single item
+    const { stream, buffer, name } = await this.fetchItemData(actor, repositories, item);
+    if (stream) {
+      return archive.addReadStream(stream, path.join(archiveRootPath, name));
+    } else if (buffer) {
+      return archive.addBuffer(buffer, path.join(archiveRootPath, name));
+    } else {
+      return console.error('cannot fetch data for item', item.id, item.type);
```
ReidyT, line 325: Use logger here ?

src/services/item/plugins/importExport/service.ts
```diff
@@ -207,6 +206,68 @@ export class ImportExportService {
...
+      }
+      case isItemType(item, ItemType.APP): {
+        return {
+          buffer: Buffer.from(buildTextContent(item.extra.app?.url, ItemType.APP)),
+          name: getFilenameFromItem(item),
+          mimetype: 'text/plain',
+        };
+      }
+    }
+    throw new Error(`cannot fetch data for item ${item.id}`);
```
ReidyT, line 268: Should we use specialised error here ?

src/services/item/plugins/importExport/test/index.test.ts
```diff
@@ -237,7 +237,7 @@ describe('ZIP routes tests', () => {
       expect(response.headers['content-disposition']).toContain(item.name);
     });
 
-    it('Export successfully h5p file', async () => {
+    it.only('Export successfully h5p file', async () => {
```
ReidyT, line 240: remove only?

ReidyT (approved):
Thanks for this PR, LGTM! I just let some comments

src/services/item/plugins/importExport/service.test.ts
```diff
@@ -59,4 +60,192 @@ describe('ZIP routes tests', () => {
...
+        app.db,
+        {} as unknown as FileItemService,
+        resolveDependency(ItemService),
+        {} as unknown as H5PService,
+        resolveDependency(BaseLogger),
+      );
+      const repositories = buildRepositories();
+      const res = await importExportService.fetchItemData(actor, repositories, item);
+
+      expect(res.name).toEqual(item.name + '.graasp');
```
spaenleh, line 167: Interesting, so when the isRaw setting is true then we use the `.html` format, but when it is rich text, we use the `.graasp` format, even when it is also valid HTML. I am not sure how we could make it so that both documents can be easily opened on the computer and we are still able to re-import them correctly.

src/services/item/plugins/importExport/service.ts
```diff
@@ -207,6 +206,68 @@ export class ImportExportService {
     }
   }
 
+  async fetchItemData(
+    actor,
+    repositories,
+    item,
+  ): Promise<{ name: string; stream?: NodeJS.ReadableStream; buffer?: Buffer; mimetype: string }> {
```
spaenleh, line 213: We return either a stream of a buffer ? Would it be possible to return a stream in all cases ?

src/services/item/plugins/importExport/test/index.test.ts
spaenleh, line 240: remove `.only`

src/services/item/plugins/importExport/utils.test.ts
```diff
@@ -0,0 +1,109 @@
...
+  it('get file name from document item', () => {
+    const item = DocumentItemFactory({
+      name: 'mydoc',
+    });
+    expect(getFilenameFromItem(item)).toEqual('mydoc.graasp');
+    const item1 = DocumentItemFactory({
+      name: 'mydoc.graasp',
+    });
+    expect(getFilenameFromItem(item1)).toEqual('mydoc.graasp');
+  });
```
spaenleh, line 108: Should we add a test for when the document is in raw mode ?

src/services/item/plugins/importExport/utils.ts
```diff
@@ -50,3 +51,55 @@ export const buildTextContent = (url: string, type: UnionOfConst<typeof ItemType
...
+const extractExtension = ({ name, mimetype }: { name: string; mimetype?: string }): string => {
+  // slice to remove . character
+  const ext = path.extname(name).slice(1);
+  if (!ext && mimetype) {
+    return mime.extension(mimetype) || '';
+  }
+  return ext;
+};
+
+// use partial of item to be usable in backend
```
spaenleh, line 70: Adapt comment ?

src/services/item/plugins/importExport/utils.ts
```diff
@@ -50,3 +51,55 @@ export const buildTextContent = (url: string, type: UnionOfConst<typeof ItemType
...
+  mimetype?: string;
+}): string => {
+  switch (item.type) {
+    case ItemType.APP: {
+      return extractFileName(item.name, 'app');
+    }
+    case ItemType.DOCUMENT: {
+      if (item.mimetype === 'text/html') {
+        return extractFileName(item.name, 'html');
+      }
```
spaenleh, line 83: So it is not based on the `isRaw` setting ? The other test files makes it look it is.

spaenleh (changes requested):
Thank you for this PR ! I have left a few comments.

src/services/item/plugins/importExport/service.test.ts
pyphilia: Actually I could use html in any case. `.graasp` is legacy. Let's consider with this issue: https://github.com/graasp/graasp/issues/433

spaenleh (approved):
Very nice ! I like the refactor to use only streams, and some improvements you made which make the code a lot more clean and enjoyable to read. 

Also the route change is a nice addition I think !

PR #1247, opened 2024-07-26
feat: handle validation pending status

ReidyT:


src/services/item/plugins/publication/validation/contentModerator.ts
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  private async executeImageValidation(item: Item, actor: Actor): Promise<ValidationProcessResult> {
+    if (isItemType(item, ItemType.S3_FILE) || isItemType(item, ItemType.LOCAL_FILE)) {
+      const { path: filepath, mimetype } =
+        item.type === ItemType.S3_FILE ? item.extra.s3File : item.extra.file;
+
+      // if file is not an image, return success
+      if (!MimeTypes.isImage(mimetype)) {
+        return { status: ItemValidationStatus.Success };
```
ReidyT, line 73: Maybe ensure to never call if not an image ?

ReidyT: Also handle invalid images like svg in nudenet

src/services/item/plugins/publication/validation/errors.ts
```diff
@@ -96,3 +96,16 @@ export class ItemValidationNotFound extends GraaspValidationError {
     );
   }
 }
+
+export class ItemValidationAlreadyExist extends GraaspValidationError {
```
ReidyT, line 100: It is still used ?

src/services/item/plugins/publication/validation/index.ts
```diff
@@ -19,11 +21,20 @@ import { ItemPublishedService } from '../published/service';
 import { itemValidation, itemValidationGroup } from './schemas';
 import { ItemValidationService } from './service';
 
+const assertItemIsFolder = (item: Item): FolderItem => {
```
ReidyT, line 24: Move in utils?

src/services/item/plugins/publication/validation/service.ts
```diff
@@ -85,160 +61,61 @@ export class ItemValidationService {
...
-    actor: Member,
-    repositories: Repositories,
-    item: Item,
-    itemValidationGroup: ItemValidationGroup,
-  ): Promise<ItemValidationStatus[]> {
-    // execute each process on item
-    const results = (
+    const allItems = [item, ...descendants];
+    // Enqueues all items and return the added items
+    const inQueueItems = (
```
ReidyT, line 79:
```suggestion
    const enqueuedItems = (
```

src/services/item/plugins/publication/validation/utils.ts
```diff
@@ -1,9 +1,34 @@
...
+
 import { TMP_FOLDER } from '../../../../../utils/config';
+import { FolderItem, Item, isItemType } from '../../../entities/Item';
 
 export const stripHtml = (str?: string | null): string => (str ? striptags(str) : '');
 
 export const buildStoragePath = (itemId: string): string =>
   path.join(TMP_FOLDER, 'validations', itemId);
+
+export const assertItemIsFolder = (item: Item): FolderItem => {
```
Defferrard, line 14: I don't know what the opinion of the team about how we're initialising functions. I feel like currently we're using both methods.
Personally, I'd use the `function assertItemIsFolder(item: Item): FolderItem` because it is hoisted and more readable. I'm only opening the discussion, I don't think it is necessary to make a decision about it.

Defferrard (approved):
LGTM ! 💯

src/services/item/plugins/publication/validation/errors.ts
```diff
@@ -96,3 +96,16 @@ export class ItemValidationNotFound extends GraaspValidationError {
...
+  constructor(itemId: string) {
+    super(
+      {
+        code: 'GPVERR008',
+        statusCode: StatusCodes.CONFLICT,
+        message: `Item validation already exist for item ${itemId}`,
```
pyphilia, line 106:
```suggestion
        message: `Item validation already exists for item ${itemId}`,
```

src/services/item/plugins/publication/validation/validationQueue.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Returns the item id of the last item in the path.
+ * @param itemPath The item path to extract the id.
+ * @returns the id of the last item in the path.
+ */
+const getItemIdFromPath = (itemPath: string) => {
+  const itemIds = getIdsFromPath(itemPath);
+  const len = itemIds.length;
+  return itemIds[len - 1];
+};
```
pyphilia, line 22: I think we have sth like this in sdk already?  `getChildFromPath`

src/services/item/plugins/publication/validation/validationQueue.ts
```diff
@@ -0,0 +1,87 @@
...
+
+@singleton()
+export class ValidationQueue {
```
pyphilia, line 25: Not sure it's a queue, but I can guess that we want to have a real queue at some point. Maybe let's add a comment to explain that at the start of the class?

pyphilia: Also since it's called `validation` maybe we don't need to add `validation` in the names of the functions. Especially since they are not real validations? Maybe mention `progress`?

src/services/item/plugins/publication/validation/validationQueue.ts
```diff
@@ -0,0 +1,87 @@
...
+    const exists = await this.isValidationExists(itemPath);
+    if (exists) {
+      throw new ItemValidationAlreadyExist(getItemIdFromPath(itemPath));
+    }
+
+    // Only the root itemId is used to track validation progress.
+    const itemId = getItemIdFromPath(itemPath);
+
+    // Set the validation status for the root itemId in Redis.
+    await this.redis.set(buildRedisProgressKey(itemId), 1);
```
pyphilia, line 75: So the value is `1`? so we don't care about the value right?

src/services/item/plugins/publication/validation/contentModerator.ts
```diff
@@ -0,0 +1,180 @@
...
+  ) {
+    this.fileService = fileService;
+    this.imageClassifierApi = imageClassifierApi;
+  }
+
+  private isSameTypeAsFileService(item: Item) {
+    return item.type === this.fileService.fileType;
+  }
+
+  private async executeTextValidation(item: Item): Promise<ValidationProcessResult> {
```
pyphilia, line 57:
```suggestion
  private async executeItemTextValidation(item: Item): Promise<ValidationProcessResult> {
```

src/services/item/plugins/publication/validation/contentModerator.ts
```diff
@@ -0,0 +1,180 @@
...
+      { name: 'description', value: stripHtml(item.description) },
+    ]);
+    const result = suspiciousFields.join(', ');
+    const status =
+      suspiciousFields.length > 0 ? ItemValidationStatus.Failure : ItemValidationStatus.Success;
+
+    return { result, status };
+  }
+
+  private async executeImageValidation(item: Item, actor: Actor): Promise<ValidationProcessResult> {
```
pyphilia, line 69:
```suggestion
  private async executeImageItemValidation(item: Item, actor: Actor): Promise<ValidationProcessResult> {
```

src/services/item/plugins/publication/validation/utils.ts
pyphilia: I've always used arrow functions because it was a recommendation from eslint. But I'm open to any.

pyphilia (changes requested):
I really like the `ContentModerator` class, but would it be interesting to give it another name? What about `ItemValidationModerator`?

src/services/item/plugins/publication/validation/validationQueue.ts
ReidyT: Oh yeah, I didn't found it, thanks !

src/services/item/plugins/publication/validation/validationQueue.ts
ReidyT: Yeah, for now this class only handles the in progress validations, I didn't found a good name, do you have suggestions ?

src/services/item/plugins/publication/validation/validationQueue.ts
ReidyT: Yes exactly, I could write a comment for that ^^

src/services/item/plugins/publication/validation/itemValidationModerator.ts
```diff
@@ -0,0 +1,183 @@
...
+  status: ItemValidationStatus;
+  result?: string;
+};
+
+type ValidationStrategy = {
+  strategy: () => Promise<ValidationProcessResult>;
+  process: ItemValidationProcess;
+};
+
+export interface IContentModerator {
```
pyphilia, line 31: Rename this?

src/services/item/plugins/publication/validation/validationQueue.ts
pyphilia: Maybe we can rename the function (again, sorry) like `addInProgressIndication`?

pyphilia (approved):

src/services/item/plugins/publication/validation/utils.ts
Defferrard: Well, I feel like the behavior of the 2 usages is very different, we should probably choose the right initialisation dependently of the use case.
To me, arrows functions are great for callbacks and offers a useful implicit return, but I'd leave the rest to the verbose one. 

I also saw we can initialise anonymous functions that way : 
```typescript
const foo = function() {}
```
Which doesn't make much sense to me, it would better suit to arrow function.

ReidyT: I'm not sure that the behaviour is very different, except when using the `this` keyword. But if you have more information concerning the differences I'm really interested 😄.

Defferrard: All what has been mentioned earlier already makes a big difference to me. There's still subtil differences that I don't think we're concerned: (`call()`, `apply()`, `bind()`, `arguments array` ) 

Otherwise, we're having difficulties to profile our code because most of our functions are anonymous

ReidyT: Yes, I agree that there are differences, what I mean is that in our use, I haven't seen any limits to one compared to the other, except in the profiler because of the anonymous functions.

Again, I don't have any preference, so it's okay to me if we decide to always use functions, except when we have to use arrow function for some reason.

src/di/container.ts
```diff
@@ -152,6 +154,8 @@ export const registerDependencies = (instance: FastifyInstance) => {
 
   registerValue(ETHERPAD_NAME_FACTORY_DI_KEY, new RandomPadNameFactory());
 
+  registerValue(ItemValidationModerator, resolveDependency(ItemValidationModerator));
```
ReidyT, line 157: Not needed anymore

src/services/item/plugins/publication/validation/utils.ts
ReidyT: I'll resolve this conversation, but we can maybe discuss more about that if needed in a new issue ?

PR #1250, opened 2024-07-31
fix: log group in dev should be `ecs/graasp`

### LINKED ISSUE #1248: Definition of log group differs between IAC and dev in AWS.
In the IAC the log group for the backend is `ecs/graasp` but the task definition sets it in dev to `ecs/graasp-development`.

spaenleh:
The log group for dev is not the same as the one defined by the the infrastructure. I correct this in this PR.

fixes #1248


Defferrard (approved):
LGTM!

PR #1259, opened 2024-07-31
fix: remove deprecated packages

spaenleh:
Remove packages:
- **mime-types** package as it isnot necessary and is a duplicate from mime package
- **@types/striptags** as it is not required since the package has already types included
- **standard-version** since we are using release please


ReidyT (approved):
🚀

Defferrard (approved):
LGTM!

PR #1271, opened 2024-08-05
feat: storage details route

Defferrard:


src/services/item/repository.ts
```diff
@@ -644,6 +650,48 @@ export class ItemRepository {
...
+      updatedAt: item.item_updated_at,
+      size: JSON.parse(item.item_extra)[itemType].size,
+      path: JSON.parse(item.item_extra)[itemType].path,
+      parent: item.parent_id
+        ? {
+            id: item.parent_id,
+            name: item.parent_name,
+          }
+        : undefined,
+    }));
```
ReidyT, line 690: Should we define a type for that ?

src/services/member/controller.test.ts
```diff
@@ -185,4 +214,225 @@ describe('Member Controller', () => {
...
+    });
+    it('returns bad request when page size is lower than 1', async () => {
+      const response = await app.inject({
+        method: HttpMethod.Get,
+        url: '/members/current/storage/files',
+        query: { pageSize: FILE_METADATA_MIN_PAGE_SIZE - 1 + '' },
+      });
+      expect(response.statusCode).toBe(StatusCodes.BAD_REQUEST);
+    });
+    it('returns bad request when page size is greater than 50', async () => {
```
ReidyT, line 241: Not sure if it is working:

```suggestion
    it(`returns bad request when page size is greater than ${FILE_METADATA_MAX_PAGE_SIZE}`, async () => {
```

src/services/member/controller.test.ts
```diff
@@ -185,4 +214,225 @@ describe('Member Controller', () => {
...
+      const response = await app.inject({
+        method: HttpMethod.Get,
+        url: '/members/current/storage/files',
+        query: { pageSize: FILE_METADATA_MAX_PAGE_SIZE + 1 + '' },
+      });
+      expect(response.statusCode).toBe(StatusCodes.BAD_REQUEST);
+    });
+
+    it('without parent', async () => {
```
ReidyT, line 250: Maybe we can be more explicit in the test's name ?

src/services/member/controller.test.ts
```diff
@@ -185,4 +214,225 @@ describe('Member Controller', () => {
...
+        for (let i = 0; i < totalFiles; i++) {
+          await saveFile(member, rootFile);
+        }
+
+        for (let i = 0; i < 5; i++) {
+          await saveFile(anotherMember);
+        }
+      });
+
+      it('default', async () => {
```
ReidyT, line 288:
```suggestion
      it('default parameters when not specified', async () => {
```

src/services/member/controller.ts
```diff
@@ -51,6 +59,32 @@ const controller: FastifyPluginAsync = async (fastify) => {
...
+        return;
+      }
+      const member = notUndefined(user?.member);
+      return {
+        ...(await storageService.getStorageFilesMetadata(
+          member,
+          fileService.fileType,
+          { page, pageSize },
+          buildRepositories(),
+        )),
```
ReidyT, line 82: Maybe this is more readable ?

```suggestion
      const storageFilesMetadata = await storageService.getStorageFilesMetadata(
          member,
          fileService.fileType,
          { page, pageSize },
          buildRepositories(),
        );
      return {
        ...storageFilesMetadata,
```

ReidyT (approved):
Thanks for this PR, LGTM ! I just let some comments 😄

src/services/member/controller.test.ts
Defferrard: Having variables in the test name makes it difficult to be runned independently from command line or Jest Runner.
What about `returns bad request when page size is greater than the maximum` ?

ReidyT: Okay I understand. Yeah I think it's good :). It was okay with the `50`, my only issue is if we change the limit we have to remember to change the test name too. Not a big deal, so it's as you want.

Defferrard: Yeah, I agree with you about the number!

package.json
```diff
@@ -60,7 +60,7 @@
     "@fastify/static": "7.0.4",
     "@fastify/websocket": "7.2.0",
     "@graasp/etherpad-api": "2.1.1",
-    "@graasp/sdk": "4.22.0",
+    "@graasp/sdk": "graasp/graasp-sdk#611-add-file-metadata-type-for-storage-route",
```
pyphilia, line 63: Update!

src/services/member/controller.ts
```diff
@@ -51,6 +59,33 @@ const controller: FastifyPluginAsync = async (fastify) => {
...
+        reply.status(StatusCodes.BAD_REQUEST).send();
+        return;
+      }
+      const member = notUndefined(user?.member);
+      const storageFilesMetadata = await storageService.getStorageFilesMetadata(
+        member,
+        fileService.fileType,
+        { page, pageSize },
+        buildRepositories(),
+      );
```
pyphilia, line 81: I usually try to keep `actor` as first arg, then `repositories` as 2nd arg for some kind of consistency.

src/services/member/schemas.ts
```diff
@@ -121,6 +121,54 @@ export const getStorage: FastifySchema = {
   },
 };
 
+export const getStorageFiles: FastifySchema = {
+  querystring: {
+    type: 'object',
+    properties: {
+      page: { type: 'number', default: 1 },
+      pageSize: { type: 'number', default: 10 },
```
pyphilia, line 129: The default here is not the same as in `src/services/member/constants.ts`? Should we use the constants here?

pyphilia (approved):
LGTM, I've added some minor comments, thanks for this PR! 📗

mariembencheikh (approved):
Good work ! Thank you !

PR #1272, opened 2024-08-05
feat: convert svg to png before sending to nudenet

ReidyT:


src/services/item/plugins/publication/validation/utils.ts
```diff
@@ -1,9 +1,37 @@
...
+};
+
+export const isFileType = (item: Item) => {
+  return isItemType(item, ItemType.S3_FILE) || isItemType(item, ItemType.LOCAL_FILE);
+};
+
+export const getMimeType = (item: Item<'s3File'> | Item<'file'>) => {
+  const { mimetype } = item.type === ItemType.S3_FILE ? item.extra.s3File : item.extra.file;
+  return mimetype;
+};
```
Defferrard, line 29: It is only a suggestion, I'm not sure if it is conventional or not. We're probably losing type safety ?
 
```suggestion
export const getMimeType = (item: Item<'s3File'> | Item<'file'>) => {
  const { mimetype } = item.extra[item.type];
  return mimetype;
};
```

Defferrard (approved):
LGTM !  Good work !

src/services/item/plugins/publication/validation/utils.ts
ReidyT: I really like the idea, but we are losing the type, so I think I'll let like it is.

src/services/item/plugins/publication/validation/constants.ts
```diff
@@ -1,3 +1,5 @@
 export const PLUGIN_NAME = 'graasp-plugin-validation';
 
 export const IMAGE_CLASSIFIER_PREDICTION_THRESHOLD = 0.35;
+
+export const SVG_MIMETYPE = 'image/svg+xml';
```
pyphilia, line 5: It already exists in sdk maybe we can reuse? https://github.com/graasp/graasp-sdk/blob/6cccfb970eaa55c93a5473323e14d93cbab9cb3e/src/mimeTypes/mimeTypes.ts#L38

src/services/item/plugins/publication/validation/utils.ts
```diff
@@ -23,12 +23,15 @@ export const isFileType = (item: Item) => {
   return isItemType(item, ItemType.S3_FILE) || isItemType(item, ItemType.LOCAL_FILE);
 };
 
+export const getMimeType = (item: Item<'s3File'> | Item<'file'>) => {
```
pyphilia, line 26: Could we use `getMimetype` of sdk?

pyphilia (approved):
LGTM really good reusage of code, I think you can even use some stuff from sdk to reduce even more the code 🐞

src/services/item/plugins/publication/validation/utils.ts
ReidyT: Great idea 🤩

src/services/item/plugins/publication/validation/constants.ts
ReidyT: Yes good suggestion ! I saw and I didn't do that because I had to export it... but I think we can reuse it 🙌

ReidyT: Oh I just realised that we didn't have to export the const of images ^^ So good suggestion 🙌

PR #1274, opened 2024-08-05
feat: update invite invitation to include membership creation

### LINKED ISSUE #1273: update invitation endpoint to create membership if member exists

pyphilia:
close #1273


src/services/item/plugins/invitation/service.ts
```diff
@@ -223,6 +224,34 @@ export class InvitationService {
...
+    itemId: Item['id'],
+    invitations: Pick<Invitation, 'email' | 'permission'>[],
+  ): Promise<{ memberships: ItemMembership[]; invitations: Invitation[] }> {
+    await this.itemService.get(actor, repositories, itemId, PermissionLevel.Admin);
+
+    if (!invitations.length) {
+      throw new NoInvitationReceivedFound();
+    }
+
+    return this._createMembershipsAndInvitationsForUserList(
```
ReidyT, line 247: Not sure but maybe I would split `_createMembershipsAndInvitationsForUserList` into two functions (createMemberships and createInvitation) and use `_partitionExistingUsersAndNewUsers` directly here.

src/services/item/plugins/invitation/service.ts
```diff
@@ -223,6 +224,34 @@ export class InvitationService {
...
+    actor: Member,
+    repositories: Repositories,
+    itemId: Item['id'],
+    invitations: Pick<Invitation, 'email' | 'permission'>[],
+  ): Promise<{ memberships: ItemMembership[]; invitations: Invitation[] }> {
+    await this.itemService.get(actor, repositories, itemId, PermissionLevel.Admin);
+
+    if (!invitations.length) {
+      throw new NoInvitationReceivedFound();
+    }
```
Defferrard, line 245: We should check if invitations is empty in the controller instead of in the service. It could prevent the `buildRepository` call and the creation of a SQL Transaction.

This is only a suggestion but if we consider the call of this function with empty invitations as bad, we can prevent it with a custom type (source : https://dev.to/martinpersson/type-safety-with-non-empty-arrays-in-typescript-k78)

```typescript
type NonEmptyArray<T> = [T, ...T[]]
function isNonEmptyArray<T>(arr: T[]): arr is NonEmptyArray<T>{
    return arr.length > 0
}

function func(_: NonEmptyArray<number>){

}

func([]) // WRONG
func([1]) // OK
func([1,2,3]) // OK

function sample(arr: number[]){
    func(arr); // WRONG
    if(arr.length>0){
        func(arr); // WRONG
    }
    if(isNonEmptyArray(arr)){
        func(arr); // OK
    }
}
```

```suggestion
  async shareItem(
    actor: Member,
    repositories: Repositories,
    itemId: Item['id'],
    invitations: NonEmptyArray<Pick<Invitation, 'email' | 'permission'>>,
  ): Promise<{ memberships: ItemMembership[]; invitations: Invitation[] }> {
    await this.itemService.get(actor, repositories, itemId, PermissionLevel.Admin);

```

src/services/item/plugins/invitation/errors.ts
```diff
@@ -176,3 +176,13 @@ export class TemplateItemDoesNotExist extends GraaspInvitationsError {
     );
   }
 }
+
+export class NoInvitationReceivedFound extends GraaspInvitationsError {
+  constructor() {
+    super({
+      code: 'GPINVERR015',
+      statusCode: StatusCodes.BAD_REQUEST,
+      message: 'No invitations received',
```
Defferrard, line 185: Should we create an entry in Graasp translation for this error ?

src/services/item/plugins/invitation/schema.ts
```diff
@@ -62,18 +62,42 @@ export const invite = {
...
+          items: {
+            $ref: 'https://graasp.org/invitations/#/definitions/invitation',
+          },
+        },
       },
     },
+    '4xx': {
+      $ref: 'https://graasp.org/#/definitions/error',
+    },
+    500: {
```
Defferrard, line 98: Currently I saw different way to write the HTTP Status Code. What do you think about this one ?
```suggestion
   [StatusCodes.INTERNAL_SERVER_ERROR]: {
```

Defferrard (approved):
Good work 🔥 LGTM ! I just have some suggestions

src/services/item/plugins/invitation/service.ts
```diff
@@ -223,6 +224,34 @@ export class InvitationService {
...
+   * @param invitations
+   * @returns
+   */
+  async shareItem(
+    actor: Member,
+    repositories: Repositories,
+    itemId: Item['id'],
+    invitations: Pick<Invitation, 'email' | 'permission'>[],
+  ): Promise<{ memberships: ItemMembership[]; invitations: Invitation[] }> {
+    await this.itemService.get(actor, repositories, itemId, PermissionLevel.Admin);
```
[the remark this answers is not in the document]
pyphilia: Actually it does it in this function already.

src/services/item/plugins/invitation/service.ts
ReidyT: Yes, it's because for me `_createMembershipsAndInvitationsForUserList` as two responsibilities as suggested in its name, it create memberships and invitations. Now if `shareItem` partitions the data and **delegates** the creation of members and invitations to the dedicated functions, it responsibility is only to orchestrate these creations. This could also be done by `_createMembershipsAndInvitationsForUserList` but in my sense this function is useless with these modifications. Otherwise, I have the feeling that `shareItem` is useless, because it does nothing special except calling `_createMembershipsAndInvitationsForUserList` (mostly).

But it's only my point of view 😆

pyphilia: I see, I think `_createMembershipsAndInvitationsForUserList ` can be used instead of `service.shareItem`, but the admin access check will then have to happen in the controller, and I'm not really sure of that. So I'll let the implementation like this, we can refactor when we refactor controllers!

Issue #1220, opened 2024-07-18
request membership to admin


pyphilia:
- request
  - allow member that has no membership on an item to request an access
  - member cannot request if he already has a membership (any permission)
    - still pending
    - was refused (frontend: not show item and not show request form)
`POST  /items/:item-id/memberships/requests`

- validate
  - admin can validate the request (define permission here) ------ **REUSE POST /item-memberships**
    - a membership has been created for the member
    - the corresponding request is deleted
    - member will receive an email notification
`POST /items/:item-id/memberships`
body: { memberid, permission: PermissionLevel = 'read' }

- admin can refuse the request (delete request)
`DELETE /items/:item-id/memberships/requests/:request-id`

- admin can see all requests for an item 
`GET /items/:item-id/memberships/requests`
  - pagination, default order createdAt DESC

- member can see its request for given item (frontend: if has submitted a pending/deleted request )
`GET /items/:item-id/memberships/requests/own`

New entity
**MembershipRequest**
- id
- member id
- item id
- createdAt

------------ 

in case we want to implement block feature
- status: pending/blocked (frontend: blocked 🚫 ) - decline delete request
  - pending: admin didn't refuse nor accept 
  - blocked: refused and user cannot request again (don't show form)
needs an endpoint to patch request

# Implementation Steps
- [x] #1280
- [x] #1286
- [x] #1291
- [x] #1292
- [x] #1298

Defferrard:
- [ ] Create MembershipRequest Entity
- [ ] MembershipRequest creation API Route
- [ ] MembershipRequest visualisation API Route
- [ ] MembershipRequest refuse API Route
- [ ] Own MembershipRequest status visualisation API Route

Defferrard:
To prevent abuse, we've defined that the admin is responsible for determining which permissions to set. This limits the damage in case of errors.

PR #1281, opened 2024-08-07
feat(membership request): create entity

Defferrard:
This pull request is a part of the main issue https://github.com/graasp/graasp/issues/1220.
I've divided it, to make it easier to review and having continuous feedback

The goal of this PR is only to create the corresponding Entity and the structure of the feature. This PR should have no impact on the application.

### DIFF
--- a/src/services/itemMembership/plugins/MembershipRequest/entities/MembershipRequest.ts
+++ b/src/services/itemMembership/plugins/MembershipRequest/entities/MembershipRequest.ts
@@ -1,0 +1,36 @@
+import {
+  BaseEntity,
+  CreateDateColumn,
+  Entity,
+  JoinColumn,
+  ManyToOne,
+  PrimaryGeneratedColumn,
+  Unique,
+} from 'typeorm';
+
+import { Item } from '../../../../item/entities/Item';
+import { Member } from '../../../../member/entities/member';
+
+@Entity()
+@Unique('UQ_membership_request_item-member', ['item', 'member'])
+export class MembershipRequest extends BaseEntity {
+  @PrimaryGeneratedColumn('uuid', { primaryKeyConstraintName: 'PK_membership_request_id' })
+  id: string;
+
+  @ManyToOne(() => Member, (member) => member.id, {
+    onDelete: 'CASCADE',
+    nullable: false,
+  })
+  @JoinColumn({ name: 'member_id', foreignKeyConstraintName: 'FK_membership_request_member_id' })
+  member: Member;
+
+  @ManyToOne(() => Item, (item) => item.id, {
+    onDelete: 'CASCADE',
+    nullable: false,
+  })
+  @JoinColumn({ name: 'item_id', foreignKeyConstraintName: 'FK_membership_request_item_id' })
+  item: Item;
+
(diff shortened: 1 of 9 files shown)


src/services/itemMembership/plugins/MembershipRequest/entities/MembershipRequest.ts
ReidyT, line 1: Isn't the permission missing in this entity?

src/services/itemMembership/plugins/MembershipRequest/entities/MembershipRequest.ts
```diff
@@ -0,0 +1,40 @@
...
+  PrimaryGeneratedColumn,
+  Unique,
+  UpdateDateColumn,
+} from 'typeorm';
+
+import { Item } from '../../../../item/entities/Item';
+import { Member } from '../../../../member/entities/member';
+
+@Entity()
+@Unique('UQ_membership_request_item-member', ['item', 'member'])
```
ReidyT, line 16: I really like the fact that we are naming the constraints and foreign key ! Do you think it is possible to also name PK and Idx ?

src/services/itemMembership/plugins/MembershipRequest/schemas.ts
ReidyT, line 1: I think the permission is missing here too ? Or it is the admin who decide the permission ?

ReidyT (approved):
Great work ! LGTM! I just let some questions 😶‍🌫️

src/services/itemMembership/plugins/MembershipRequest/schemas.ts
ReidyT: Perfect, I think it is a really good decision! Maybe we just have to update it in the #1220 issue ?

src/services/itemMembership/plugins/MembershipRequest/entities/MembershipRequest.ts
Defferrard: Yes we can, I'll name the PK!

pyphilia (approved):
Nice PR, I just have a question about a column. Otherwise I think it's good to divide the work in many PRs, however since we deploy on dev on merge, you will have to write migrations if you make any change in between. Or we manually down the previous migration, which also a solution (we usually do it if we tested the migration before merging).

src/services/itemMembership/plugins/MembershipRequest/entities/MembershipRequest.ts
```diff
@@ -0,0 +1,40 @@
...
+  })
+  @JoinColumn({ name: 'item_id', foreignKeyConstraintName: 'FK_membership_request_item_id' })
+  item: Item;
+
+  @CreateDateColumn({ name: 'created_at', nullable: false })
+  createdAt: Date;
+
+  @UpdateDateColumn({ name: 'updated_at', nullable: false })
+  updatedAt: Date;
+}
```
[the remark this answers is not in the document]
Defferrard: It was for the block feature. If we decide to set a state on the request. However, I agree it has no use right now so I can remove it !

PR #1287, opened 2024-08-08
refactor(repository): converts AppActionRepository to AbstractRepository

ReidyT:


src/repositories/repository.ts
```diff
@@ -0,0 +1,75 @@
...
+
+import { AppDataSource } from '../plugins/datasource';
+import { IdPkBaseEntity } from '../services/item/plugins/app/appAction/appAction';
+import {
+  EntryNotFoundAfterInsertException,
+  IllegalArgumentException,
+  InsertionException,
+} from './errors';
+
+export abstract class AbstractRepository<T extends BaseEntity> {
```
pyphilia, line 19: Should all those repositories have their own file? Also I was wondering if we should go for the file naming convention of using a capital letter? ie. `../../ReadRepository.ts`?

pyphilia (commented):
I think it's really interesting. However I wonder how it would fit with more complex repositories, because `AppAction` is quite straightforward. Do you think you could try out on another repository? I think of `item` but maybe it's too big.

src/repositories/repository.ts
ReidyT: I think it's a good idea to have one repo per file! concerning the convention, has always I think what matters is that the whole team accepts and follows the convention. Personally, I like the `PascalCase` convention, but I'm open to any other file convention like the `kebab-case` or `camelCase`.

ReidyT:
> I think it's really interesting. However I wonder how it would fit with more complex repositories, because `AppAction` is quite straightforward. Do you think you could try out on another repository? I think of `item` but maybe it's too big.

There are two goals here: trying to standardise the share methods signature in the different concrete repositories (like item, appAction, and other custom entity repositories) and standardise the behaviour of the standard methods (implementation in the concrete repositories).

The first goal, achieved by using interfaces should allow us to enforce that we could always use our repositories the same way for searching, creating, updating or removing one element. In this way, will always know that if we want to search one element, in all our repositories (at least those who implement the search feature), the method's name is findOne (for example, the name is not defined now) and we always pass one argument, which is the primary key --> `anyRepository.findOne(id)`. We also know that the findOne method could always return `Entity | null` if we decide that the repo return null if not found instead of throwing an exception.

The second purpose, harmonising the behaviour, allows us to ensure that each time we implement the `findOne` method (for example), by delegating the find to the `readRepository`, we know that it will throw if the given pk value is null (avoid the issue where TypeORM returns absolutely all the rows!) and we are also sure that if the value is not found, null is returned instead of throwing.

I'm reexplaining here to clarify myself and to try to highlight that it should not cause issues with more complex repositories because we are just indicating the signature of the common methods that the client code will use and the common behaviours for simple request like the basic CRUD. That means, for example, the `getDescendant` of the item's repository will not be affected by this kind of refactor.

Of course, it's more than possible that I've forgot something and we could encounter issues, so I'm open to discussion to found issues earlier or improvements of the code 😄.

I don't know what do you think @pyphilia, @Defferrard ? Sorry for the long comment by the way ^^.

src/repositories/repository.ts
Defferrard: Personnally, when the file is intented to be a class file, I'd prefer PascalCase. However, I'm okay with any other convention !

Defferrard (commented):
To be honest, I feel like this PR is adding a way too much complexity to our Repository since it should be simple.

Here is the simple class diagram I've made from what I've understood :
```mermaid
classDiagram
    AppActionRepository *-- ReadRepository
    AppActionRepository *-- CreateRepository
    AppActionRepository <|-- AbstractRepository
    CreateRepository *-- ReadRepository
    CreateRepository <|-- AbstractRepository
    ReadRepository <|-- AbstractRepository

    AppActionRepository ..|> PostRepository
```

If the result is intended, I think we should probably at least rename some of our class, because everything is a Repository now. I'm also scared that we're gonna create too much TypeORM repository that way.

Except having an AbstractCRUDRepository that have all the necessary functions needed even the functions we don't need, I don't have any better solution...

ReidyT:
> To be honest, I feel like this PR is adding a way too much complexity to our Repository since it should be simple.
> 
> Here is the simple class diagram I've made from what I've understood :
> 
> ```mermaid
> classDiagram
>     AppActionRepository *-- ReadRepository
>     AppActionRepository *-- CreateRepository
>     AppActionRepository <|-- AbstractRepository
>     CreateRepository *-- ReadRepository
>     CreateRepository <|-- AbstractRepository
>     ReadRepository <|-- AbstractRepository
> 
>     AppActionRepository ..|> PostRepository
> ```
> If the result is intended, I think we should probably at least rename some of our class, because everything is a Repository now. I'm also scared that we're gonna create too much TypeORM repository that way.
> 
> Except having an AbstractCRUDRepository that have all the necessary functions needed even the functions we don't need, I don't have any better solution...

Thanks @Defferrard for your answer! I totally understand your comment. I also thought about the inheritance of the `AbstractCRUDRepository`, but my issue with that is we are forcing to expose the CRUD operations, even for immutable data as it is for the `actions`. I don't really know how many repositories are not a CRUD, maybe not enough to add the complexity I proposed here...

I tried to refactor in this PR because since the refactor to abstract repository, I have to check that the inserted data exist in the database, otherwise the function could return `Entity | null`. To avoid repeating my self in the futur refactors of the other repositories, I thought about a potential solution (which could be not the good one ^^).

Maybe another potential solution could be to have an EntityRepository (like the ItemRepository) composed with a CRUDRepository (the merge of my new Read and create repositories). In that way, we only have one new class repository (the CRUD) instead of four (CreateRepository, ReadRepository... ). With that we could keep the common interfaces Read, Create, Update and Delete, and maybe having an interface which is the merge of the four (CRUD interface). With that, a repository who only want to allow create and read could use the appropriate methods of the CRUDRepository and only exposes the create and read methods by implementing the corresponding interfaces. On the other hand, a repository who want to allow CRUD (like ItemRepository), could exploits all the methods of the CRUDRepository and implements the CRUD interface.

What do you think ?

Defferrard:
> > To be honest, I feel like this PR is adding a way too much complexity to our Repository since it should be simple.
> > Here is the simple class diagram I've made from what I've understood :
> > 
> >   ```mermaid
> > classDiagram
> >     AppActionRepository *-- ReadRepository
> >     AppActionRepository *-- CreateRepository
> >     AppActionRepository <|-- AbstractRepository
> >     CreateRepository *-- ReadRepository
> >     CreateRepository <|-- AbstractRepository
> >     ReadRepository <|-- AbstractRepository
> > 
> >     AppActionRepository ..|> PostRepository
> > ```
> >   
> >     
> >   
> >   
> >   Loading
> > 
> >   
> > 
> >       classDiagram
> >     AppActionRepository *-- ReadRepository
> >     AppActionRepository *-- CreateRepository
> >     AppActionRepository <|-- AbstractRepository
> >     CreateRepository *-- ReadRepository
> >     CreateRepository <|-- AbstractRepository
> >     ReadRepository <|-- AbstractRepository
> > 
> >     AppActionRepository ..|> PostRepository
> > 
> >     
> > 
> > If the result is intended, I think we should probably at least rename some of our class, because everything is a Repository now. I'm also scared that we're gonna create too much TypeORM repository that way.
> > Except having an AbstractCRUDRepository that have all the necessary functions needed even the functions we don't need, I don't have any better solution...
> 
> Thanks @Defferrard for your answer! I totally understand your comment. I also thought about the inheritance of the `AbstractCRUDRepository`, but my issue with that is we are forcing to expose the CRUD operations, even for immutable data as it is for the `actions`. I don't really know how many repositories are not a CRUD, maybe not enough to add the complexity I proposed here...
> 
> I tried to refactor in this PR because since the refactor to abstract repository, I have to check that the inserted data exist in the database, otherwise the function could return `Entity | null`. To avoid repeating my self in the futur refactors of the other repositories, I thought about a potential solution (which could be not the good one ^^).
> 
> Maybe another potential solution could be to have an EntityRepository (like the ItemRepository) composed with a CRUDRepository (the merge of my new Read and create repositories). In that way, we only have one new class repository (the CRUD) instead of four (CreateRepository, ReadRepository... ). With that we could keep the common interfaces Read, Create, Update and Delete, and maybe having an interface which is the merge of the four (CRUD interface). With that, a repository who only want to allow create and read could use the appropriate methods of the CRUDRepository and only exposes the create and read methods by implementing the corresponding interfaces. On the other hand, a repository who want to allow CRUD (like ItemRepository), could exploits all the methods of the CRUDRepository and implements the CRUD interface.
> 
> What do you think ?

I agree with your last proposal ! Seems good to me !


From your explanation I've also thought about that implementation : 
```mermaid
classDiagram
    class ReadableRepository {
        findOne()
        findAll()
    }

    class ImmutableRepository {
        create()
    }

    class CRUDRepository {
        edit()
        delete()
    }

    ReadableRepository --|> AbstractRepository
    ImmutableRepository --|> ReadableRepository
    CRUDRepository --|> ImmutableRepository

    AppActionRepository ..|> ImmutableRepository

    ItemRepository ..|> CRUDRepository
```

I think it is more OOP but can also be hard to maintain I guess..
To me, both solutions are equivalent, so I don't have any preference.

What do you think ?

ReidyT:
> > > To be honest, I feel like this PR is adding a way too much complexity to our Repository since it should be simple.
> > > Here is the simple class diagram I've made from what I've understood :
> > > 
> > >   ```mermaid
> > > classDiagram
> > >   AppActionRepository *-- ReadRepository
> > >   AppActionRepository *-- CreateRepository
> > >   AppActionRepository <|-- AbstractRepository
> > >   CreateRepository *-- ReadRepository
> > >   CreateRepository <|-- AbstractRepository
> > >   ReadRepository <|-- AbstractRepository
> > > 
> > >   AppActionRepository ..|> PostRepository
> > > ```
> > >   
> > >     
> > >   
> > >   
> > >   Loading
> > > 
> > >   
> > > 
> > >       classDiagram
> > >   AppActionRepository *-- ReadRepository
> > >   AppActionRepository *-- CreateRepository
> > >   AppActionRepository <|-- AbstractRepository
> > >   CreateRepository *-- ReadRepository
> > >   CreateRepository <|-- AbstractRepository
> > >   ReadRepository <|-- AbstractRepository
> > > 
> > >   AppActionRepository ..|> PostRepository
> > > 
> > >     
> > > 
> > > Loading
> > > ```
> > >   classDiagram
> > > AppActionRepository *-- ReadRepository
> > > AppActionRepository *-- CreateRepository
> > > AppActionRepository <|-- AbstractRepository
> > > CreateRepository *-- ReadRepository
> > > CreateRepository <|-- AbstractRepository
> > > ReadRepository <|-- AbstractRepository
> > > 
> > > AppActionRepository ..|> PostRepository
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > If the result is intended, I think we should probably at least rename some of our class, because everything is a Repository now. I'm also scared that we're gonna create too much TypeORM repository that way.
> > > Except having an AbstractCRUDRepository that have all the necessary functions needed even the functions we don't need, I don't have any better solution...
> > 
> > 
> > Thanks @Defferrard for your answer! I totally understand your comment. I also thought about the inheritance of the `AbstractCRUDRepository`, but my issue with that is we are forcing to expose the CRUD operations, even for immutable data as it is for the `actions`. I don't really know how many repositories are not a CRUD, maybe not enough to add the complexity I proposed here...
> > I tried to refactor in this PR because since the refactor to abstract repository, I have to check that the inserted data exist in the database, otherwise the function could return `Entity | null`. To avoid repeating my self in the futur refactors of the other repositories, I thought about a potential solution (which could be not the good one ^^).
> > Maybe another potential solution could be to have an EntityRepository (like the ItemRepository) composed with a CRUDRepository (the merge of my new Read and create repositories). In that way, we only have one new class repository (the CRUD) instead of four (CreateRepository, ReadRepository... ). With that we could keep the common interfaces Read, Create, Update and Delete, and maybe having an interface which is the merge of the four (CRUD interface). With that, a repository who only want to allow create and read could use the appropriate methods of the CRUDRepository and only exposes the create and read methods by implementing the corresponding interfaces. On the other hand, a repository who want to allow CRUD (like ItemRepository), could exploits all the methods of the CRUDRepository and implements the CRUD interface.
> > What do you think ?
> 
> I agree with your last proposal ! Seems good to me !
> 
> From your explanation I've also thought about that implementation :
> 
> ```mermaid
> classDiagram
>     class ReadableRepository {
>         findOne()
>         findAll()
>     }
> 
>     class ImmutableRepository {
>         create()
>     }
> 
>     class CRUDRepository {
>         edit()
>         delete()
>     }
> 
>     ReadableRepository --|> AbstractRepository
>     ImmutableRepository --|> ReadableRepository
>     CRUDRepository --|> ImmutableRepository
> 
>     AppActionRepository ..|> ImmutableRepository
> 
>     ItemRepository ..|> CRUDRepository
> ```
> I think it is more OOP but can also be hard to maintain I guess.. To me, both solutions are equivalent, so I don't have any preference.
> 
> What do you think ?

Mmmh I think your proposal is very interesting ! I'm just not sure if the diagram is correct, I think all the relations should be inheritance and not implementation ?

I know that I made the same in my proposal, but maybe we could merge `Readable` and `Immutable` repositories. Indeed, generally, if we want to read, we have to create the data first.

I think your solution has the advantage to be more simple than mine, and thats a good point. On the other hand, your proposal is less flexible than mine, in the sense if we want to add a feature in the `Readable`, all the subclass will have to implement it. With the composition, only the class who are interested by the new feature will have the possibility to implement it, but at the cost to more boilerplate code.

To me the two solutions are interesting and have their own advantages and disadvantages.

src/repositories/ImmutableRepository.ts
```diff
@@ -0,0 +1,124 @@
...
+    return await this.findOne(primaryKey);
+  }
+
+  /********************************** Subclass Methods **********************************/
+  /**
+   * Throws an IllegalArgumentException if the given primary key is undefined or an empty array.
+   *
+   * @param primaryKey The primary key to check.
+   */
+  protected throwsIfPKIsUndefined(primaryKey: string | string[]) {
```
Defferrard, line 66: I'm not sure if the name of this function is enough accurate since it also throw if the PK array is empty.
What if a repository want to override this function so it can check the string respect an email format ?

What do you think about this one ? 
```suggestion
  protected throwsIfPKIsInvalid(primaryKey: string | string[]) {
```

src/repositories/ImmutableRepository.ts
```diff
@@ -0,0 +1,124 @@
...
+   *
+   * @param entity The entity data to insert.
+   * @param relations Optional relations to load with the inserted entity.
+   * @throws EntryNotFoundAfterInsertException if the inserted entity is not found.
+   * @returns A promise that resolves to the inserted entity.
+   */
+  protected async insert(entity: DeepPartial<T>, relations?: FindOptionsRelations<T>): Promise<T> {
+    try {
+      const insertResult = await this.repository.insert(entity as QueryDeepPartialEntity<T>);
+      const insertedEntity = await this.findOne(insertResult.identifiers[0].id, { relations });
```
Defferrard, line 112: I'm probably a bit anxious, but I'm worried about : What if insertResult.identifiers.length is equals to 0 ?
It would throw an OutOfBand exception, that is catched and rethrowed as an InsertionException.

This behavior seems okay to me !

src/repositories/ImmutableRepository.ts
```diff
@@ -0,0 +1,124 @@
...
+
+      // Should never happen, if an error occurs, it should throw during the insert.
+      if (!insertedEntity) {
+        throw new EntryNotFoundAfterInsertException(this.entity);
+      }
+
+      return insertedEntity;
+    } catch (e) {
+      throw new InsertionException(e);
+    }
```
Defferrard, line 122: If insertedEntity is effectively `null`, which should not happen, then the double throw is unnecessary. Unless you want to nest the exception inside this `InsertionException`. What do you think about this one ?

```suggestion
    try {
      const insertResult = await this.repository.insert(entity as QueryDeepPartialEntity<T>);
      const insertedEntity = await this.findOne(insertResult.identifiers[0].id, { relations });
    } catch (e) {
      throw new InsertionException(e);
    }

    // Should never happen, if an error occurs, it should throw during the insert.
    if (!insertedEntity) {
    throw new EntryNotFoundAfterInsertException(this.entity);
    }

    return insertedEntity;
```

src/repositories/ImmutableRepository.ts
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Abstract class representing a repository for immutable entities.
+ * Extends the AbstractRepository class to provide additional functionality for adding and retrieving entities.
+ *
+ * @template T The type of the entity.
+ */
+export abstract class ImmutableRepository<T extends BaseEntity> extends AbstractRepository<T> {
+  /** The primary key of the entity used during the find. */
+  protected readonly primaryKey: keyof T;
```
Defferrard, line 29: This `primaryKey` variable's name is already used in functions, I feel like it can be confusing. What about `primaryKeyName` ? Otherwise, do you have other ideas ?

src/utils/errors.ts
```diff
@@ -5,6 +5,14 @@ import { FAILURE_MESSAGES } from '@graasp/translations';
 
 export const ConfigurationError = ErrorFactory('config');
 
+export abstract class CustomError extends Error {
```
Defferrard, line 8: I feel like we could easily inherit CustomError on bad use cases. It shouldn't be returned to the client since there is no Error Code, then the name should probably be more system/backend related ?

src/services/item/plugins/app/appAction/repository.ts
```diff
@@ -1,77 +1,81 @@
...
 import { mapById } from '../../../../utils';
 import { ManyItemsGetFilter, SingleItemGetFilter } from '../interfaces/request';
 import { AppAction } from './appAction';
 import { InputAppAction } from './interfaces/app-action';
 
-export const AppActionRepository = AppDataSource.getRepository(AppAction).extend({
-  async post(itemId: string, memberId: string, body: Partial<InputAppAction>): Promise<AppAction> {
-    const created = await this.insert({
+type PostAppActionBody = Partial<InputAppAction> & { itemId: string; memberId: string };
+type PatchAppActionBody = Partial<Omit<AppAction, 'item'>>;
```
Defferrard, line 16: As we discussed, `Post` and `Patch` are more HTTP related. Should we rename theses to
```suggestion
type CreateAppActionBody = Partial<InputAppAction> & { itemId: string; memberId: string };
type UpdateAppActionBody = Partial<Omit<AppAction, 'item'>>;
```

Defferrard, line 16: As we discussed, `Post` and `Patch` are more HTTP related. Should we renamed theses to :
```suggestion
type CreateAppActionBody = Partial<InputAppAction> & { itemId: string; memberId: string };
type UpdateAppActionBody = Partial<Omit<AppAction, 'item'>>;
```

Defferrard (approved):
LGTM, Good job ! 🔥
I've just some suggestion, feel free to accept them or not !

We should probably create an issue for theses cases so we can discuss about it with the team.

- Naming of the classes (B.e MutableRepository extends ImmutableRepository) 
- PrimaryKey type is only supported with strings
- PrimaryKey can't be a composition of keys
- Typecasting in functions

src/utils/errors.ts
ReidyT: I'll rename it and move directly in the Repositories error.

PR #1289, opened 2024-08-09
fix: new email builder and title escape fix

### LINKED ISSUE #1160: Special characters are not displayed correctly in mail's subject
`kim has shared Offres d&#39;opportunités et de services with you on Graasp`
Where it's supposed to be `kim has shared Offres d'opportunités et de services with you on Graasp`

MatveyK:
Fixes #1160

This is my first commit, so any suggestions, commentaries, critiques are welcome !


src/services/itemMembership/service.ts
```diff
@@ -46,7 +46,11 @@ export class ItemMembershipService {
         ${this.mailerService.buildButton(link, t(MAIL.SHARE_ITEM_BUTTON))}
       `;
 
-    const title = t(MAIL.SHARE_ITEM_TITLE, { creatorName: actor.name, itemName: item.name });
+    const title = t(MAIL.SHARE_ITEM_TITLE, {
+      creatorName: actor.name,
+      itemName: item.name,
+      interpolation: { escapeValue: false },
```
ReidyT, line 52: I'm just wondering if we have to escape some html or not ? I know that recent mail clients protect us from malicious html, but maybe it could be an added protection to at least replace common html like the `<>` ? What do you think ?

ReidyT (changes requested):
Thank you for your first commit 🥳. It looks good, I'll just leave you with one question and the possibility to write a test.

Concerning the git messages and other conventions we have, you can check out our [developer guide] (https://graasp.github.io/docs/developer/guides/style#submitting-code-for-review), but the commit messages look fine, except for the capitalization. In fact, we always start the description of our commit with a lower case letter (`feat: my feature` and not `feat: My feature`). Personally, I try to remove commits that I used during the implementation like fixing tests, or issues in the CI, to keep the merged commit as clean as possible.

Another recommendation is to add the PR to the Graasp project with the current sprint and the status `ready to review`.  This can be done from the menu on the right in the PR. This way, we can easily spot when your PR is ready to be reviewed.

The last thing that comes to mind is that we always write tests for new features or bug fixes to make sure the code works properly and doesn't break in the future. Do you think you can write a small test that checks the subject is not escaped ? I think you could mock the MailServer and check that the subject passed to sendEmail is not escaped.

Don't hesitate if you have questions or if you need more clarifications 😄.

Defferrard (commented):
Great Solution 🔥
Unfortunately it doesn't solve every issues since we're sending a lot of notification by mail.

I've shared a Folder with a mail that is not registered yet on the database, and the subject still have escaped characters.

Do you think there is a way to have a global fix ?

src/services/itemMembership/service.ts
MatveyK: Yes, I thought about it too. In this particular case it should be fine since the mail subject should always be treated as plain text by the email clients.

But it's definitely a good practice to sanitize user input. But if we want to sanitize the input in this case, we would have to roll out our own sanitization function which would remove the `<>"` but not the `'`. Seems like a bit of an overkill just for the email subject to me.

What do your thoughts ?

MatveyK:
Thank you for your comments !

@ReidyT Thanks for the clarification on the commit message styling. I can reset this branch with one single and correct message as soon as I'm done. Is this fine by you ? Or you prefer some other way ?

I've also set the PR status to "In review", thanks for the remark.

@Defferrard It seems like there's no trivial global fix since we need to escape the special characters when it's HTML, but we shouldn't escape them when it's the email subject. One option would be to define a translation function in the service that does not escape the special characters and use it only when translating plaintext. Which means that you must know which function to use depending on the input. Another (simpler) option could be to just present a generic message `Jeremy has shared a resource with you on Graasp`. This way, we can avoid the unsanitized user input altogether.

What are your thoughts ?

MatveyK:
As discussed on Tuesday, I have simplified the e-mail subject strings in english language. It seems like a reasonable solution to me.

I have not touched the subject strings for other languages. I think that it's better to wait for the discussion of this solution next week before proceeding with the translation modifications.

MatveyK:
This is the final version of the refactor that puts all the email composition in the mail service, where I think it should be.

ReidyT (approved):
Thanks for this PR and for your great job! I think it's a really nice refactor 🙌. I just let you some comments and suggestions.

src/plugins/mailer/service.ts
```diff
@@ -33,6 +35,74 @@ export class MailerService {
...
+  public async composeAndSendEmail(
+    email: string,
+    userLang: string,
+    subject: string,
+    buttonText: string,
+    text: string,
+    translationVariables: { [key: string]: string },
+    callToActionLink: string,
+    includeUserAgreement: boolean = false,
+    signUpNotRequested: boolean = false,
```
pyphilia, line 68: Actually I preferred when the email is built where it will be used, otherwise you have lots of options in your function that are only relevant for one use case. And the day you want a new email format you will need to create a new function just for this email.

pyphilia (changes requested):
Thanks for your first commit! 🍗

src/plugins/mailer/service.ts
pyphilia: Any opinion? @ReidyT @spaenleh @Defferrard ?

ReidyT: That's a bit the direction I would take with the Builder. You'd build the email where you use it, and then you just pass the mail object to the mailer. What do you think about that @pyphilia? Also, I think if we have to rebuild the same mail in multiple place, we could use a director ([still in the builder pattern](https://refactoring.guru/design-patterns/builder)) to define specific configuration of a mail (specific mail "template").

ReidyT: But maybe and advantage with the mail builder, is that the other service don't care how to build the mail, it just have to define what will be in the mail.

src/plugins/mailer/service.ts
```diff
@@ -33,6 +35,74 @@ export class MailerService {
...
+    i18next.changeLanguage(lang);
+    return i18next.t;
+  }
+
+  public async sendEmailChangeRequestConfirmation(
+    oldEmail: string,
+    newEmail: string,
+    userLang: string,
+  ) {
+    const lang = userLang ?? DEFAULT_LANG;
```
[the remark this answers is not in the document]
MatveyK: Indeed !

src/plugins/mailer/service.ts
```diff
@@ -33,6 +35,74 @@ export class MailerService {
...
+    email: string,
+    userLang: string,
+    subject: string,
+    buttonText: string,
+    text: string,
+    translationVariables: { [key: string]: string },
+    callToActionLink: string,
+    includeUserAgreement: boolean = false,
+    signUpNotRequested: boolean = false,
+  ) {
```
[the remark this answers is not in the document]
MatveyK: Agreed with the object for the function parameter.

Indeed, the builder could work here and looks nice from your example. Thanks for the suggestion.

src/plugins/mailer/test/index.test.ts
```diff
@@ -25,7 +33,11 @@ describe('Mailer', () => {
     it('generates button given link and text', () => {
       const link = 'mylink';
       const text = 'mytext';
-      const button = mailerService.buildButton(link, text);
+
+      // this is a hack to get the private function of the class
```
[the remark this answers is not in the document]
MatveyK: Yeah, I felt pretty dirty after doing this manipulation. But I think we can completely bypass this point by using the builder.

src/plugins/mailer/service.ts
```diff
@@ -33,6 +35,74 @@ export class MailerService {
...
 
+  public translate(lang: string = DEFAULT_LANG) {
+    i18next.changeLanguage(lang);
+    return i18next.t;
+  }
+
+  public async sendEmailChangeRequestConfirmation(
+    oldEmail: string,
+    newEmail: string,
+    userLang: string,
```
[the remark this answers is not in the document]
MatveyK: Agreed.

src/plugins/mailer/service.ts
MatveyK: @pyphilia I found that the problem with constructing the emails all over the place was that eventually everyone just copy-pasted the same code with slight modifications everywhere. This makes sense, since it's an action that you don't write often enough.

Although I admit, I may have gone overboard by trying to centralise e-mail composition. It makes the system a bit too rigid. I think that the solution proposed by @ReidyT could work. With a centralised builder, we can have a guided way to create an e-mail, without it becoming completely rigid.

What do you think ?

MatveyK:
I have implemented the email builder, as we have specified. I think that it's a fine middle ground between constructing emails all over the place and having one strict mega-function that constructs emails for you.

Let me know what you think.

src/plugins/mailer/builder.ts
```diff
@@ -0,0 +1,169 @@
...
+    this.htmlText += this.buildText(
+      escapedText,
+      // Add margin top of -15px to remove 15px margin bottom of the button.
+      { 'text-align': 'center', 'font-size': '10px', 'margin-top': '-15px' },
+    );
+
+    return this;
+  }
+
+  public signUpNotRequested(): MailBuilder {
```
ReidyT, line 94:
```suggestion
  public addSignUpNotRequested(): MailBuilder {
```

src/plugins/mailer/builder.ts
```diff
@@ -0,0 +1,169 @@
...
+      this.t(buttonText, this.translationVariables),
+    );
+
+    // keep this text for user agreement
+    this.buttonText = buttonText;
+
+    return this;
+  }
+
+  public includeUserAgreement(): MailBuilder {
```
ReidyT, line 69: I'm wondering if we should not call it `addUserAgreement` to be more coherent with the other function names? (I know that was my proposition ^^)

Maybe we would like to keep `includeUserAgreement`, but in this case, what about including once, even if we call this function twice? I don't know, I'm just thinking about that, it's not mandatory, but it could reflect that you include once.

ReidyT (changes requested):
It's really nice changes! It looks really good to me 🤩. I just let you some questions and suggestions, I'm open to discussion 💬. Good job 🎇🎆

src/plugins/mailer/builder.ts
```diff
@@ -0,0 +1,169 @@
...
+  private htmlText = '';
+  private buttonText: string;
+  private translationVariables: { [key: string]: string };
+  private language: string = DEFAULT_LANG;
+
+  private t: TFunction;
+
+  constructor({
+    subject,
+    translationVariables,
```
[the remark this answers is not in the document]
pyphilia: I agree! 

Maybe `subject` could be `subject: { key: string, translationVariables: Map<string, string> }` 
I forgot if Map is better than your syntax.

src/plugins/mailer/builder.ts
```diff
@@ -0,0 +1,169 @@
...
+    this.plainText += `\n${translatedText}\n`;
+    this.htmlText += this.buildText(this.t(MAIL.SIGN_UP_NOT_REQUESTED));
+
+    return this;
+  }
+
+  public build(): Mail {
+    const emailSubject = this.t(this.subject, {
+      ...this.translationVariables,
+      interpolation: { escapeValue: false },
```
pyphilia, line 106: Why not do this in the constructor where we define the subject?

src/plugins/mailer/builder.ts
```diff
@@ -0,0 +1,169 @@
...
+import i18next from './i18n';
+import { MAIL } from './langs/constants';
+import { Mail } from './service';
+
+type CssStyles = { [key: string]: string };
+
+export class MailBuilder {
+  private subject: string;
+  private plainText = '';
+  private htmlText = '';
```
Defferrard, line 15: Theses 2 properties are really confusing to me. I've understood that they're are strongly related, so is it possible to have something that represent that, or having a comment to give some explanation ?

It would be awesome (maybe a bit difficult), to have a private function `appendText(...)` who is the only responsible of the content of theses properties. It could eventually reduce the amount of this file's lines

src/plugins/mailer/builder.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  public signUpNotRequested(): MailBuilder {
+    const translatedText = this.t(MAIL.SIGN_UP_NOT_REQUESTED);
+
+    this.plainText += `\n${translatedText}\n`;
+    this.htmlText += this.buildText(this.t(MAIL.SIGN_UP_NOT_REQUESTED));
+
+    return this;
+  }
```
Defferrard, line 101: Is it possible to have something that way ?

```suggestion
  public signUpNotRequested(): MailBuilder {
    return this.addText(MAIL.SIGN_UP_NOT_REQUESTED);
  }
```

src/services/auth/plugins/mobile/mobile.test.ts
```diff
@@ -603,13 +580,15 @@ describe('Mobile Endpoints', () => {
 
       expect(responseRegister.statusCode).toBe(StatusCodes.NO_CONTENT);
       expect(mockSendEmail).toHaveBeenCalledTimes(1);
-      const authUrl = new URL(mockSendEmail.mock.calls[0][2]);
+
+      const tokenRegex = /\?t=([\w\-\.]{1,1000})/;
```
Defferrard, line 584: Apparently this regex is also used in `password/index.test.ts` could we extract it into a test utils file ?

Defferrard (commented):
Great job 🔥 ! I just have some comments, otherwise I agree with what the others said

src/plugins/mailer/builder.ts
```diff
@@ -0,0 +1,169 @@
...
+  public addButton(buttonText: string, callToActionLink: string): MailBuilder {
+    this.plainText += `\n${callToActionLink}\n`;
+
+    this.htmlText += this.buildButton(
+      callToActionLink,
+      this.t(buttonText, this.translationVariables),
+    );
+
+    // keep this text for user agreement
+    this.buttonText = buttonText;
```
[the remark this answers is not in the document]
MatveyK: Agree, passing the parameters explicitly is the better option.

src/plugins/mailer/builder.ts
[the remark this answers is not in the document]
MatveyK: Yes, I will add comments to the builder properties.

src/plugins/mailer/service.ts
```diff
@@ -33,6 +35,10 @@ export class MailerService {
     });
   }
 
+  public async send(mail: Mail, to: string, from: string = this.fromEmail) {
```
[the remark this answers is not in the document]
MatveyK: I thought that it was better to leave two functions: one for the e-mail object produced by the builder and one function to send "raw" e-mails. The second function would serve as an escape hatch for when you would need to bypass the e-mail builder.

src/plugins/mailer/builder.ts
MatveyK: Indeed !

src/plugins/mailer/builder.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    if (lang) {
+      this.language = lang;
+    }
+
+    i18next.changeLanguage(this.language);
+    this.t = i18next.t;
+  }
+
+  public addText(stringId: string): MailBuilder {
```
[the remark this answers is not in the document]
MatveyK: Indeed, it is possible ! I have changed the type signature to `(typeof MAIL)[keyof typeof MAIL]` and it works flawlessly.

src/plugins/mailer/builder.ts
MatveyK: I have added comments that explain the need for these two separate properties.

src/plugins/mailer/builder.ts
MatveyK: Indeed.

MatveyK:
I have improved the mail builder based on your feedback.

ReidyT (approved):
Thank you very much for those changes! It look very nice! Great job 🚀

src/plugins/mailer/builder.ts
```diff
@@ -0,0 +1,169 @@
...
+      graaspLandingPageOrigin: GRAASP_LANDING_PAGE_ORIGIN,
+    });
+    this.htmlText += this.buildText(
+      escapedText,
+      // Add margin top of -15px to remove 15px margin bottom of the button.
+      { 'text-align': 'center', 'font-size': '10px', 'margin-top': '-15px' },
+    );
+
+    return this;
+  }
```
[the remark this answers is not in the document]
MatveyK: Yes, I have thought about that, but having multiple interfaces seems a bit cumbersome to me. I think that it's acceptable to have the developer be responsible for not putting the user agreement before other elements. I have also put a comment in the function description.

test/utils.ts
```diff
@@ -17,3 +17,5 @@ export function mockCaptchaValidationOnce(action: RecaptchaActionType) {
     score: 1,
   });
 }
+
+export const tokenRegex = /\?t=([\w\-\.]{1,1000})/;
```
spaenleh, line 21: This should be in ALL_CAPS.

Also where does the `{1,1000}`come from ? Is it arbitrary ?

src/services/item/plugins/invitation/service.ts
```diff
@@ -54,37 +54,26 @@ export class InvitationService {
...
+    })
+      .addText(MAIL.INVITATION_TEXT, {
+        itemName: item.name,
+        creatorName: member.name,
+      })
+      .addButton(MAIL.SIGN_UP_BUTTON_TEXT, link)
+      .addUserAgreement(MAIL.SIGN_UP_BUTTON_TEXT)
+      .build();
+
+    this.mailerService.send(mail, member.email).catch((err) => {
```
spaenleh, line 76: I think this is an error. We are supposed to send the email to the email that was invited not to the user that created the invitation. Does a test ensure this ? It should have broken. 
```suggestion
    this.mailerService.send(mail, invitation.email).catch((err) => {
```

src/services/auth/service.ts
```diff
@@ -45,37 +45,28 @@ export class AuthService {
...
     // don't wait for mailerService's response; log error and link if it fails.
     this.mailerService
-      .sendEmail(subject, member.email, link, html, footer)
+      .send(mail, member.email)
       .catch((err) => this.log.warn(err, `mailerService failed. link: ${link}`));
   };
 
   generateLoginLinkAndEmailIt = async (
     member: Member,
     options: { challenge?: string; lang?: string; url?: string } = {},
```
spaenleh, line 67: If `lang` is not used in the function, remove it from the options and the call site.

src/services/auth/service.ts
```diff
@@ -45,37 +45,28 @@ export class AuthService {
...
-
-    const footer = this.mailerService.buildFooter(lang);
+    const mail = new MailBuilder({
+      subject: { text: MAIL.SIGN_UP_TITLE },
+      lang: member.lang,
+    })
+      .addText(MAIL.GREETINGS)
+      .addText(MAIL.SIGN_UP_TEXT)
+      .addButton(MAIL.SIGN_UP_BUTTON_TEXT, link)
+      .addUserAgreement(MAIL.SIGN_UP_BUTTON_TEXT)
```
spaenleh, line 55: It looks strange to me that we are providing text in this function, and the text is the same as the signup button ? either rename the translation string to better fit the purpose, or remove the argument and have the text be defined inside the builder function.

src/services/auth/plugins/mobile/mobile.test.ts
```diff
@@ -34,7 +35,9 @@ const memberRawRepository = AppDataSource.getRepository(Member);
 const mockCaptchaValidation = (action: RecaptchaActionType) => {
   (fetch as jest.MockedFunction<typeof fetch>).mockImplementation(async () => {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
spaenleh, line 37: Can we remove this ignore ?

src/plugins/mailer/service.ts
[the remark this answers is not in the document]
spaenleh: I would rename the `sendEmail` to `sendRawEmail` and add a JSDocString comment that explains the purpose of it, so it is not confused. Actually I think having a JSDoc on both functions would be good.

src/plugins/mailer/test/index.test.ts
```diff
@@ -1,48 +1,149 @@
...
     ({ app } = await build());
     mailerService = resolveDependency(MailerService);
   });
 
-  afterAll(async () => {
+  beforeEach(async () => {
+    mockSendEmail = jest
+      .spyOn(mailerService, 'sendEmail')
+      .mockImplementation(async () => Promise.resolve());
+  });
```
spaenleh, line 26: Does this need to be re-created before each test ? Can't it be defined once in the `beforeAll` block and simply reset in the afterEach (already done)?

src/plugins/mailer/test/index.test.ts
```diff
@@ -1,48 +1,149 @@
...
+        .addButton(MAIL.MEMBERSHIP_REQUEST_BUTTON_TEXT, link, { itemName })
+        .build();
+
+      await mailerService.send(mail, email);
+
+      expect(mockSendEmail).toHaveBeenCalledWith(
+        expect.stringContaining(memberName),
+        email,
+        expect.stringContaining(link),
+        expect.stringContaining(itemName) && expect.stringContaining(link),
```
spaenleh, line 87: Not sure what this does ...

src/plugins/mailer/builder.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  public addSignUpNotRequested(): MailBuilder {
```
spaenleh, line 119: It confused me when I read the Builder invocations with this function. It is not exclusive to the signUp. Maybe rename it to something less signUp-related.
```suggestion
  public addEmailSubscriptionNotice(): MailBuilder {
```

Since the message is something like: `You are receiving this email because of your account on the Graasp platform.`

test/utils.ts
MatveyK: OK for all caps. The limit was indeed arbitrary, I've changed it to `+`.

src/services/item/plugins/invitation/service.ts
MatveyK: Nice catch. I have also added a test that checks the invitation email address.

src/services/auth/plugins/mobile/mobile.test.ts
MatveyK: Sure, I'm not even sure what this is for.

src/plugins/mailer/test/index.test.ts
MatveyK: Indeed.

src/plugins/mailer/builder.ts
MatveyK: I have changed the string name to better reflect what it actually does.

src/services/auth/service.ts
MatveyK: I have changed the actual string for the user agreement so that it no longer requires the button text. This way I could also remove the parameter for the `addUserAgreement` function.

src/plugins/mailer/test/index.test.ts
```diff
@@ -1,48 +1,146 @@
...
   beforeAll(async () => {
     ({ app } = await build());
     mailerService = resolveDependency(MailerService);
+    mockSendEmail = jest
+      .spyOn(mailerService, 'sendRaw')
+      .mockImplementation(async () => Promise.resolve());
   });
 
-  afterAll(async () => {
+  afterEach(async () => {
```
spaenleh, line 25: Split this into an `afterEach` that resets jet mocks and an `afterAll` that resets the DB and closes the server.

spaenleh (approved):

src/plugins/mailer/test/index.test.ts
MatveyK: Done.

pyphilia (approved):

PR #1290, opened 2024-08-12
feat(membership request): request creation api route

Defferrard:
subfix https://github.com/graasp/graasp/issues/1220


src/services/itemMembership/plugins/MembershipRequest/service.ts
```diff
@@ -1,11 +1,60 @@
...
   private readonly mailerService: MailerService;
-  constructor(mailerService: MailerService) {
+  private readonly log: BaseLogger;
+
+  constructor(mailerService: MailerService, log: BaseLogger) {
     this.mailerService = mailerService;
+    this.log = log;
+  }
+
+  async get({ membershipRequestRepository }: Repositories, memberId: string, itemId: string) {
```
ReidyT, line 22: It is possible to pass only the needed repository here ?

```suggestion
  async get(membershipRequestRepository: MembershipRequestRepository, memberId: string, itemId: string) {
```

src/services/itemMembership/plugins/MembershipRequest/service.ts
```diff
@@ -1,11 +1,60 @@
...
+    return await membershipRequestRepository.get(memberId, itemId);
+  }
+
+  async post({ membershipRequestRepository }: Repositories, memberId: string, itemId: string) {
```
ReidyT, line 26: Same here ?

src/services/itemMembership/plugins/MembershipRequest/service.ts
```diff
@@ -1,11 +1,60 @@
...
+    return await membershipRequestRepository.post(memberId, itemId);
+  }
+
+  async notifyAdmins(member: Member, { itemMembershipRepository }: Repositories, item: Item) {
```
ReidyT, line 30: And here ?

src/services/itemMembership/plugins/MembershipRequest/service.ts
```diff
@@ -1,11 +1,60 @@
...
+    const adminMemberships = await itemMembershipRepository.getAllWithPermission(
+      item.path,
+      PermissionLevel.Admin,
+    );
+
+    const link = new URL(`/items/${item.id}/share`, BUILDER_HOST.url).toString();
```
ReidyT, line 36: What do you think about building the url path using a function ? in this way, if the share url change, we just have to update the method ?

src/services/itemMembership/repository.ts
```diff
@@ -442,6 +443,16 @@ export const ItemMembershipRepository = AppDataSource.getRepository(ItemMembersh
     });
   },
 
+  async getAllWithPermission(
```
ReidyT, line 446: What do you think about the following name ? 

```suggestion
  async getByItemPathAndPermission(
```

ReidyT (approved):
Thanks for this PR, great job! It looks good to me, I just let you some suggestions and comments 🧑‍🚀🚀

PR #1293, opened 2024-08-13
feat(membership request): get all requests from an item

Defferrard:
subfix from https://github.com/graasp/graasp/issues/1220

### DIFF
--- a/src/services/itemMembership/plugins/MembershipRequest/repository.ts
+++ b/src/services/itemMembership/plugins/MembershipRequest/repository.ts
@@ -26,10 +26,26 @@ export class MembershipRequestRepository extends AbstractRepository<MembershipRe
   }
 
   async post(memberId: string, itemId: string) {
+    if (!memberId) {
+      throw new MemberNotFound();
+    } else if (!itemId) {
+      throw new ItemNotFound(itemId);
+    }
+
     await this.repository.insert({
       member: { id: memberId },
       item: { id: itemId },
     });
     return await this.get(memberId, itemId);
+  }
+
+  async getAllByItem(itemId: string) {
+    if (!itemId) {
+      throw new ItemNotFound(itemId);
+    }
+    return await this.repository.find({
+      where: { item: { id: itemId } },
+      relations: ['member'],
+    });
   }
 }
--- a/src/services/itemMembership/plugins/MembershipRequest/service.ts
+++ b/src/services/itemMembership/plugins/MembershipRequest/service.ts
@@ -19,6 +19,10 @@ export class MembershipRequestService {
     this.log = log;
   }
 
+  async getAllByItem({ membershipRequestRepository }: Repositories, itemId: string) {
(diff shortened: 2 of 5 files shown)


src/services/itemMembership/plugins/MembershipRequest/repository.ts
```diff
@@ -7,4 +7,30 @@ export class MembershipRequestRepository extends AbstractRepository<MembershipRe
   constructor(manager?: EntityManager) {
     super(MembershipRequest, manager);
   }
+
+  async get(memberId: string, itemId: string) {
+    return await this.repository.findOne({
```
ReidyT, line 12: I think you have to check that memberId and itemId are defined to avoid TypeORM to returns all the entries.

src/services/itemMembership/plugins/MembershipRequest/repository.ts
```diff
@@ -7,4 +7,30 @@ export class MembershipRequestRepository extends AbstractRepository<MembershipRe
...
+    await this.repository.insert({
+      member: { id: memberId },
+      item: { id: itemId },
+    });
+    const result = await this.get(memberId, itemId);
+    return result;
+  }
+
+  async getAllByItem(itemId: string) {
+    return await this.repository.find({
```
ReidyT, line 31: same here for the itemId

src/services/itemMembership/plugins/MembershipRequest/repository.ts
```diff
@@ -7,4 +7,30 @@ export class MembershipRequestRepository extends AbstractRepository<MembershipRe
...
+  async post(memberId: string, itemId: string) {
+    await this.repository.insert({
+      member: { id: memberId },
+      item: { id: itemId },
+    });
+    const result = await this.get(memberId, itemId);
+    return result;
+  }
+
+  async getAllByItem(itemId: string) {
```
ReidyT, line 30: Maybe `getByItem` is enough here ? But if you prefer `getAllByItem` it's ok for me ^^

src/services/itemMembership/plugins/MembershipRequest/service.ts
```diff
@@ -1,11 +1,16 @@
...
+import { Repositories } from '../../../../utils/repositories';
 
 @singleton()
 export class MembershipRequestService {
   private readonly mailerService: MailerService;
   constructor(mailerService: MailerService) {
     this.mailerService = mailerService;
   }
+
+  async getAllByItem({ membershipRequestRepository }: Repositories, itemId: string) {
```
ReidyT, line 13: Same comment about the name here (as you want ^^). Also I Think you can just pass the wanted repo:

```suggestion
  async getAllByItem(membershipRequestRepository: MembershipRequestRepository, itemId: string) {
```

ReidyT (approved):
Thanks for your work, LGTM! 🌞

PR #1296, opened 2024-08-13
feat(refactor): item repository extends mutable repository

ReidyT:


Defferrard (approved):
LGTM ! Goob job ! I've just an interrogation !

src/services/auth/plugins/passport/strategies/jwtApps.ts
```diff
@@ -58,7 +58,7 @@ export default (
           });
         } catch (err) {
           // Exception occurred while fetching item
-          // itemRepository.get() can fail for many reasons like the item was not found, database error, etc.
+          // itemRepository.getOne() can fail for many reasons like the item was not found, database error, etc.
```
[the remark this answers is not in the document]
ReidyT: In fact, I haven't made this change now for the Item, but I'll do it right now!

src/repositories/errors.ts
```diff
@@ -64,3 +64,13 @@ export class DeleteException extends RepositoryException {
     super(message, 'DeleteException');
   }
 }
+
+// TODO: use entity class when rebase on main!
```
ReidyT, line 68: Update before merge!

src/services/auth/plugins/passport/strategies/jwtApps.ts
```diff
@@ -58,7 +58,7 @@ export default (
           });
         } catch (err) {
           // Exception occurred while fetching item
-          // itemRepository.get() can fail for many reasons like the item was not found, database error, etc.
+          // itemRepository.getOne() can fail for many reasons like database error, etc.
```
ReidyT, line 61:
```suggestion
          // itemRepository.getOneOrThrow() can fail for many reasons like the item was not found, database error, etc.
```

PR #1299, opened 2024-08-14
feat(membership request): delete request route

Defferrard:
subfix #1220


ReidyT (approved):
Thanks for you job! LGTM 🎃 I just let some comments

PR #1302, opened 2024-08-14
refactor(AppDataRepository): extends MutableRepository

ReidyT:


src/repositories/errors.ts
```diff
@@ -32,8 +32,8 @@ export class InsertionException extends RepositoryException {
 export class EntryNotFoundAfterInsertException<T extends BaseEntity> extends RepositoryException {
   constructor(entity: EntityTarget<T>) {
     super(
-      `The insertion of a new ${entity} failed, the created id was not found.`,
-      'InsertionException',
+      `The insertion of a new ${entity} failed, the created entry was not found.`,
+      'EntryNotFoundAfterInsertException',
```
Defferrard, line 36: I've just remembered that we should probably be careful with this kind of exception.
Recently we had an issue where we were throwing too much information to the user, so Basile decided to throw only the Id of the entity.

I don't know what we're expecting the error message to be here, I feel like it's gonna be :
 `The insertion of a new [object Object] failed, the created entry was not found.` which doesn't make sense either if I'm right.

src/services/item/plugins/app/appData/repository.ts
```diff
@@ -1,65 +1,60 @@
...
+      },
+      RELATIONS,
+    );
+  }
 
-  async deleteOne(itemId: string, appDataId: string): Promise<UUID> {
-    const deleteResult = await this.delete(appDataId);
+  async updateOne(appDataId: string, body: UpdateAppDataBody): Promise<AppData> {
+    // shouldn't update file data
+    // TODO: optimize and refactor
```
Defferrard, line 38: We should probably create an Issue instead of having a todo here ?

src/services/item/plugins/app/appData/repository.ts
```diff
@@ -68,10 +63,10 @@ export const AppDataRepository = AppDataSource.getRepository(AppData).extend({
    */
   async getForMemberExport(memberId: string): Promise<AppData[]> {
     if (!memberId) {
-      throw new MemberIdentifierNotFound();
+      throw new IllegalArgumentException('The memberId must be defined!');
     }
```
Defferrard, line 67: Should we use `this.throwsIfPKIsInvalid(memberId);`, or an equivalent ?

Defferrard (approved):
LGTM Good job ! I just have some suggestion !

src/repositories/errors.ts
ReidyT: I totally agree with you! Here the entity is not the instance of an BaseEntity, but just the class of the Entity. It is just used to display the Entity's name in the exception like:

"The deletion of class Item". But I just saw that it generate something like "The deletion of class Item extends typeorm_1.BaseEntity", so I have to fix that ^^.

src/services/item/plugins/app/appData/repository.ts
ReidyT: Yeah I agree, it was already here ^^

Defferrard: Ah my bad then ! I though you added it 😅

src/repositories/errors.ts
ReidyT: But I just saw that I effectively mad a mistake, so thanks!

src/services/item/plugins/app/appData/repository.ts
ReidyT: Yes I agree, I just moved the code, it's why there is still the TODO here ^^
I think I'll remove the TODO, because if we want to prevent updating app data file, we can't do it without doing a select before...

PR #1307, opened 2024-08-15
feat: pdf thumbnail

Defferrard:


src/services/item/plugins/file/service.ts
```diff
@@ -141,6 +143,22 @@ class FileItemService {
         } catch (e) {
           console.error(e);
         }
+      } else if (MimeTypes.isPdf(mimetype)) {
+        try {
+          const outputImg = await fromPath(path, {
+            width: 200,
+            height: 200,
+          })(1, { responseType: 'buffer' });
```
ReidyT, line 151: Maybe just precise that 1 is for the first page?

ReidyT (approved):
Really nice PR! Great job 🔥

PR #1313, opened 2024-08-19
refactor(repositories): shortlink into new repositories

Defferrard:


src/repositories/ImmutableRepository.ts
```diff
@@ -152,7 +152,9 @@ export abstract class ImmutableRepository<T extends BaseEntity> extends Abstract
   protected async insert(entity: DeepPartial<T>, relations?: FindOptionsRelations<T>): Promise<T> {
     try {
       const insertResult = await this.repository.insert(entity as QueryDeepPartialEntity<T>);
-      const insertedEntity = await this.findOne(insertResult.identifiers[0].id, { relations });
+      const insertedEntity = await this.findOne(insertResult.identifiers[0][this.primaryKeyName], {
```
ReidyT, line 155: Good catch!

spaenleh (approved):
LGTM !

src/services/item/plugins/shortLink/repository.ts
```diff
@@ -52,57 +72,42 @@ export const ShortLinkRepository = AppDataSource.getRepository(ShortLink).extend
...
-
-      await this.update(alias, postLink);
-      const aliasId = postLink['alias'] ?? alias;
-      return await this.get(aliasId); // return the updated short link
+      return await super.updateOne(alias, entity);
     } catch (e) {
-      // if a duplication entry error throw, the postlink must contain the alias
-      if (isDuplicateEntryError(e)) {
-        throw new ShortLinkDuplication(postLink['alias']);
+      if (isDuplicateEntryError(e.message)) {
```
ReidyT, line 107: Isn't `isDuplicatedEntryError` waiting to receive an error instead of a string?

ReidyT (approved):
Thanks for this PR! LGTM 🚀

src/services/item/plugins/shortLink/repository.ts
Defferrard: Yes it does, but apparently, because of InsertionException is encapsulating the error in message property I have to fetch it that way (So I don't break Frontend).

It works because `error` type is `any`

ReidyT: Ok I better understand what you proposed with the nested exception.. What do you think about 1) handling the DuplicatedError in the insert method of the Repository and throw a EntityDuplicatedError in it and 2) using the exception's message in the generic InsertionException instead of the error (I made a mistake here, I should have to use the message error)?

PR #1314, opened 2024-08-19
refactor(ItemCategoryRepository): extends MutableRepository

ReidyT:


src/services/item/plugins/itemCategory/repositories/itemCategory.ts
```diff
@@ -1,80 +1,80 @@
...
 import { ItemCategory } from '../entities/ItemCategory';
 import { DuplicateItemCategoryError } from '../errors';
 
+type CreateItemCategoryBody = { itemPath: string; categoryId: string };
+const RELATIONS = { category: true, item: true };
+
 /**
  * Database's first layer of abstraction for Categorys
  */
+export class ItemCategoryRepository extends MutableRepository<ItemCategory, never> {
```
Defferrard, line 18: I think it's a really good Idea to use the type `never` here. Do you think we could use the same kind of thing to select which CRUD operations are available, so we can merge Mutable and Immutable repository ?

Defferrard (approved):
LGTM 🔥 !

src/services/item/plugins/itemCategory/repositories/itemCategory.ts
ReidyT: I used the never here because it's the only entity (for now until the beginning of the refactor) which allow CRD (not the update), so I thought it could be okay to set the update unusable.

But I don't think we should do that too often, because at the end we will have the function available in the autocomplete, but we can't use it.

Also, if we have a lot of CRD repositories, maybe we'll have to think to another solution...

What do you think?

PR #1319, opened 2024-08-19
refactor(ChatMessageRepository): extends MutableRepository

ReidyT:
- refactor(ChatMessageRepository): extends MutableRepository
- fix(MutableRepository): use getOne in updateOne
- refactor(errors): handle the errors in the controller


src/repositories/interfaces.ts
```diff
@@ -1,9 +1,38 @@
...
+
 export interface GetAll<T extends BaseEntity> {
   /**
    * Retrieves all the entities.
    * @returns An array containing all the entities.
    */
   getAll(): Promise<T[]>;
 }
+
+export interface GetForItem<T extends BaseEntity> {
```
Defferrard, line 13: I'm nitpicking here so don't hesitate to ignore if you think it is not worthy.

I feel like `getForItem(...)` is a confusing name to me. I've the habit of having `getByItem(...)` or `getByItemId(...)`.

When I see `getForItem` I'm never totally sure of the use of this function. "Does it get something in a format that satisfy an Item, or does it get something related to an Item ?"

Just to make a comparaison (I didn't made a profesionnal analysis) from the search bar of Public GitHub :
- There is 88.6k files that contains `getByEmail`
- There is 298 files that contains `getForEmail`

- There is 2.4M files that contains `getById`
- There is 7.6k files that contains `getForId`

src/services/chat/repository.ts
```diff
@@ -1,117 +1,125 @@
...
+type IChatMessageRepository<T extends ChatMessage> = GetForItem<T> &
+  GetForItems<T> &
+  GetForMemberExport<T>;
+
+type ChatMessageUpdateBody = Partial<ChatMessage>;
+type ChatMessageCreateBody = { itemId: string; creator: Member; body: string };
+
+export class ChatMessageRepository
+  extends MutableRepository<ChatMessage, ChatMessageUpdateBody>
+  implements IChatMessageRepository<ChatMessage>
```
Defferrard, line 25: I think it is a really interesting way to keep the naming of our functions consistant.

But, we need to keep in mind that in the future, we're gonna use tsyringe and will get our Repository from the class name, and will rarely use this interface.. To be honest, I feel like we're not using interface how they're intented to be used here.

Have you considered to use ES-Lint or any equivalent ? I just saw that : https://typescript-eslint.io/rules/naming-convention/, could this apply to our issue ?

By the way, I think we should keep it as it is, and find a solution on another step of this refactoring.

src/services/chat/repository.ts
```diff
@@ -1,117 +1,125 @@
...
-  },
+  async updateOne(id: string, data: ChatMessageUpdateBody): Promise<ChatMessage> {
+    return await super.updateOne(id, data);
+  }
 
   /**
    * Remove all messages for the given chat
    * @param chatId Id of chat
    */
   async clearChat(itemId: string): Promise<ChatMessage[]> {
```
Defferrard, line 109:
```suggestion
  async deleteManyForItem(itemId: string): Promise<ChatMessage[]> {
```

Defferrard (approved):
LGTM ! I'm only opening some discussion, feel free to consider them or not. We can also think about them during an another iteration !

src/repositories/interfaces.ts
ReidyT: I totally agree with you, I personally prefer getBy instead of getFor, we can change that for me...

src/repositories/errorFactory.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const isQueryFailedError = (err: unknown): err is QueryFailedError & DatabaseError =>
+  err instanceof QueryFailedError;
+
+export const errorFactory = <T extends BaseEntity>({
+  error,
+  classEntity,
+  fallBackError,
+}: {
+  error: unknown & { message?: string };
```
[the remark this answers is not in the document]
ReidyT: Yeah for no reason, I started using unknwon and i forgot to remove it ^^

src/services/chat/repository.ts
ReidyT: On my point of view, using interfaces allow us to describe the contract between the repository and the client code (code that are using the repository) as well as allowing consistency in the signature of the functions. Each time a class implements the interface (GetAll) for example, the class guarantees the existence of the method getAll() which will return an array of T.

I didn't check in depth, but I'm not sure using ES-Lint allows us to enforce naming functions and return types, and even if it's the case, I'm wondering if it's the best solution. Maybe I'm wrong, but in my sense, using interfaces give us the possibility to enforce a consistent method signature per action.

I understand one potential pitfall which is the greater number of interfaces, but I don't think it's a big issue.

What do you think @Defferrard and @spaenleh @MatveyK @pyphilia ? I'm always open to changes ;)

spaenleh (approved):

ReidyT:
As we discussed this morning, the interfaces are too complexe for our purpose (writing documentation guide line is enough) as well as the errors handling, so I'll close this PR.

PR #1322, opened 2024-08-20
refactor(repositories): recycled item data extends MutableRepository

Defferrard:


src/services/item/plugins/recycled/repository.ts
```diff
@@ -1,17 +1,41 @@
...
+  async addOne({ itemPath, creatorId }: CreateRecycledItemDataBody) {
+    return await super.insert({ item: { path: itemPath }, creator: { id: creatorId } });
+  }
+
+  // warning: this call insert in the table
+  // but does not soft delete the item
+  // should we move to core item?
+  async addMany(items: Item[], creator: Member) {
+    const recycled = items.map((item) => this.repository.create({ item, creator }));
+    this.repository.insert(recycled as QueryDeepPartialEntity<RecycledItemData>[]);
```
ReidyT, line 31: you have to await here

src/services/item/plugins/recycled/repository.ts
```diff
@@ -1,17 +1,41 @@
...
+    return recycled;
+  }
+
+  async getManyByMember(member: Member): Promise<RecycledItemData[]> {
```
ReidyT, line 35: is `getByMember` enough if we choose to specify `getOne...` when returning a single entity?

src/services/item/plugins/recycled/repository.ts
```diff
@@ -25,52 +49,17 @@ export const RecycledItemDataRepository = AppDataSource.getRepository(RecycledIt
...
-    const recycled = items.map((item) => this.create({ item, creator }));
-    this.insert(recycled);
-    return recycled;
-  },
-
-  // warning: this call insert in the table
-  // but does not soft delete the item
-  // should we move to core item?
-  async restoreOne(item: Item) {
+  async deleteManyByItemPath(itemsPath: Item['path'][]) {
```
ReidyT, line 57: Same comment here?
```suggestion
  async deleteByItemPath(itemsPath: Item['path'][]) {
```

ReidyT (approved):
LGTM 🎃 I know that it's not you, but I have a global question, what is the purpose of this repository? I have the feeling that we have to manage two source of truth we that repo + the item repository. Indeed, when we soft remove an item, we also have to add in this repo, and if we restore or delete permanently, we have to use these two repositories right? @spaenleh do you know why this repository is needed?

src/services/item/plugins/recycled/repository.ts
pyphilia: I think I like it with the many precision.

pyphilia (approved):
LGTM!

src/repositories/ImmutableRepository.ts
```diff
@@ -104,7 +104,7 @@ export abstract class ImmutableRepository<T extends BaseEntity> extends Abstract
       throw new IllegalArgumentException(`The given ${name} is undefined!`);
     }
 
-    if (Array.isArray(value) && value.length === 0) {
+    if (Array.isArray(value) && value.length === 0 && value.every((v) => !v)) {
```
spaenleh, line 107: I am not sure to understand how the `value` array could have values inside when the length is zero ...

src/services/item/plugins/recycled/repository.ts
spaenleh, line 31: Is the casting fixed by an update of the sdk ? I think I removed the `extra.app.setting: unknown` property last week ?

src/services/item/plugins/recycled/repository.ts
```diff
@@ -25,52 +49,17 @@ export const RecycledItemDataRepository = AppDataSource.getRepository(RecycledIt
         { permission: PermissionLevel.Admin, accountId: member.id },
       )
       .getMany();
-
-    return recycledItemEntries;
-  },
+  }
 
   // warning: this call insert in the table
```
spaenleh, line 54:
```suggestion
  // warning: this call removes from the table
```

spaenleh (approved):

Defferrard:
> LGTM 🎃 I know that it's not you, but I have a global question, what is the purpose of this repository? I have the feeling that we have to manage two source of truth we that repo + the item repository. Indeed, when we soft remove an item, we also have to add in this repo, and if we restore or delete permanently, we have to use these two repositories right? @spaenleh do you know why this repository is needed?

I've created an issue for that
https://github.com/graasp/graasp/issues/1345

Issue #1288, opened 2024-08-08
Validation, serialization and documentation


Defferrard:
# Actual Configuration
- AJV
  - JSON Schema (>Draft 6) 
  - Fluent Schema
- OpenAPI 3.0

# Issues identified by our current implementation
- We have to describe our data multiple times (DTO, Schema)
- References are broken.
  - They're pointing to an URI, this is usually used to point to a distant schema on another server, but we're using it to reference a local schema. Because of that, the Swagger page will try to fetch the schema on the URI and will have broken references.
  - We can't easily get the content of a reference.
- Output validation takes too much time 
- Not compliant with OpenAPI 3.0
  - We can't namespace our components as we do with our definitions. Because of that, the Swagger page have broken components.
- `fastify.addSchema(...)` needs a parameter of type `unknown`

# Proposed Solution
## Refactor our current implementation
By renaming our references (https://json-schema.org/understanding-json-schema/structuring#schema-identification) and removing the namespaces.

## DTO to Schema
- Class Validator & Class Validator JSON Schema
  - https://github.com/typestack/class-validator
  - https://github.com/epiphone/class-validator-jsonschema

## Schema to DTO
- Zod
  - https://zod.dev/
- Typed AJV
  - https://github.com/keplr-team/typed-ajv

PR #1326, opened 2024-08-21
refactor(repositories): item flag extends immutable

Defferrard:


ReidyT (approved):
LGTM 🧑‍🚀

src/services/item/plugins/itemFlag/index.ts
```diff
@@ -20,25 +21,21 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
 
   // get flags
-  fastify.get(
-    '/flags',
-    { schema: getFlags, preHandler: optionalIsAuthenticated },
-    async ({ user }) => {
-      return itemFlagService.getAllFlags(user?.account, buildRepositories());
-    },
-  );
+  fastify.get('/flags', { schema: getFlags }, async () => {
```
spaenleh, line 24: I see you removed the requirement for authentication, what is the plan for this endpoint ?

spaenleh (approved):

src/services/item/plugins/itemFlag/index.ts
spaenleh: Mhmmm... okay I thought this was returning all the elements that were flagged, But in reality it just returns the static array of flag types. Maybe adding a small comment would help disambiguate ? In fine we should probably change the name of the endpoint, maybe `/flagTypes` ?

Defferrard: I agree ! I'd name that route `GET /flags/types` I've added this route to https://github.com/graasp/graasp/issues/1288

PR #1330, opened 2024-08-22
refactor(repositories): invitations extends mutable repository

Defferrard:


src/services/item/plugins/invitation/repository.ts
```diff
@@ -1,73 +1,107 @@
...
    * Get invitations for item path and below
    * @param itemPath Item path
    */
-  async getForItem(itemPath: string): Promise<Invitation[]> {
-    return this.createQueryBuilder('invitation')
-      .innerJoinAndSelect('invitation.item', 'item')
-      .where(':path <@ item.path', { path: itemPath })
-      .getMany();
-  },
+  async getManyByItem(itemPath: ItemPath) {
```
ReidyT, line 79: Because we are using `getOne...` to specify the unique returned row, what do you think about using `get...` for the queries that return multiple rows?

```suggestion
  async getByItem(itemPath: ItemPath) {
```

ReidyT (approved):
Good job, LGTM! 🐦‍🔥

pyphilia (approved):

spaenleh (approved):

PR #1332, opened 2024-08-23
refactor(repositories): item like extends mutable repository

Defferrard:


src/services/item/plugins/itemLike/repository.ts
```diff
@@ -1,96 +1,98 @@
...
-    if (!deleteResult.raw.length) {
-      throw new ItemLikeNotFound({ creatorId: creator.id, itemId: item.id });
+    if (!entity) {
+      throw new EntryNotFoundBeforeDeleteException(this.entity);
     }
 
-    return deleteResult.raw[0].id;
-  },
-});
+    super.delete(entity.id);
```
ReidyT, line 94:
```suggestion
    await super.delete(entity.id);
```

ReidyT (approved):
Nice work 👻

src/services/item/plugins/itemLike/index.ts
```diff
@@ -102,7 +102,7 @@ const plugin: FastifyPluginAsync = async (fastify) => {
           },
         };
         await actionService.postMany(member, buildRepositories(manager), request, [action]);
-        return newItemLike;
+        return newItemLike.id;
```
pyphilia, line 105: Need to reflect in schema?

pyphilia (approved):

src/services/item/plugins/itemLike/index.ts
Defferrard: In fact, the schema is already valid. The previous implementation was not respecting the schema

spaenleh (approved):

PR #1333, opened 2024-08-23
refactor(repositories): itemmembership extend mutable repository

Defferrard:


src/services/item/test/fixtures/items.ts
```diff
@@ -152,7 +153,7 @@ export class ItemTestUtils {
     account: Actor;
     permission?: PermissionLevel;
   }) => {
-    return ItemMembershipRepository.save({ item, account, permission });
+    return itemMembershipRawRepository.save({ item, account, permission });
```
ReidyT, line 156: should we await here?

src/services/item/types.ts
```diff
@@ -17,6 +17,17 @@ export enum Ordering {
   DESC = 'DESC',
 }
 
+export function orderingToUpperCase(ordering: Ordering): Ordering.ASC | Ordering.DESC {
```
ReidyT, line 20: not sure to understand the purpose of this

src/services/itemMembership/repository.ts
```diff
@@ -34,14 +35,57 @@ import { mapById } from '../utils';
...
+    const itemsMemberships = memberships.map((m) =>
+      this.repository.create({
+        permission: m.permission,
+        item: { id: getChildFromPath(m.itemPath), path: m.itemPath },
+        account: { id: m.accountId },
+        creator: { id: m.creatorId },
+      }),
+    );
+    await this.repository.insert(itemsMemberships);
+    return itemsMemberships;
```
ReidyT, line 87: When using the `create` method, is the id generated too?

src/services/itemMembership/repository.ts
```diff
@@ -677,25 +728,23 @@ export const ItemMembershipRepository = AppDataSource.getRepository(ItemMembersh
...
+    }
 
     const { path: newParentItemPath } = newParentItem;
     const index = item.path.lastIndexOf('.');
 
     const parentItemPath = index > -1 ? item.path.slice(0, index) : undefined;
     const itemIdAsPath = index > -1 ? item.path.slice(index + 1) : item.path;
 
-    const rows = await this.query(`
+    const rows = (await this.repository.query(`
```
ReidyT, line 742: should we also use query builder here? Or it is not possible?

src/services/itemMembership/repository.ts
```diff
@@ -628,41 +680,40 @@ export const ItemMembershipRepository = AppDataSource.getRepository(ItemMembersh
...
-      deletes: [] as Partial<ItemMembership>[],
+
+    const rows = (await this.repository
+      .createQueryBuilder('item_membership')
+      .select('account_id', 'accountId')
+      .addSelect('max(item_path::text)::ltree', 'itemPath') // Get the longest path
+      .addSelect('max(permission)', 'permission') // Get the best permission
+      .where('item_path @> :path', { path: item.path })
+      .groupBy('account_id')
+      .getRawMany()) as { accountId: string; itemPath: string; permission: PermissionLevel }[];
```
ReidyT, line 694: I would create a type here

src/services/itemMembership/repository.ts
```diff
@@ -715,14 +764,21 @@ export const ItemMembershipRepository = AppDataSource.getRepository(ItemMembersh
...
-    const changes = {
-      inserts: [] as Partial<ItemMembership>[],
+    `)) as {
+      accountId: string;
+      itemPath: string;
+      permission: PermissionLevel;
+      action: number;
+      inherited: PermissionLevel;
+      action2IgnoreInherited: boolean;
+    }[];
```
ReidyT, line 774: I'd create a type here too

ReidyT (approved):
Thanks for this PR, great job! 🙌

spaenleh (approved):
Agree with the other requested changes, nothing to add.

src/services/item/types.ts
Defferrard: Because TypeORM doesn't accept 'asc' or 'desc' as ordering, and our API accept this kind of input (but not `AsC` or `deSc`) we need to normalize it with toUpperCase().
I don't like having a typecast in our repository, so I kinda moved it here.. I've not found any better solution, I'm open to suggestions

src/services/itemMembership/repository.ts
Defferrard: It is, but I perfer doing it in another PR

src/services/item/types.ts
ReidyT: I think it's a good solution in this case :).

src/services/itemMembership/repository.ts
Defferrard: I'm not sure about which `id` you're talking about.
If you're thinking about the ItemMembership ID, then no. It will be generated on the database.

PR #1341, opened 2024-08-26
refactor(repositories): itemloginschema extends mutable repository

Defferrard:


src/services/itemLogin/index.ts
```diff
@@ -96,7 +97,21 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
       assertIsMember(member);
       return db.transaction(async (manager) => {
-        return itemLoginService.put(member, buildRepositories(manager), itemId, type);
+        const repositories = buildRepositories(manager);
+
+        const item = await itemService.get(member, repositories, itemId, PermissionLevel.Admin); // Validate permissions
+        const schema = await itemLoginService.getOneByItem(repositories, item.id);
+        if (schema) {
+          // If exists, then update the existing one
+          return itemLoginService.update(
```
ReidyT, line 106: Should we await here?

```suggestion
          return await itemLoginService.update(
```

src/services/itemLogin/index.ts
```diff
@@ -96,7 +97,21 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+            repositories,
+            schema.id,
+            type ?? ItemLoginSchemaType.Username,
+          );
+        } else {
+          // If not exists, then create a new one
+          return itemLoginService.create(repositories, item.path, type);
```
ReidyT, line 113: Should we await here?

```suggestion
          return await itemLoginService.create(repositories, item.path, type);
```

src/services/itemLogin/index.ts
```diff
@@ -96,7 +97,21 @@ const plugin: FastifyPluginAsync = async (fastify) => {
       const member = notUndefined(user?.account);
       assertIsMember(member);
       return db.transaction(async (manager) => {
```
ReidyT, line 99: I know it's not you, but I'm asking here as a global question. Because we are returning `db.transaction` without awaiting, what would happening if the code crash inside the transaction? Will the error handler catch the exception or will the error cause a server crash?

ReidyT (approved):
LGTM, I just let you some comments and a global question 😶‍🌫️

spaenleh (approved):

PR #1346, opened 2024-08-27
fix(MutableRepository): use getOne in updateOne

ReidyT:
Update the `updateOne` method to use `getOne` instead of `findOne` to ensure to return the updated entity with the same relations as in the `getOne`.


Defferrard (approved):
LGTM ! 🔥

PR #1348, opened 2024-08-27
refactor(ChatMessageRepository): extends MutableRepository

ReidyT:


src/services/chat/index.ts
```diff
@@ -107,13 +113,26 @@ const plugin: FastifyPluginAsync<GraaspChatPluginOptions> = async (fastify) => {
...
+              messageId,
+              body,
+            );
+            await actionChatService.postPatchMessageAction(request, repositories, message);
+            return message;
+          });
+        } catch (e: unknown) {
+          if (e instanceof EntryNotFoundAfterUpdateException || e instanceof EntityNotFound) {
+            throw new ChatMessageNotFound(messageId);
+          }
```
pyphilia, line 133: Should this go a bit lower in the tree, like at the service level?

src/services/chat/index.ts
```diff
@@ -130,12 +149,19 @@ const plugin: FastifyPluginAsync<GraaspChatPluginOptions> = async (fastify) => {
...
+          return await db.transaction(async (manager) => {
+            const repositories = buildRepositories(manager);
+            const message = await chatService.deleteOne(member, repositories, itemId, messageId);
+            await actionChatService.postDeleteMessageAction(request, repositories, message);
+            return message;
+          });
+        } catch (e: unknown) {
+          if (e instanceof EntryNotFoundBeforeDeleteException || e instanceof EntityNotFound) {
+            throw new ChatMessageNotFound(messageId);
+          }
```
pyphilia, line 162: Same as above.

pyphilia (approved):
Thanks for the work, just a question about error handling, maybe just me being confused.

src/services/chat/index.ts
ReidyT: Because it is an HTTP error, I'd catch and rethrow directly in the controller (because it's an HTTP controller). In this way, if we decided for some reason to add a graphQL endpoint for example, we won't mixed the HTTP errors with the graphQL errors (because they'll probably use the same services). But as always, I'm open to the discussion, so I dont know what do you think and if you prefer to throw in the service? @pyphilia, @spaenleh, @Defferrard

Defferrard: To me, I think it's better to throw it in the controller!

Defferrard (approved):
LGTM! Well done 🚀

src/services/chat/index.ts
spaenleh: I agree with the justification for throwing in the controller.

PR #1350, opened 2024-08-27
refactor: notUndefined instanciate the error when necessary

Defferrard:


src/utils/assertions.test.ts
```diff
@@ -1,30 +1,29 @@
-import { notUndefined } from './assertions';
+import { asNonNull } from './assertions';
 
 describe('notUndefined', () => {
```
spaenleh, line 3: Update test name.

src/utils/assertions.test.ts
spaenleh, line 1: Add tests for the `isNonNull` function ?

spaenleh (approved):
Nice !

src/utils/assertions.ts
```diff
@@ -1,16 +1,31 @@
 import { UnexpectedError } from './errors';
 
+export type Nullable<T> = T | null | undefined;
+
+/**
+ *  Returns true if the object is not null or undefined.
+ * @param object Object to check
+ * @returns True if the object is not null or undefined, false otherwise
+ */
+export function isNonNull<T>(object: Nullable<T>): object is T {
```
ReidyT, line 10: What about this suggestion?

```suggestion
export function isDefined<T>(object: Nullable<T>): object is T {
```

src/utils/assertions.ts
```diff
@@ -1,16 +1,31 @@
...
+ * Throw an error if the object is undefined or null. Otherwise, return the object in a not-undefined type.
  * @param object Object to check
- * @param error Error to throw if the object is undefined
+ * @param error Error to throw if the object is undefined or null
  * @returns Object in a not-undefined type
  */
-export function notUndefined<T>(object: T | undefined, error?: Error): T {
-  if (object === undefined) {
-    throw error || new UnexpectedError();
+export function asNonNull<T, Err extends Error, Args extends unknown[]>(
```
ReidyT, line 20: What do you think about:

```suggestion
export function assertIsDefined<T, Err extends Error, Args extends unknown[]>(
```

`asNonNull` doesn't really reflect the throw if null. Also, I'm wondering if we could not merge it with `assertNonNull`?

ReidyT (approved):
Nice PR! LGTM, I just let you some comments 🤖

PR #1352, opened 2024-08-27
fix: membership request out of fastify cors scope

Defferrard:


ReidyT (approved):
LGTM 🚀

Defferrard:
I've created an Issue related to this PR
https://github.com/graasp/graasp/issues/1358

pyphilia (approved):
All good for me it works!

src/services/itemMembership/index.ts
```diff
@@ -28,7 +28,7 @@ const plugin: FastifyPluginAsync = async (fastify) => {
   // schemas
   fastify.addSchema(common);
 
-  fastify.register(MembershipRequestAPI);
+  fastify.register(MembershipRequestAPI, { prefix: '/items/:itemId/memberships/requests' });
```
spaenleh, line 31: As discussed offline, I think this hides the `:itemId` parameter. But we said that a bigger refactor is needed in order to have CORS at only 2 places (for the core api and for the apps API). This means we will have to move the cors instantiation up and extract the apps api from the item plugins.

spaenleh (approved):

PR #1364, opened 2024-08-30
feat: can create a permission on item login without admin's confirmation

Defferrard:


package.json
```diff
@@ -63,7 +63,7 @@
     "@fastify/websocket": "7.2.0",
     "@graasp/etherpad-api": "2.1.1",
     "@graasp/sdk": "4.26.0",
-    "@graasp/translations": "1.35.1",
+    "@graasp/translations": "graasp/graasp-translations#515-failure-message-when-item-login-schema-exists",
```
Defferrard, line 66: reminder.

src/services/item/plugins/invitation/index.ts
```diff
@@ -192,6 +210,59 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+
+        const item = await itemService.get(
+          member,
+          repositories,
+          itemId,
+          PermissionLevel.Read,
+          false,
+        );
+
+        if (!(await itemLoginService.getByItemPath(repositories, item.path))) {
```
ReidyT, line 236: split in two lines? I mean use a variable?

src/services/item/plugins/invitation/index.ts
```diff
@@ -192,6 +210,59 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+          const error = new ItemLoginSchemaNotFound();
+          error.statusCode = StatusCodes.FORBIDDEN;
```
ReidyT, line 238: I'm wondering if we shouldn't throw another error here? ErrorNotFound is a 404, but the status code is updated to 403?

src/services/item/plugins/invitation/index.ts
```diff
@@ -192,6 +210,59 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+        // Check if the member already has a membership, if so, throw an error
+        let hasPermission = true;
+        try {
+          await validatePermission(repositories, PermissionLevel.Read, member, item);
+        } catch (err) {
+          hasPermission = err.statusCode !== StatusCodes.FORBIDDEN;
+        }
+        if (hasPermission) {
+          throw new ItemMembershipAlreadyExists();
+        }
```
ReidyT, line 251: I think you already wrote a code like that in another service few weeks ago. Should we extract this logic in another service to simplify the other controllers? 

```ts
const hasAccess = await hasPermission(...)
if (hasAccess) throw new ItemMembershipAlreadyExists();
```

src/services/item/plugins/invitation/index.ts
```diff
@@ -192,6 +210,59 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+
+        return await itemMembershipService.post(
```
ReidyT, line 253: I hope one day we will rename the methods of the services ^^ `post` --> `create`

src/services/itemMembership/plugins/MembershipRequest/index.ts
```diff
@@ -97,7 +86,11 @@ const plugin: FastifyPluginAsync = async (fastify) => {
           return reply.send({ status: MembershipRequestStatus.Approved });
         }
 
-        return reply.send({ status: MembershipRequestStatus.NotSubmittedOrDeleted });
+        if (await itemService.get(member, repositories, itemId, PermissionLevel.Read, false)) {
```
ReidyT, line 89: extract in a variable?

src/services/itemMembership/plugins/MembershipRequest/index.ts
```diff
@@ -132,8 +125,9 @@ const plugin: FastifyPluginAsync = async (fastify) => {
           PermissionLevel.Read,
           false,
         );
-        if (item.type !== ItemType.FOLDER) {
-          throw new ItemNotFolder({ id: itemId });
+
+        if (await itemLoginService.getByItemPath(repositories, item.path)) {
```
ReidyT, line 129: same here?

ReidyT (approved):
great job! LGTM 🔥

src/services/itemMembership/service.ts
```diff
@@ -110,13 +110,15 @@ export class ItemMembershipService {
     actor: Account,
     repositories: Repositories,
     membership: { permission: PermissionLevel; itemId: UUID; memberId: UUID },
+    throwOnForbiddenPermission?: boolean,
```
pyphilia, line 113: I see you start to need these a lot, should we try to move all `validatePermission` in the controller? to avoid creating many booleans to bypass verification.

src/services/item/plugins/invitation/index.ts
pyphilia: I would indeed throw sth like "cannot enroll" 403

src/services/item/plugins/invitation/index.ts
```diff
@@ -192,6 +210,59 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+        if (!(await itemLoginService.getByItemPath(repositories, item.path))) {
+          const error = new ItemLoginSchemaNotFound();
+          error.statusCode = StatusCodes.FORBIDDEN;
+          throw error;
+        }
+
+        // Check if the member already has a membership, if so, throw an error
+        let hasPermission = true;
+        try {
+          await validatePermission(repositories, PermissionLevel.Read, member, item);
```
pyphilia, line 245: Perhaps you want to look for memberships directly? This function will not throw if the item is public but the member has no membership. The question is also, do we want to let users to enroll if the item is public? probably yes.

src/services/item/plugins/invitation/schema.ts
```diff
@@ -159,3 +159,22 @@ export const sendOne = {
     },
   },
 };
+
+export const enroll = {
+  tags: ['itemMemberships'],
+  summary: 'Create an item membership for the logged in user if there is an Item Login',
+  description: `Create an item membership on the item with the given ID for the logged in user. 
+    The item be associated with an Item Login.`,
```
pyphilia, line 167:
```suggestion
    The item is associated with an Item Login.`,
```
not sure what it means with the remaining of the sentence tho? You want to add that it is allowed if the item has an item login right?

pyphilia (changes requested):

spaenleh (changes requested):

src/services/itemMembership/service.ts
Defferrard: To easily solve that, I could ask an item id/path. I prefer to wait the refactor of ItemMembership before fixing it that way.

Otherwise, moving all the verification in the controller need big changes. I've made an Issue for that #1371, I'll fix it in another PR. Thank you for the feedback !

src/services/item/plugins/invitation/index.ts
```diff
@@ -192,6 +210,52 @@ const plugin: FastifyPluginAsync = async (fastify) => {
       reply.status(StatusCodes.NO_CONTENT);
     },
   );
+
+  fastify.post<{ Params: { itemId: string } }>(
+    '/:itemId/enroll',
```
pyphilia, line 215: I'm just using this branch for updating my front for now, but should this be in membership instead? or item login? It was not trivial for me that it is in invitation.

spaenleh (approved):

pyphilia (changes requested):

src/services/itemMembership/plugins/MembershipRequest/index.ts
```diff
@@ -137,18 +145,14 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
 
-        // Check if the member already has a membership, if so, throw an error
-        let hasPermission = true;
-        try {
-          await validatePermission(repositories, PermissionLevel.Read, member, item);
-        } catch (err) {
-          hasPermission = err.statusCode !== StatusCodes.FORBIDDEN;
+        const itemLoginSchema = await itemLoginService.getByItemPath(repositories, item.path);
+        if (itemLoginSchema) {
+          throw new ItemLoginSchemaExists();
```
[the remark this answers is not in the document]
Defferrard: On private _create function of itemMembership service, it delete the eventual membership request that exists

pyphilia (approved):

PR #1365, opened 2024-08-30
feat: cannot get item login schema type of hidden item

Defferrard:


src/services/itemLogin/index.ts
```diff
@@ -34,9 +39,34 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+          itemId,
+          PermissionLevel.Read,
+          false,
+        );
+
+        const tags = await itemTagService.getByItemPath(repositories, item.path);
+        const isHidden = Boolean(tags.find((t) => t.type === ItemTagType.Hidden));
+        if (
+          isHidden &&
+          (user?.account == undefined ||
```
ReidyT, line 56:
```suggestion
          (!user?.account ||
```

src/services/itemLogin/index.ts
```diff
@@ -34,9 +39,34 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+            !(await itemMembershipService.getByAccountAndItem(
+              repositories,
+              user?.account?.id,
+              item.id,
```
ReidyT, line 60: Extract this in a variable to improve clarity ^^

ReidyT (approved):
LGTM, good job!

pyphilia (changes requested):

src/services/itemLogin/index.ts
```diff
@@ -34,9 +39,34 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+        );
+
+        const tags = await itemTagService.getByItemPath(repositories, item.path);
+        const isHidden = Boolean(tags.find((t) => t.type === ItemTagType.Hidden));
+        if (
+          isHidden &&
+          (user?.account == undefined ||
+            !(await itemMembershipService.getByAccountAndItem(
+              repositories,
+              user?.account?.id,
```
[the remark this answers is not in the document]
spaenleh: Yes because it is an `||` (or) so the account can still be undefined.

spaenleh (approved):
Looks good, I think a small refactoring of the condition would be nice, either in temporary variables or split across multiple nested control statements.

src/services/itemLogin/index.ts
Defferrard: Since `user` has type `any`, there's a way that `account` is a boolean. (Which should never happen...). So I prefer to be explicit here. Otherwise I can move this check in an utils function

src/services/itemLogin/index.ts
Defferrard: If I extract we'll always make a request to the Database. It could also expose us to some issues if this request fail and throws an exception when we don't need to.

I'll add some comments !

src/services/itemLogin/index.ts
ReidyT: Okay so in this case
```suggestion
          (user?.account === undefined ||
```

spaenleh (approved):
LGTM !

pyphilia:
You can resolve conflict and merge!

pyphilia (approved):

PR #1373, opened 2024-09-12
feat: use max ZIP file size from SDK

ReidyT:
Because the Builder and Backend used to different size, the backend also use the ZIP size defined in the SDK.


spaenleh (approved):

src/services/item/plugins/importExport/index.ts
```diff
@@ -33,7 +32,7 @@ const plugin: FastifyPluginAsync = async (fastify) => {
       // fieldNameSize: 0,             // Max field name size in bytes (Default: 100 bytes).
       // fieldSize: 1000000,           // Max field value size in bytes (Default: 1MB).
       fields: 0, // Max number of non-file fields (Default: Infinity).
-      fileSize: DEFAULT_MAX_FILE_SIZE, // For multipart forms, the max file size (Default: Infinity).
```
spaenleh, line 36: Remove this one from the constants ?

ReidyT: This default limit is used by other services like upload items, thumbnails, etc. That's why I didn't removed it. Do you think we should use the same limit for all the uploads?

PR #1395, opened 2024-09-18
docs: move house keeping  behavior

Defferrard:
Just some docs that I wrote last Friday, so we don't lose theses explanations in the future.

### DIFF
--- a/src/services/itemMembership/repository.ts
+++ b/src/services/itemMembership/repository.ts
@@ -683,17 +683,25 @@ export class ItemMembershipRepository extends MutableRepository<
   // UTILS
 
   /**
+   * Retrieves all memberships related to the ancestors of the given item and returns the memberships
+   * we need to create so that accounts with permissions further up the tree maintain their best permissions
+   * on the item after it has been moved to the root.
+   *
    * Identify any new memberships that will be necessary to create
    * after moving the item from its parent item to *no-parent*.
    *
    * Moving to *no-parent* is simpler so this method is used instead of `moveHousekeeping()`.
    * @param item Item that will be moved
    * @param account Member used as `creator` for any new memberships
+   * @returns Object with `inserts` and `deletes` arrays of memberships to create and delete after moving the item to the root. `deletes` will always be empty.
    */
   async detachedMoveHousekeeping(item: Item, account: Account) {
+    // Get the Id of the item when it will be moved to the root
     const index = item.path.lastIndexOf('.');
     const itemIdAsPath = item.path.slice(index + 1);
 
+    // For each account that belongs to an ancestor of the element,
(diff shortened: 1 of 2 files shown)


pyphilia (approved):

src/services/itemMembership/repository.ts
```diff
@@ -683,17 +683,26 @@ export class ItemMembershipRepository extends MutableRepository<
   // UTILS
 
   /**
+   * Get all memberships related to the the given item's ancestors,
+   * and return the memberships we need to create so the accounts that have permissions higher in the tree
+   * keeps their best permission on the item after it is moved to the root.
+   *
```
ReidyT, line 689:
```suggestion
   * Retrieves all memberships related to the ancestors of the given item and returns the memberships we need to create so that accounts with permissions further up the tree maintain their best permissions on the item after it has been moved to the root.
```

src/services/itemMembership/repository.ts
```diff
@@ -683,17 +683,26 @@ export class ItemMembershipRepository extends MutableRepository<
...
+   * @returns Object with `inserts` and `deletes` arrays of memberships to create and delete after moving the item to the root. `deletes` will always be empty.
    */
   async detachedMoveHousekeeping(item: Item, account: Account) {
+    // Get the Id of the item when it will be moved to the root
     const index = item.path.lastIndexOf('.');
     const itemIdAsPath = item.path.slice(index + 1);
 
+    // For each account that has a membership on an ancestor of the item
+    // Show his Best Permission
+    // and the Path of the deepest Item (The closest to the item).
```
ReidyT, line 705:
```suggestion
    // For each account that belongs to an ancestor of the element, 
    // retrieve its best permission and the path to the deepest element (closest to the element).
```

src/services/itemMembership/repository.ts
```diff
@@ -740,14 +749,18 @@ export class ItemMembershipRepository extends MutableRepository<
...
+      // Moving to the root
       return this.detachedMoveHousekeeping(item, account);
     }
 
     const { path: newParentItemPath } = newParentItem;
     const index = item.path.lastIndexOf('.');
 
-    const parentItemPath = index > -1 ? item.path.slice(0, index) : undefined;
-    const itemIdAsPath = index > -1 ? item.path.slice(index + 1) : item.path;
+    // If the a '.' has been found (>= 0) in item's path, it means it has a parent
```
ReidyT, line 759:
```suggestion
    // If a '.' has been found (>= 0) in item's path, this means that it has a parent.
```

ReidyT (approved):

spaenleh:
See #1404 for a fix of the sonarcloud issues.

PR #1398, opened 2024-09-18
feat: adds thumbnails urls in PackedItem

ReidyT:


src/services/item/plugins/recycled/index.ts
```diff
@@ -32,7 +34,7 @@ const plugin: FastifyPluginAsync<RecycledItemDataOptions> = async (fastify, opti
   const { db, websockets } = fastify;
   const { maxItemsInRequest = MAX_TARGETS_FOR_READ_REQUEST } = options;
 
-  const recycleBinService = new RecycledBinService();
+  const recycleBinService = new RecycledBinService(resolveDependency(ItemThumbnailService));
```
Defferrard, line 37: I'm wondering if there is a specific reason of why `RecycledBinService` is not a dependency ? Could we inject the needed parameters instead of doing it here ?

Defferrard, line 37: I'm wondering if there is a specific reason of why RecycledBinService is not a dependency ? Could we inject the needed parameters instead of doing it here ?

src/services/authorization.ts
```diff
@@ -305,6 +309,7 @@ export const filterOutPackedDescendants = async (
   repositories,
```
Defferrard, line 309: I know it's not yours but I'm triggered by this one
```suggestion
  repositories: Repositories,
```

src/services/item/plugins/thumbnail/service.ts
```diff
@@ -66,6 +77,53 @@ export class ItemThumbnailService {
...
+      return {};
+    }
+
+    const itemsIdWithThumbnail = items
+      .filter((i) => Boolean(i.settings.hasThumbnail))
+      .map((i) => i.id);
+    Object.values(ThumbnailSize);
+    const thumbnailSizes = Object.values(ThumbnailSizeInPackedItem);
+
+    // TODO: check if optimization is possible (bulk fetch for example)
```
Defferrard, line 101: Should we create an Issue for that, so we don't keep TODO in the code ?

Defferrard (approved):
LGTM ! Good work ! I just have some minor comment

src/services/item/plugins/recycled/index.ts
ReidyT: Yeah I think it's a good idea :)

src/services/item/plugins/thumbnail/service.ts
```diff
@@ -66,6 +77,52 @@ export class ItemThumbnailService {
...
+   *
+   * **PLEASE NOTE** : Permissions on items must be validated before calling this method!
+   *
+   * @param items The items for which thumbnails are to be retrieved. Only thumbnails of items with `hasThumbnail` setting will be retrieved.
+   * @returns An object whose keys are the item id and whose values are the URLs stored by size.
+   * */
+  async getManyUrls(
+    items: (Pick<Item, 'id'> & { settings: Pick<Item['settings'], 'hasThumbnail'> })[],
+  ) {
+    if (!items || items.length === 0) {
```
pyphilia, line 91:
```suggestion
    if (!items?.length) {
```

src/services/item/fluent-schema.ts
```diff
@@ -92,7 +92,8 @@ export const packedItem = item
       .prop('createdAt', S.string())
       .prop('type', S.string())
       .prop('item', item),
-  );
+  )
+  .prop('thumbnails', S.oneOf([S.null(), S.object().additionalProperties(true)]));
```
pyphilia, line 96: Add more details in the object? Since we know which sizes we return.

src/services/item/plugins/geolocation/index.test.ts
```diff
@@ -95,6 +129,33 @@ describe('Item Geolocation', () => {
         expectPackedItem(result.item, packedItem!);
       });
 
+      it.only('Get geolocation with thumbnails', async () => {
```
pyphilia, line 132: Remove only.

pyphilia (changes requested):
LGTM, maybe I would rename `getManyUrls`, as it suppose it's always for packed items, and we don't know with this name that you get many urls from **many** items.

src/services/item/plugins/thumbnail/service.ts
```diff
@@ -66,6 +77,52 @@ export class ItemThumbnailService {
...
+      return {};
+    }
+
+    const itemsIdWithThumbnail = items
+      .filter((i) => Boolean(i.settings.hasThumbnail))
+      .map((i) => i.id);
+    Object.values(ThumbnailSize);
```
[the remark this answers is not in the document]
ReidyT: Nope, thanks I forgot to remove it.

src/services/item/plugins/thumbnail/const.ts
```diff
@@ -0,0 +1,3 @@
+import { ThumbnailSizeInPackedItem } from '@graasp/sdk';
```
pyphilia, line 1: Rename this file to `constants.ts`

pyphilia (approved):
LGTM! Just a file rename and it's all good 🐦

PR #1401, opened 2024-09-19
fix: set jest extension to run on-demand instead of watch

spaenleh:
@Defferrard I think this fixes the issue where the tests are run as soon as you start the container. 

I had this configuration set in my personnal settings, I put it in the shared settings so it can be applied to everyone using it.


Defferrard (approved):
LGTM! Thank you for sharing this !

PR #1404, opened 2024-09-20
chore: update code to fix sonar cloud issue

### LINKED ISSUE #1403: [REFACTOR] Reduce return value should be used
https://sonarcloud.io/project/issues?pullRequest=1395&open=AZIFOp3Raj5kJiILYjG2&id=graasp_graasp

This code has several issues:
- we do not use the return value from the `reduce`
- we use instead the constant that is modified ... which is really not a good practice.

spaenleh:
@Defferrard this should fix your sonarCloud issues. It should also improve the code's reliability.

fix #1403

### DIFF
--- a/src/services/itemMembership/repository.ts
+++ b/src/services/itemMembership/repository.ts
@@ -33,7 +33,7 @@ import { itemMembershipSchema } from '../member/plugins/export-data/schemas/sche
 import { schemaToSelectMapper } from '../member/plugins/export-data/utils/selection.utils';
 import { mapById } from '../utils';
 import { ItemMembership } from './entities/ItemMembership';
-import { getPermissionsAtItemSql } from './utils';
+import { PermissionType, getPermissionsAtItemSql } from './utils';
 
 type ItemPath = Item['path'];
 type AccountId = Account['id'];
@@ -703,11 +703,11 @@ export class ItemMembershipRepository extends MutableRepository<
       .groupBy('account_id')
       .getRawMany()) as DetachedMoveHousekeepingType[];
 
-    const changes = rows.reduce<ResultMoveHousekeeping>(
-      (chngs, row) => {
+    return rows.reduce<ResultMoveHousekeeping>(
+      (changes, row) => {
         const { accountId, itemPath, permission } = row;
         if (itemPath !== item.path) {
-          chngs.inserts.push({
+          changes.inserts.push({
             accountId,
             itemPath: itemIdAsPath,
             permission,
@@ -715,15 +715,13 @@ export class ItemMembershipRepository extends MutableRepository<
           });
         }
 
-        return chngs;
+        return changes;
       },
       {
         inserts: [],
         deletes: [],
       },
     );
-
-    return changes;
   }
 
   /**
@@ -764,7 +762,7 @@ export class ItemMembershipRepository extends MutableRepository<
           account_id,
           '${newParentItemPath}'::ltree AS item_path,
           max(permission) AS permission,
-          0 AS action -- 0: inherited at destination (no action)
+          ${PermissionType.InheritedAtDestination} AS action -- 0: inherited at destination (no action)
         FROM item_membership
         WHERE item_path @> '${newParentItemPath}'
         GROUP BY account_id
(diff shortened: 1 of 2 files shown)


src/services/itemMembership/repository.ts
```diff
@@ -703,24 +703,25 @@ export class ItemMembershipRepository extends MutableRepository<
...
-      const { accountId, itemPath, permission } = row;
-      if (itemPath !== item.path) {
-        chngs.inserts.push({
-          accountId,
-          itemPath: itemIdAsPath,
-          permission,
-          creatorId: account.id,
-        });
-      }
+    const changes = rows.reduce<ResultMoveHousekeeping>(
```
ReidyT, line 706:
```suggestion
    return rows.reduce<ResultMoveHousekeeping>(
```

src/services/itemMembership/repository.ts
```diff
@@ -703,24 +703,25 @@ export class ItemMembershipRepository extends MutableRepository<
...
+      (chngs, row) => {
```
ReidyT, line 707:
```suggestion
      (changes, row) => {
```

src/services/itemMembership/repository.ts
```diff
@@ -776,48 +777,49 @@ export class ItemMembershipRepository extends MutableRepository<
...
+        if (action === 2 && action2IgnoreInherited) {
+          return chngs;
+        }
 
-      // permission worse or equal to inherited one at "destination"
-      if (action === 2 && PermissionLevelCompare.lte(permission, inherited)) {
-        chngs.deletes.push({ accountId, itemPath });
-      }
+        const permission = p;
+        const inherited = ip;
```
ReidyT, line 799: Why it is not possible to use the `permission` and `inherited` directly from the `row`?
```suggestion
```

src/services/itemMembership/repository.ts
```diff
@@ -776,48 +777,49 @@ export class ItemMembershipRepository extends MutableRepository<
...
+    const changes = rows.reduce<ResultMoveHousekeeping>(
+      (chngs, row) => {
+        const {
           accountId,
           itemPath,
-          permission,
-          creatorId: accountId,
-        });
-      }
+          permission: p,
```
ReidyT, line 785:
```suggestion
          permission,
```

src/services/itemMembership/repository.ts
```diff
@@ -776,48 +777,49 @@ export class ItemMembershipRepository extends MutableRepository<
...
+          action,
+          inherited: ip,
```
ReidyT, line 787:
```suggestion
          inherited,
```

src/services/itemMembership/repository.ts
```diff
@@ -776,48 +777,49 @@ export class ItemMembershipRepository extends MutableRepository<
...
-      }
-
-      const permission = p;
-      const inherited = ip;
-
-      // permission (inherited) at the "origin" better than inherited one at "destination"
-      if (action === 1 && PermissionLevelCompare.gt(permission, inherited)) {
-        chngs.inserts.push({
+    const changes = rows.reduce<ResultMoveHousekeeping>(
+      (chngs, row) => {
```
ReidyT, line 781:
```suggestion
   return rows.reduce<ResultMoveHousekeeping>(
      (changes, row) => {
```

src/services/itemMembership/repository.ts
```diff
@@ -776,48 +777,49 @@ export class ItemMembershipRepository extends MutableRepository<
...
-          creatorId: accountId,
-        });
-      }
+          permission: p,
+          action,
+          inherited: ip,
+          action2IgnoreInherited,
+        } = row;
+
+        if (action === 0) {
```
ReidyT, line 791: What action 0, 1, or 2 means? Should we use an enum here?

ReidyT (changes requested):
Thanks for this PR, LGTM! I just let some suggestions to avoid using misspelling variables (like `chnges`) and to avoid using magic numbers. Let me know if you disagree 😆.

src/services/itemMembership/repository.ts
spaenleh: I have no f***!ng idea 🤷

spaenleh:
@ReidyT I have made changes according to your comment. Let me know If you think this is clearer... I honestly do not really understand what this actually does, so i tried to improve the code as best as I could.

ReidyT (approved):
Thanks for your changes, LGTM 🚀

ReidyT:
> @ReidyT I have made changes according to your comment. Let me know If you think this is clearer... I honestly do not really understand what this actually does, so i tried to improve the code as best as I could.

Yes it's clearer, I know it was small issues, but thanks for your changes 👍

PR #1406, opened 2024-09-20
fix: do not allow special characters in username

### LINKED ISSUE #1311: Do not allow special characters in username input
Currently, the username `<<@@#!%^&*()__+>>` is completely valid in the system. I definitely do not think that it's a good idea. The username string should not allow special characters on input.

MatveyK:
Fixes #1311

Related PRs : https://github.com/graasp/graasp-account/pull/356 https://github.com/graasp/graasp-auth/pull/457

### DIFF
--- a/src/services/auth/plugins/magicLink/magicLink.test.ts
+++ b/src/services/auth/plugins/magicLink/magicLink.test.ts
@@ -1,3 +1,4 @@
+import { faker } from '@faker-js/faker';
 import { ReasonPhrases, StatusCodes } from 'http-status-codes';
 import { sign } from 'jsonwebtoken';
 import fetch from 'node-fetch';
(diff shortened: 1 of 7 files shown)


Defferrard (approved):
LGTM ! Nice work 🚀  I'm wondering about the regex, should we be more restrictive and whitelisting the characters instead of a blacklist ?

With the current regex, this following is a valid username : `-.,🍔æ¶{}`

src/services/auth/plugins/magicLink/magicLink.test.ts
```diff
@@ -234,6 +234,22 @@ describe('Auth routes tests', () => {
       expect(response.statusMessage).toEqual(ReasonPhrases.BAD_REQUEST);
       expect(response.statusCode).toEqual(StatusCodes.BAD_REQUEST);
     });
+
+    it('Bad request if the username contains special characters', async () => {
+      const email = 'test@tust.lol';
```
Defferrard, line 239:
```suggestion
      const email = faker.internet.email().toLowerCase();
```

MatveyK: Nice one ! Faker internet is my favourite internet.

MatveyK:
@Defferrard Yeah, this was definitely mentioned. I've approached this issue from a security standpoint more than anything. Since we do not use the username as a unique identifier, I think we can be quite permissive with it, as long is it doesn't pose security problems.

Also, even if we define a whitelist, how do we know which characters to include in it ? People's names can contain all sorts of weird characters, and it seems like a challenge.

ReidyT (approved):
Nice PR! LGTM 🎃

MatveyK:
@pyphilia Added the mobile check.

This PR is all done.

PR #1414, opened 2024-09-23
feat: optimize thumbnails caching

ReidyT:


src/services/file/fixtures.ts
```diff
@@ -0,0 +1,37 @@
+import { FileServiceUrlCaching } from './caching';
```
pyphilia, line 1: I think you can also mock with a folder like `_mock_` https://jestjs.io/docs/manual-mocks so it's automatic maybe that's what you want?

src/services/file/caching.ts
```diff
@@ -0,0 +1,49 @@
+import { hoursToSeconds } from 'date-fns';
```
pyphilia, line 1: Looks like this could be just a Redis service? Maybe like defining a constructor with `expiringTime` `key`, `buildKey`, `callback` and it could be reuse for anything, right?

src/services/file/fixtures.ts
Defferrard, line 1: All test files should be in a `test` folder, or be name with `*.test.*`. Look at `.dockerignore` for more information.

src/services/file/service.ts
```diff
@@ -96,21 +104,25 @@ class FileService {
...
-    );
+    const getUrl = () =>
+      this.repository.getUrl(
+        {
+          expiration,
+          filepath,
+          id,
+        },
+        this.logger,
+      );
```
Defferrard, line 115: Could this be in a private function ?

Defferrard (approved):
LGTM ! Just have some comments !

src/services/file/service.ts
ReidyT: I can put it into a private function, but there is no advantage compared to store it in a const.

```
private async t({
    expiration,
    filepath,
    id,
  }: {
    expiration?: number;
    id?: string;
    filepath: string;
  }) {
    return this.repository.getUrl(
      {
        expiration,
        filepath,
        id,
      },
      this.logger,
    );
  }

  async getUrl(data: { expiration?: number; id?: string; path?: string }): Promise<string> {
    const { expiration, id, path: filepath } = data;
    if (!filepath || !id) {
      throw new DownloadFileInvalidParameterError({
        filepath,
        id,
      });
    }

    return (
      this.caching?.getOrCache(filepath, () => this.t({ expiration, filepath, id }), expiration) ??
      this.t({ expiration, filepath, id })
    );
  }
```

What do you think @Defferrard?

Defferrard: It is mainly to improve the readability. However, I'm okay with both solutions

src/services/file/caching.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Redis } from 'ioredis';
+
+import { IServiceCaching } from './types';
+
+const EXPIRES_IN_SECONDS = hoursToSeconds(1);
+
+export class ServiceCaching implements IServiceCaching {
```
pyphilia, line 8:
```suggestion
export class CachingService implements ICachingService {
```

pyphilia: Maybe now you can have this service out of `/file`? Same for the interface.

pyphilia (changes requested):

pyphilia (approved):

spaenleh:
@all-contributors please add @ReidyT for code, testing, documentation and bugs

PR #1417, opened 2024-09-24
feat: migrate guest account

Defferrard:


src/migrations/1727167973931-migrate-guest-account.ts
```diff
@@ -0,0 +1,27 @@
...
+        is_validated = false, 
+        item_login_schema_id = item_login_schema.id 
+    
+    FROM item_membership 
+        INNER JOIN item_login_schema 
+            ON item_login_schema.item_path = item_membership.item_path 
+    WHERE item_membership.account_id = account.id 
+    AND account.email ~* '^[0-9a-f]{4}-[0-9]{10,}@graasp.org' 
+    AND account.type = 'individual'
+        `,
```
Defferrard, line 22:
```suggestion
    AND account.type = 'individual'
    AND (
      item_membership.account_id = item_membership.creator_id
      OR item_membership.creator_id = '12345678-1234-1234-1234-123456789012'
    )
        `,
```

pyphilia (approved):

src/migrations/1727167973931-migrate-guest-account.ts
```diff
@@ -0,0 +1,31 @@
...
+    
+    FROM item_membership 
+        INNER JOIN item_login_schema 
+            ON item_login_schema.item_path = item_membership.item_path 
+    WHERE item_membership.account_id = account.id 
+    AND account.email ~* '^[0-9a-f]{4}-[0-9]{10,}@graasp.org' 
+    AND account.type = 'individual'
+    AND (
+      item_membership.account_id = item_membership.creator_id
+      OR item_membership.creator_id = '12345678-1234-1234-1234-123456789012'
```
ReidyT, line 24: What is this creator id? Maybe add a comment to explain.

src/migrations/1727167973931-migrate-guest-account.ts
ReidyT, line 1: I don't know the email address of the Graasper account, but have we check that this account is not transformed into a guest account?

ReidyT (changes requested):
LGTM, I just let you some questions 🧑‍🚀

src/migrations/1727167973931-migrate-guest-account.ts
pyphilia: I checked, should be all good. Normally graasper is not supposed to have an item login schema (it has a legit graasp email).

src/migrations/1727167973931-migrate-guest-account.ts
pyphilia: At the very beginning of Graasp, public things were created by Graasper (this id). We don't use this fake member anymore.

ReidyT: Thanks for this explanation

ReidyT (approved):
Thanks for the explanations and changes, LGTM 🚀

PR #1419, opened 2024-09-24
fix: can delete item login schema

Defferrard:


src/services/itemLogin/index.ts
```diff
@@ -167,8 +167,13 @@ const plugin: FastifyPluginAsync = async (fastify) => {
       return db.transaction(async (manager) => {
         const member = asDefined(user?.account);
         assertIsMember(member);
+        const repositories = buildRepositories(manager);
+
+        // Validate permission
+        await itemService.get(user?.account, repositories, itemId, PermissionLevel.Admin);
```
pyphilia, line 173: Use `validatePermission`?

src/services/itemLogin/repositories/itemLoginSchema.ts
```diff
@@ -68,4 +72,21 @@ export class ItemLoginSchemaRepository extends MutableRepository<
 
     return await super.insert({ item: { id: getChildFromPath(itemPath), path: itemPath }, type });
   }
+
+  async deleteOneByItemId(itemId: ItemId) {
+    this.throwsIfParamIsInvalid('itemId', itemId);
+
+    const entity = await this.getOneByItem(itemId);
```
pyphilia, line 79:
```suggestion
    const entity = await this.getOneByItemId(itemId);
```

src/services/itemLogin/index.ts
```diff
@@ -167,8 +167,13 @@ const plugin: FastifyPluginAsync = async (fastify) => {
...
+
         try {
-          return (await itemLoginService.delete(member, buildRepositories(manager), itemId)).id;
+          return (await itemLoginService.delete(member, repositories, itemId)).id;
```
ReidyT, line 176:
```suggestion
          const { id } = await itemLoginService.delete(member, repositories, itemId);
          return id;
```

ReidyT (changes requested):
Thanks for this PR 🐍, I agree with what Kim said.

PR #1426, opened 2024-09-26
fix: guest account migration

Defferrard:


pyphilia (approved):
Maybe add a comment why we need this migration, and why this is so complicated.

ReidyT (approved):
Great job 👑 LGTM 🚀

PR #1441, opened 2024-10-02
refactor(schemas): remove refs of invitation's schemas

Defferrard:


spaenleh (approved):
I have nothing special to say, looks good 🛴

src/services/item/test/fixtures/items.ts
```diff
@@ -168,12 +170,16 @@ export class ItemTestUtils {
...
+    const { item, permission, creator, parentItem } = options;
+    let member = options.member;
     const newItem = await this.saveItem({
       item,
       actor: creator ?? member,
       parentItem,
     });
+    if (!member) {
+      member = await saveMember();
+    }
```
ReidyT, line 182: Why should we save the member if not null? Also, do you know why the member can be undefined? I know it's not you who initially allowed undefined member...

ReidyT: Maybe it's just because the actor type in the tests is `Member | null`.

ReidyT (approved):
LGTM I just let one question 🫠

src/services/item/test/fixtures/items.ts
Defferrard: I've created a use for that :)

PR #1446, opened 2024-10-02
feat: remove packed from own recycled, add pagination

### LINKED ISSUE #572: `GET /recycled` is slow
We might have to add indexes on the `item_path` to not do a full table scan.

This should improve performance.

pyphilia:
close #572

### DIFF
--- a/src/services/item/plugins/recycled/index.ts
+++ b/src/services/item/plugins/recycled/index.ts
@@ -2,10 +2,10 @@ import { StatusCodes } from 'http-status-codes';
 
 import { FastifyPluginAsync } from 'fastify';
 
-import { MAX_TARGETS_FOR_READ_REQUEST } from '@graasp/sdk';
+import { MAX_TARGETS_FOR_READ_REQUEST, Pagination } from '@graasp/sdk';
 
 import { resolveDependency } from '../../../../di/utils';
-import { IdParam, IdsParams } from '../../../../types';
+import { IdsParams } from '../../../../types';
 import { asDefined } from '../../../../utils/assertions';
 import { buildRepositories } from '../../../../utils/repositories';
 import { isAuthenticated } from '../../../auth/plugins/passport';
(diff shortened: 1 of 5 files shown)


src/services/item/utils.ts
```diff
@@ -133,3 +139,95 @@ export const getPostItemPayloadFromFormData = (
...
+        break;
+      case SortBy.ItemName:
+        mappedSortBy = `${item}.name`;
+        break;
+    }
+    if (mappedSortBy) {
+      query.orderBy(mappedSortBy, orderingToUpperCase(ordering));
+    }
+  }
+};
```
pyphilia, line 233: What do you think of this function? @spaenleh @Defferrard @MatveyK @ReidyT  I used already twice this process (accessible, children) and will have to use it for recycled items too.

spaenleh: Doesn't this function do a bit too much ??

pyphilia: I can cut in pieces, but then call all of them every time? Have a wrapped function?

src/services/item/plugins/recycled/index.ts
```diff
@@ -28,11 +29,18 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+    },
+    async ({ user, query }) => {
       const member = asDefined(user?.account);
       assertIsMember(member);
-      const result = await recycleBinService.getAll(member, buildRepositories());
+      const { page = 1, pageSize = ITEMS_PAGE_SIZE } = query;
+      const result = await recycleBinService.getOwn(member, buildRepositories(), {
+        page,
+        pageSize,
+      });
```
Defferrard, line 43: I'm wondering if it is mandatory to define the default values here since you've already defined it in the schema ?
```suggestion
      const result = await recycleBinService.getOwn(member, buildRepositories(), query);
```

src/services/item/plugins/recycled/schemas.ts
```diff
@@ -28,28 +29,45 @@ export const recycledItemSchemaRef = registerSchemaAsRef(
...
-      creator: nullableMemberSchemaRef,
-      createdAt: customType.DateTime(),
+      page: Type.Number({
+        minimum: 0,
+        default: 1,
+      }),
+      pageSize: Type.Number({
+        minimum: 1,
+        default: ITEMS_PAGE_SIZE,
+      }),
```
Defferrard, line 44: There is a `customType.Pagination(...)`, does it fill your needs ?

Defferrard (approved):
LGTM!

src/services/item/plugins/recycled/service.ts
```diff
@@ -28,28 +28,16 @@ export class RecycledBinService {
...
-      const item = packedItems.find((i) => i.id === r.item.id);
-      // should never pass here
-      if (!item) {
-        throw new Error(`item should be defined`);
-      }
-      return { ...r, item };
-    });
+    const recycled = await recycledItemRepository.getOwn(member, pagination);
+
+    return recycled;
```
ReidyT, line 40:
```suggestion
    return await recycledItemRepository.getOwn(member, pagination);
```

ReidyT (approved):
LGTM 🎇

PR #1460, opened 2024-10-02
refactor(schemas): remove refs of member's schemas

Defferrard:


src/plugins/typebox.ts
```diff
@@ -18,11 +18,19 @@ const schemas: TSchema[] = [];
 /**
  * Register a schema in the Fastify instance.
  * This function needs to be called before the server starts, so the schema will be registered in the Fastify instance.
- * Each schema needs to have defined the `$id` property.
+ * Each schema needs to have an unique `$id` property.
```
spaenleh, line 21:
```suggestion
 * Each schema needs to have a unique `$id` property.
```

src/server.ts
```diff
@@ -36,6 +36,7 @@ const start = async () => {
         // This allow routes that take array to correctly interpret single values as an array
         // https://github.com/fastify/fastify/blob/main/docs/Validation-and-Serialization.md
         coerceTypes: 'array',
+        discriminator: true,
```
spaenleh, line 39: Add an explanation of what this does ?

src/services/account/schemas.ts
```diff
@@ -4,44 +4,76 @@ import { AccountType } from '@graasp/sdk';
...
+    $id: 'augmentedAccount',
+  },
+);
+
+export const nullableAugmentedAccountSchemaRef = registerSchemaAsRef(
+  Type.Union([
+    // The augmented account can either be an individual or a guest
+    customType.Nullable(compositeMemberAccountSchema),
+    customType.Nullable(compositeGuestAccountSchema),
+  ]),
```
spaenleh, line 74: You do not need to set the discriminator here ?

src/services/item/plugins/publication/validation/schemas.ts
```diff
@@ -10,40 +12,16 @@ export const status = {
 
 export const itemValidation = {
   params: {
-    itemId: {
-      $ref: 'https://graasp.org/#/definitions/uuid',
-    },
+    itemId: customType.UUID(),
```
spaenleh, line 15: use the schema ref here ?

src/services/item/plugins/publication/validation/schemas.ts
```diff
@@ -10,40 +12,16 @@ export const status = {
...
-
 export const itemValidationGroup = {
   params: {
-    itemId: {
-      $ref: 'https://graasp.org/#/definitions/uuid',
-    },
-    itemValidationGroupId: {
-      $ref: 'https://graasp.org/#/definitions/uuid',
-    },
+    itemId: customType.UUID(),
```
spaenleh, line 23: Here too ?

src/services/item/schema.ts
```diff
@@ -48,23 +52,25 @@ export const itemSchemaRef = registerSchemaAsRef(
       id: customType.UUID(),
       name: Type.String(),
       displayName: Type.String(),
-      description: Type.Optional(Type.Union([Type.String(), Type.Null()])),
+      description: Type.Optional(customType.Nullable(Type.String())),
```
spaenleh, line 55: They need to be optional ? because they are already nullable.

src/services/item/schema.ts
```diff
@@ -48,23 +52,25 @@ export const itemSchemaRef = registerSchemaAsRef(
...
       type: Type.String(),
       path: Type.String(),
       lang: Type.String(),
       extra: Type.Object({}, { additionalProperties: true }),
       settings: Type.Object({}, { additionalProperties: true }),
-      creator: Type.Optional(Type.Ref('https://graasp.org/members/#/definitions/member')),
+      creator: Type.Optional(nullableAccountSchemaRef),
```
spaenleh, line 61: Same question for whether the optional is required here

src/services/member/schemas.ts
```diff
@@ -34,74 +34,91 @@ import { FILE_METADATA_DEFAULT_PAGE_SIZE, FILE_METADATA_MIN_PAGE } from './const
...
-    },
+const memberSchema = Type.Object(
+  {
+    // Object definition
+    id: customType.UUID(),
+    name: Type.String({
+      format: 'username',
+      minLength: MIN_USERNAME_LENGTH,
+      maxLength: MAX_USERNAME_LENGTH,
+    }),
```
spaenleh, line 45: Should this be a custom type or a re-usable schema or even just a variable that is copy-pasted in the openapi, but can be changed everywhere in the code at the same time.

spaenleh (approved):
Really nice ! I like that you were able to enforce the id and title by using the function signature !

src/services/member/schemas.ts
Defferrard: It will be transformed to customType.Username() in the nexts PR's !

src/server.ts
```diff
@@ -36,6 +36,8 @@ const start = async () => {
         // This allow routes that take array to correctly interpret single values as an array
         // https://github.com/fastify/fastify/blob/main/docs/Validation-and-Serialization.md
         coerceTypes: 'array',
+        // This enables the use of discriminator keyword in oneof or anyof in schemas so it optimize validation.
```
pyphilia, line 39:
```suggestion
        // This enables the use of discriminator keyword in oneof or anyof in schemas so it optimizes validation.
```

src/plugins/typebox.ts
```diff
@@ -18,11 +18,19 @@ const schemas: TSchema[] = [];
...
- * Each schema needs to have defined the `$id` property.
+ * Each schema needs to have a unique `$id` property.
  * @param schema The schema to be registered.
+ * @param options The options for the schema.
  * @returns The schema passed as argument.
  */
-export function registerSchemaAsRef<T extends TSchema>(schema: T): TRef<T> {
+export function registerSchemaAsRef<T extends TSchema>(
+  schema: T,
+  options: { $id: string; title: string },
```
pyphilia, line 28: I would reverse the args, or even split like `id, title, schema` unless options are optional? But my feeling is that you usually give an id, then a title, and finally the content.

src/plugins/typebox.ts
```diff
@@ -18,11 +18,19 @@ const schemas: TSchema[] = [];
...
+): TRef<T> {
+  // Set schema options
+  schema.$id = options.$id;
+  schema.title = options.title;
```
pyphilia, line 32: Why was it not working to have `$id` and `title` inside schema directly?

src/services/chat/index.ts
```diff
@@ -74,13 +77,18 @@ const plugin: FastifyPluginAsyncTypebox<GraaspChatPluginOptions> = async (fastif
...
         } = request;
-        const member = asDefined(user?.account);
-        return await db.transaction(async (manager) => {
-          const repositories = buildRepositories(manager);
-          const message = await chatService.postOne(member, repositories, itemId, body);
-          await actionChatService.postPostMessageAction(request, repositories, message);
-          return message;
-        });
+        const account = asDefined(user?.account);
+        if (isMember(account) || isGuest(account)) {
```
pyphilia, line 81: Is there a way to factor this out (with the throw)?  Looks like we will need this often.

src/services/chat/service.ts
```diff
@@ -32,12 +33,11 @@ export class ChatMessageService {
...
     await this.itemService.get(actor, repositories, itemId);
 
-    const messages = await chatMessageRepository.getByItem(itemId);
-    return messages;
+    return await chatMessageRepository.getByItem(itemId);
   }
 
   async postOne(
-    actor: Account,
+    actor: Guest | Member,
```
pyphilia, line 40: Question: so we can never use `Account`?

pyphilia (changes requested):

src/plugins/typebox.ts
Defferrard: Firstly because, they're not required in the schema.
Also, because of the `nullable` issue and some other case (B.e : We can not use `Type.Composite([])` with a `Ref`), we're forced to be able to extract the schema from the ref. So we want to avoid a schema to be used at multiple place with an `$id`. The `$id` property should only be present in `components.schemas` in the OpenAPI JSON, and should not be present in `paths`.

src/services/chat/service.ts
Defferrard: We should never be able to use an account that is neither a guest or a member.
I think it would be betther to pass the `account.id` as parameter here.

package.json
```diff
@@ -64,7 +64,7 @@
     "@fastify/websocket": "7.2.0",
     "@graasp/etherpad-api": "2.1.1",
     "@graasp/sdk": "4.31.0",
-    "@graasp/translations": "1.39.0",
+    "@graasp/translations": "graasp/graasp-translations#555-add-failure-message-for-notmemberorguest",
```
Defferrard, line 67: reminder.

pyphilia (approved):

PR #1467, opened 2024-10-04
fix: save membership on item login schema item on item login

### LINKED ISSUE #1466: [BUG] Item login request on child does not create item login on parent
## Current behaviour 

If you make a login request on the child of an item login, the permission for the user will be created on the child instead of being created on the parent as would be expected.

## Correct behaviour

When making a login request on the child of an item login, the permission should be created on the item that has the item login set.

pyphilia:
!!! NEED TO REBASE

`test.yml` got changed to run tests.

fix #1466

### DIFF
--- a/src/services/itemLogin/test/index.test.ts
+++ b/src/services/itemLogin/test/index.test.ts
@@ -24,6 +24,7 @@ import { encryptPassword } from '../../auth/plugins/password/utils';
 import { Item } from '../../item/entities/Item';
 import { ItemTag } from '../../item/plugins/itemTag/ItemTag';
 import { ItemTestUtils } from '../../item/test/fixtures/items';
+import { ItemMembership } from '../../itemMembership/entities/ItemMembership';
 import { Member } from '../../member/entities/member';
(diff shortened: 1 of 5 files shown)


src/services/itemLogin/test/index.test.ts
```diff
@@ -477,6 +479,40 @@ describe('Item Login Tests', () => {
...
+        const member = res.json();
+
+        // membership is saved on the right path
+        const membership = await rawItemMembershipRepository.findOne({
+          where: { account: { id: member.id } },
+          relations: { item: true, account: true },
+        });
+        expect(membership?.item.path).toEqual(itemLoginSchema.item.path);
+
+        expect(res.statusCode).toBe(StatusCodes.OK);
```
spaenleh, line 512: Test this before using the response ? if the status code is not OK, it is likely that the json will not be a member.

spaenleh (approved):
Looks good to me. We can create a tag from this branch or simply use this branch as the release for the quick fix to deploy.

PR #1470, opened 2024-10-08
refactor(schemas): infer type in auth service

Defferrard:


src/services/auth/plugins/magicLink/index.ts
```diff
@@ -24,23 +24,14 @@ import { MagicLinkService } from './service';
...
-    Body: {
-      name: string;
-      email: string;
-      captcha: string;
-      url?: string;
-      enableSaveActions?: boolean;
-    };
-    Querystring: { lang?: string };
-  }>(
+  fastify.post(
```
ReidyT, line 34: Really nice! Well more readable.

src/services/auth/plugins/magicLink/schemas.ts
```diff
@@ -1,72 +1,53 @@
...
-      captcha: { type: 'string' },
-      url: {
-        type: 'string',
-        format: 'uri',
-      },
-      enableSaveActions: { type: 'boolean' },
+      }),
+      email: Type.String({ format: 'email' }),
+      captcha: Type.String(),
+      url: Type.Optional(Type.String({ format: 'uri' })),
```
ReidyT, line 17: I know it was already like that, but should we rename the format by `url` instead of `uri`? As the property is an url and not an uri?

src/services/auth/plugins/magicLink/schemas.ts
```diff
@@ -1,72 +1,53 @@
...
-        format: 'jwt',
-      },
-      url: {
-        type: 'string',
-        format: 'uri-reference',
-      },
+  querystring: Type.Object(
+    {
+      [SHORT_TOKEN_PARAM]: Type.String({ format: 'jwt' }),
+      url: Type.Optional(Type.String({ format: 'uri-reference' })),
```
ReidyT, line 49: same here?

ReidyT (approved):
Thanks for this PR! LGTM! I just let some comments ^^

src/services/auth/plugins/magicLink/schemas.ts
```diff
@@ -1,72 +1,53 @@
...
+    { additionalProperties: false },
+  ),
+  querystring: Type.Object(
+    {
+      lang: Type.Optional(Type.String()),
     },
-    additionalProperties: false,
-  },
+    { additionalProperties: false },
+  ),
```
spaenleh, line 43: Not sure about the usage of this ? What are we doing with this ?

src/services/auth/plugins/magicLink/schemas.ts
spaenleh, line 49: Why is this `uri-reference` and the other is simply `uri` ? What is the difference, should we use only one ?

src/services/auth/plugins/mobile/schemas.ts
```diff
@@ -1,81 +1,84 @@
...
-  querystring: {
-    type: 'object',
-    properties: {
-      lang: { type: 'string' },
+    { additionalProperties: false },
+  ),
+  querystring: Type.Object(
+    {
+      lang: Type.Optional(Type.String()),
     },
```
spaenleh, line 46: Same question about the usage of the language here.

src/services/auth/plugins/password/schemas.ts
```diff
@@ -1,134 +1,109 @@
...
+  querystring: Type.Object(
+    {
+      lang: Type.Optional(Type.String()),
     },
-    additionalProperties: false,
-  },
-};
+    { additionalProperties: false },
+  ),
+} as const satisfies FastifySchema;
```
spaenleh, line 28: I see not all schemas are annotated with this, is it necessary, or can it be omitted ?

src/services/auth/plugins/password/schemas.ts
```diff
@@ -1,134 +1,109 @@
...
-    properties: {
-      authorization: { type: 'string', format: 'bearer' },
+  body: Type.Object(
+    {
+      password: Type.String({ format: 'strongPassword' }),
     },
-  },
+    { additionalProperties: false },
+  ),
+  headers: Type.Object({ authorization: Type.String({ format: 'bearer' }) }),
```
spaenleh, line 95: nice !

spaenleh (commented):
Looks good ! I have left a few comments. let's see.

spaenleh (approved):

src/services/auth/plugins/password/schemas.ts
Defferrard: It can be omitted, but I think it is a good practice to mention it, so we have autocompletion when editing the schema. Also there was some schema that has the properties `query` instead of `querystring`...

src/services/auth/plugins/magicLink/schemas.ts
Defferrard: I've made a task in the big refactor issue

src/services/auth/plugins/magicLink/schemas.ts
Defferrard: same answer

PR #1472, opened 2024-10-08
refactor(schemas): infer type in membership service

Defferrard:


ReidyT (approved):
LGTM! I'm just wondering why we have all properties of the error schema as optional? If an error occurred, we should have at least the statusCode always defined right?

spaenleh (approved):

Defferrard:
> LGTM! I'm just wondering why we have all properties of the error schema as optional? If an error occurred, we should have at least the statusCode always defined right?

I've added a task to https://github.com/graasp/graasp/issues/1288

PR #1474, opened 2024-10-08
refactor(schemas): infer type in itemlogin service

Defferrard:


src/services/itemLogin/schemas.ts
```diff
@@ -1,83 +1,88 @@
...
-  .prop('item', item)
-  .prop('createdAt', S.string())
-  .prop('updatedAt', S.string())
-  .prop('id', uuid);
+export const credentialsSchemaRef = registerSchemaAsRef(
+  'itemLoginCredentials',
+  'Item Login Credentials',
+  Type.Object(
+    {
+      username: Type.String({ minLength: 3, maxLength: 50, pattern: '^\\S+( \\S+)*$' }),
```
ReidyT, line 17: I don't know if we want to extract the pattern in the same const for the username and password for allowing their to evolve independently, but should we extract the patterns in varioble or other, to simplify the readability of the Regex?

src/services/itemLogin/schemas.ts
```diff
@@ -1,83 +1,88 @@
...
+  'Item Login Schema',
+  itemLoginSchemaSchema,
+);
 
 export const login = {
-  params: idParam,
-  querystring: S.object().additionalProperties(false).prop('m', S.boolean()),
-  body: credentials,
+  params: entityIdSchemaRef,
+  body: credentialsSchemaRef,
```
ReidyT, line 45: What happens to the query string?

ReidyT (approved):
LGTM

pyphilia (approved):
LGTM 💪

src/services/itemLogin/schemas.ts
Defferrard: It is not used, so I felt free to remove it!

src/services/itemLogin/schemas.ts
Defferrard: It'll use the customType username in a future PR 👍🏻

PR #1478, opened 2024-10-08
refactor(schemas): infer type in item plugins' service

Defferrard:


spaenleh (approved):
Nice ! LGTM 🏑

src/services/item/plugins/shortLink/schemas.ts
```diff
@@ -1,49 +1,78 @@
...
-};
+);
 
-const create = {
-  body: {
-    ...shortLinkPayload,
-    required: ['alias', 'platform', 'itemId'],
+const shortLinkPayload = Type.Object(
+  {
+    alias: Type.String({ minLength: 6, maxLength: 255, pattern: '^[a-zA-Z0-9-]*$' }),
```
ReidyT, line 24: Should we reuse the constraint to avoid to change it twice?

ReidyT (approved):
LGTM 🫕

PR #1479, opened 2024-10-08
fix: app actions structure is wrong for export

spaenleh:
In this PR: I fix the structure of the apps file in the exported zip.

Now the structure is as follows:
 - In JSON, there will be a single file `apps.json` that will contain:
    ```json
    {
      "actions": [ array of actions ],
      "data": [ array of data ],
      "settings": [ array of data ],
    }
    ```
- In CSV, there will be 3 files, `app_actions.csv`, `app_data.csv` and `app_settings.csv` which will contain tabular data from all items.

### DIFF
--- a/src/services/action/utils/export-utils.test.ts
+++ b/src/services/action/utils/export-utils.test.ts
@@ -6,6 +6,9 @@ import { writeFileForFormat } from './export';
 
 const writeFileSyncMock = jest.spyOn(fs, 'writeFileSync').mockImplementation(() => {});
 describe('writeFileForFormat', () => {
+  beforeEach(() => {
+    jest.resetAllMocks();
+  });
   describe('CSV format', () => {
     it('Empty array should not write anything', () => {
       writeFileForFormat('./test.csv', ExportActionsFormatting.CSV, []);
--- a/src/services/action/utils/export-utils.test.ts
+++ b/src/services/action/utils/export-utils.test.ts
@@ -12,7 +12,7 @@ describe('writeFileForFormat', () => {
       expect(writeFileSyncMock).not.toHaveBeenCalled();
     });
 
-    it('Object should not write anything', () => {
+    it('Empty object should not write anything', () => {
       writeFileForFormat('./test.csv', ExportActionsFormatting.CSV, {});
       expect(writeFileSyncMock).not.toHaveBeenCalled();
     });
@@ -43,7 +43,7 @@ describe('writeFileForFormat', () => {
       expect(writeFileSyncMock).not.toHaveBeenCalled();
     });
 
-    it('Object should not write anything', () => {
+    it('Empty object should not write anything', () => {
       writeFileForFormat('./test.json', ExportActionsFormatting.JSON, {});
       expect(writeFileSyncMock).not.toHaveBeenCalled();
     });
--- a/src/services/action/utils/export-utils.test.ts
+++ b/src/services/action/utils/export-utils.test.ts
@@ -1,0 +1,66 @@
+import fs from 'fs';
+
+import { ExportActionsFormatting } from '@graasp/sdk';
+
+import { writeFileForFormat } from './export';
+
+const writeFileSyncMock = jest.spyOn(fs, 'writeFileSync').mockImplementation(() => {});
+describe('writeFileForFormat', () => {
+  describe('CSV format', () => {
+    it('Empty array should not write anything', () => {
+      writeFileForFormat('./test.csv', ExportActionsFormatting.CSV, []);
+      expect(writeFileSyncMock).not.toHaveBeenCalled();
+    });
+
+    it('Object should not write anything', () => {
+      writeFileForFormat('./test.csv', ExportActionsFormatting.CSV, {});
+      expect(writeFileSyncMock).not.toHaveBeenCalled();
+    });
+
+    it('Array should be written', () => {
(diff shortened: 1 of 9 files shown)


src/services/action/utils/export-utils.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const writeFileSyncMock = jest.spyOn(fs, 'writeFileSync').mockImplementation(() => {});
+describe('writeFileForFormat', () => {
+  describe('CSV format', () => {
+    it('Empty array should not write anything', () => {
+      writeFileForFormat('./test.csv', ExportActionsFormatting.CSV, []);
+      expect(writeFileSyncMock).not.toHaveBeenCalled();
+    });
+
+    it('Object should not write anything', () => {
```
ReidyT, line 15:
```suggestion
    it('Empty object should not write anything', () => {
```

src/services/action/utils/export-utils.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    });
+  });
+
+  describe('JSON format', () => {
+    it('Empty array should not write anything', () => {
+      writeFileForFormat('./test.json', ExportActionsFormatting.JSON, []);
+      expect(writeFileSyncMock).not.toHaveBeenCalled();
+    });
+
+    it('Object should not write anything', () => {
```
ReidyT, line 46:
```suggestion
    it('Empty object should not write anything', () => {
```

src/services/action/utils/export.ts
```diff
@@ -40,14 +43,15 @@ const flattenObject = (obj: RecursiveObject, prefix: string = ''): ReturnObject
   }, {});
 };
 
-const writeFileForFormat = (
+export const writeFileForFormat = (
   path: string,
   format: ExportActionsFormatting,
-  data: object[], // TODO: Replace 'object' with specific type(s) when defining schema for data to export
+  // TODO: Replace 'object' with specific type(s) when defining schema for data to export
```
ReidyT, line 49: Should we create an issue?

src/services/action/utils/export.ts
```diff
@@ -130,24 +141,58 @@ export const exportActionsInArchive = async (args: {
...
+      case ExportActionsFormatting.JSON: {
+        // create files for the apps
+        const appsPath = path.join(
+          fileFolderPath,
+          buildActionFileName('apps', archiveDate, format),
+        );
+        writeFileForFormat(appsPath, format, { appActions, appData, appSettings });
+        break;
+      }
+      // For CSV format there will be one file for actions, one fofr data and one for settings
```
ReidyT, line 170:
```suggestion
      // For CSV format there will be one file for actions, one for data and one for settings
```

src/services/action/utils/export.ts
```diff
@@ -130,24 +141,58 @@ export const exportActionsInArchive = async (args: {
...
+          fileFolderPath,
+          buildActionFileName('app_data', archiveDate, format),
+        );
+        writeFileForFormat(appDataPath, format, appData);
+
+        const appSettingsPath = path.join(
+          fileFolderPath,
+          buildActionFileName('app_settings', archiveDate, format),
+        );
+        writeFileForFormat(appSettingsPath, format, appSettings);
```
ReidyT, line 190: Should we refactor in another function to then call it for the three CSV exports, as well as for the JSON export?

src/services/item/plugins/action/service.ts
```diff
@@ -207,12 +207,12 @@ export class ActionItemService {
...
     } = {};
-    for (const { id: appId } of [item, ...descendants].filter(
-      ({ type }) => type === ItemType.APP,
-    )) {
+    const appItems = [item, ...descendants].filter(({ type }) => type === ItemType.APP);
+    for (const { id: appId } of appItems) {
       const appData = await repositories.appDataRepository.getForItem(
         appId,
         {},
+        // needs investigating: does this mean a reader could export the actions of an item and see all users responses ?
```
ReidyT, line 215: Only admin should be allowed to do that right?

ReidyT (approved):
LGTM

src/services/action/utils/export.ts
spaenleh: I don't know, is the function too lengthy like this ? Not sure refactoring to a different function will make it more readable.

src/services/item/plugins/action/service.ts
spaenleh: Add an issue for this.

swouf: Yes, we cannot allow a reader to get access to other users' data with the export. We may give them the possibility to export their own data.

src/services/action/utils/export.ts
ReidyT: I'm okay with what your did too ^^

swouf (approved):

src/services/item/plugins/action/schemas.ts
```diff
@@ -128,7 +130,7 @@ export const getAggregateActions = {
   },
 };
 
-export const exportAction = {
+export const exportAction: FastifySchema = {
```
Defferrard, line 133: If we want to be able to infer type in the handler we should not type the schema to FastifySchema. You should do `as const satisfies FastifySchema`

This schema will be refactored anyway

Defferrard (approved):

PR #1489, opened 2024-10-09
refactor(schemas): infer type in member's service

Defferrard:


src/services/member/plugins/profile/schemas.ts
```diff
@@ -1,66 +1,70 @@
...
 
-const memberSharedSchema = S.object()
-  .additionalProperties(false)
-  .prop('bio', S.string())
-  .prop('facebookID', S.anyOf([S.string(), S.null()]))
-  .prop('linkedinID', S.anyOf([S.string(), S.null()]))
-  .prop('twitterID', S.anyOf([S.string(), S.null()]));
+const inputProfileMemberSchemaRef = registerSchemaAsRef(
+  'memberShared',
+  'Member Shared',
```
spaenleh, line 11: What do you think of this naming ?
```suggestion
  'publicProfile',
  'Public Profile',
```

src/services/member/plugins/profile/schemas.ts
```diff
@@ -1,66 +1,70 @@
...
-export const getProfileForMember: FastifySchema = {
-  params: S.object().additionalProperties(false).prop('memberId', uuid),
+export const getProfileForMember = {
+  params: Type.Object({ memberId: customType.UUID() }, { additionalProperties: false }),
   response: {
-    [StatusCodes.OK]: profileMember,
-    [StatusCodes.NO_CONTENT]: S.null(),
-    [StatusCodes.UNAUTHORIZED]: error,
+    [StatusCodes.OK]: profileMemberSchemaRef,
+    [StatusCodes.NO_CONTENT]: Type.Null(),
```
spaenleh, line 51: I think we should annotate why this uses a 204 instead of a 404. I think it has something to do with how the query-client reacts to 4xx codes. They are treated as errors, but we want it to be a null return with success. Not sure this is good pattern since I was asking myself if this should not have been a 404 code.

src/services/member/plugins/profile/schemas.ts
```diff
@@ -1,66 +1,70 @@
...
+} as const satisfies FastifySchema;
 
-export const getOwnProfile: FastifySchema = {
+export const getOwnProfile = {
   response: {
-    [StatusCodes.OK]: profileMember,
-    [StatusCodes.NO_CONTENT]: S.null(),
-    [StatusCodes.UNAUTHORIZED]: error,
+    [StatusCodes.OK]: profileMemberSchemaRef,
+    [StatusCodes.NO_CONTENT]: Type.Null(),
```
spaenleh, line 59: I guess same here ?

src/services/member/plugins/thumbnail/schemas.ts
```diff
@@ -1,42 +1,29 @@
...
-import { entityIdSchemaRef, errorSchemaRef } from '../../../../schemas/global';
+import { customType } from '../../../../plugins/typebox';
+import { errorSchemaRef } from '../../../../schemas/global';
 
 const upload = {
-  params: {
-    type: 'object',
-    additionalProperties: false,
-  },
+  params: Type.Object({}, { additionalProperties: false }),
```
spaenleh, line 10: ? no params ?

spaenleh (approved):

spaenleh (approved):

ReidyT (approved):
LGTM! Maybe we should have to return 404 if no profile is found, and having a foreign key in the member to its profile. In this way, we could enable/disable the query-client's query if the fk is null (like we did for the thumbnails).

I don't know if it's a good idea, but anyway, this is not the purpose of this PR.

PR #1490, opened 2024-10-09
refactor: set up vitest for one test

pyphilia:
- setup vitest config
- rename `utils.test.ts` -> `utils.vitest.ts` The goal is to move all tests bit by bit and rename everything at the end. It is necessary if we want to keep jest tests in the meantime.
- typed some columns for typeorm to work in esm (when running tests).


src/services/item/entities/Item.ts
```diff
@@ -103,6 +105,7 @@ export class Item<T extends ItemTypeEnumKeys = ItemTypeEnumKeys> extends BaseEnt
     default: ItemType.FOLDER,
     nullable: false,
     enum: Object.values(ItemType),
+    type: 'varchar',
```
spaenleh, line 108: For this, it matches the DB type but actually we should maybe have it as an enum ? Not to change now, but would have been better typed for the db I think.

src/services/item/plugins/geolocation/ItemGeolocation.ts
```diff
@@ -34,10 +35,10 @@ export class ItemGeolocation extends BaseEntity {
   @Column({ type: 'float' })
   lng: number;
 
-  @Column({ length: 300, nullable: true })
-  addressLabel: string;
+  @Column({ length: 300, nullable: true, type: 'varchar' })
+  addressLabel: Relation<string>;
```
spaenleh, line 39: This is not a relation as far as I can see.

src/services/member/entities/member.ts
```diff
@@ -32,20 +33,23 @@ export class Member extends Account {
   @Column({
     nullable: true,
     name: 'user_agreements_date',
+    type: 'date',
```
spaenleh, line 36: Same comment here, use `timestamp`

vite.config.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export default defineConfig({
+  test: {
+    include: ['**/*.vitest.ts'],
+    coverage: {
+      reporter: ['lcov', 'text'],
+    },
+  },
+  resolve: {
+    alias: [{ find: '@/', replacement: fileURLToPath(new URL('./', import.meta.url)) }],
```
spaenleh, line 13: Do we need this ? also currently it is a shorthand for the project root, not `src` which is usually the goal.

vite.config.ts
```diff
@@ -0,0 +1,15 @@
+import { URL, fileURLToPath } from 'url';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
spaenleh, line 2: Why do we need this ignore ? it should be in the devdependencies which are excluded from this rule.

spaenleh (changes requested):
Nice ! See my comments, we can discuss the issues you faced. 

This looks promising ! 🚤 ✅

Defferrard (approved):
I agree with @spaenleh, nothing to add, LGTM !

PR #1492, opened 2024-10-09
fix: register double encoding uri component

Defferrard:


spaenleh (approved):
Nice ! Looks good to me, test is also 💯

src/services/auth/plugins/magicLink/magicLink.test.ts
```diff
@@ -418,4 +418,38 @@ describe('Auth routes tests', () => {
...
+      });
+      expect(responseRegister.statusCode).toBe(StatusCodes.NO_CONTENT);
+
+      let m = await memberRawRepository.findOneBy({ email });
+      expect(m?.lastAuthenticatedAt).toBeNull();
+      expect(m?.isValidated).toBeFalsy();
+
+      expect(mockSendEmail).toHaveBeenCalledTimes(1);
+      const fetchedURL = new URL(mockSendEmail.mock.calls[0][2]);
+      const authURL = fetchedURL.pathname + fetchedURL.search;
```
spaenleh, line 443: This should also work.

Maybe add a comment about why this stringification was needed. 
```suggestion
      const fetchedURL = new URL(mockSendEmail.mock.calls[0][2]);
      const authURL = fetchedURL.toString();
```

spaenleh (approved):
Looks good !

pyphilia (approved):

PR #1499, opened 2024-10-10
fix: remove dependency to fix deploy

spaenleh:
Bump resolution for `sodium-native` to `4.2.0`

Remove dependency `is-url-http` and use `url-http` instead as it requires an up-to-date version of `re2` which does not fail the build in docker

### DIFF
--- a/src/services/item/plugins/embeddedLink/utils.ts
+++ b/src/services/item/plugins/embeddedLink/utils.ts
@@ -1,7 +1,7 @@
-import isUrlHttp from 'is-url-http';
+import urlHttp from 'url-http';
 
-export const isValidUrl = (url: string) => {
-  return isUrlHttp(ensureProtocol(url));
+export const isValidUrl = (url: string): boolean => {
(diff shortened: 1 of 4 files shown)


src/services/item/plugins/embeddedLink/utils.ts
```diff
@@ -1,7 +1,7 @@
-import isUrlHttp from 'is-url-http';
+import urlHttp from 'url-http';
 
-export const isValidUrl = (url: string) => {
-  return isUrlHttp(ensureProtocol(url));
+export const isValidUrl = (url: string): boolean => {
+  return !!urlHttp(ensureProtocol(url));
```
ReidyT, line 4: Should we use `Boolean` instead of double negation here?

```suggestion
  return Boolean(urlHttp(ensureProtocol(url)));
```

ReidyT (approved):
LGTM, I just let a suggestion ^^

pyphilia (approved):

PR #1511, opened 2024-10-11
refactor(schemas): infer type in item's service

Defferrard:


src/services/item/validation.ts
```diff
@@ -1,12 +1,36 @@
...
+  .prop('displayCoEditors', S.boolean())
+  .prop('descriptionPlacement', S.enum(Object.values(DescriptionPlacement)))
+  .prop('isCollapsible', S.boolean())
+  .prop('enableSaveActions', S.boolean())
+  // link settings
+  .prop('showLinkIframe', S.boolean())
+  .prop('showLinkButton', S.boolean())
+  // file settings
+  .prop('maxWidth', S.enum(Object.values(MaxWidth)))
+  .prop('alignment', S.enum(Object.values(Alignment)));
```
pyphilia, line 30: Is it still needed for these schemas to be in fluent? Can't we move fully to typebox?
Or I think this is already defined in another file?

src/services/item/service.ts
```diff
@@ -95,7 +96,7 @@ export class ItemService {
     member: Member,
     repositories: Repositories,
     args: {
-      item: Partial<Item> & Pick<Item, 'name' | 'type'>;
+      item: Partial<Omit<Item, 'geolocation'>> & Pick<Item, 'name' | 'type'>;
```
pyphilia, line 99: Maybe we can set `geolocation` to `null` or `undefined`, otherwise I don't understand why you we shouldn't omit `creator` for example.

src/services/item/schemas.ts
```diff
@@ -0,0 +1,575 @@
...
+      {
+        geolocation: Type.Optional(
+          Type.Object(
+            {
+              lat: Type.Number(),
+              lng: Type.Number(),
+            },
+            { additionalProperties: false },
+          ),
+        ),
```
pyphilia, line 159: I think a schemaRef has been defined for that?

src/services/item/schemas.ts
```diff
@@ -0,0 +1,575 @@
...
+      linkItemCreateSchema,
+      localFileItemCreateSchema,
+      s3FileItemCreateSchema,
+      etherpadItemCreateSchema,
+      h5pItemCreateSchema,
+      shortcutItemCreateSchema,
+    ],
+    'type',
+  ),
+);
```
pyphilia, line 297: Maybe it makes sense to move all "createSchema" related code to its own file?
Or maybe we agree on having an endpoint for each type it's not necessary.

src/services/item/schemas.ts
```diff
@@ -0,0 +1,575 @@
...
+        maxItems: MAX_TARGETS_FOR_MODIFY_REQUEST,
+        uniqueItems: true,
+      }),
+    },
+    { additionalProperties: false },
+  ),
+  body: itemUpdateSchemaRef,
+  response: {
+    [StatusCodes.OK]: Type.Array(Type.Intersect([itemSchemaRef, errorSchemaRef])),
+    [StatusCodes.ACCEPTED]: Type.Array(customType.UUID()), // ids > MAX_TARGETS_FOR_MODIFY_REQUEST_W_RESPONSE
```
pyphilia, line 515: What about the comment?

src/services/item/schemas.ts
```diff
@@ -0,0 +1,575 @@
...
+    {
+      id: Type.Array(customType.UUID(), {
+        maxItems: MAX_TARGETS_FOR_MODIFY_REQUEST,
+        uniqueItems: true,
+      }),
+    },
+    { additionalProperties: false },
+  ),
+  response: {
+    [StatusCodes.ACCEPTED]: Type.Array(customType.UUID()), // ids > MAX_TARGETS_FOR_MODIFY_REQUEST_W_RESPONSE
```
pyphilia, line 540: Same as above.

src/services/item/schemas.ts
```diff
@@ -0,0 +1,575 @@
...
+);
+const shortcutItemCreateSchema = itemCreateSchemaFactory(
+  ItemType.SHORTCUT,
+  Type.Object({ target: customType.UUID() }, { additionalProperties: false }),
+);
+
+export const augmentedItemCreateSchemaRef = registerSchemaAsRef(
+  'augmentedItemCreate',
+  'Augmented Item Create',
+  customType.Discrimineable(
```
pyphilia, line 283: Question: So this is a oneOf but infer the type in the endpoints then, right?

pyphilia (changes requested):
Thanks for this PR! We're finally getting rid of the hacky schemas.
Let me know if you have questions!

src/services/item/validation.ts
Defferrard: It was defined in another file. I suggest to use TypeCompiler from Typebox which should have the same effect. I think it is better to do in another PR
(Source)[https://github.com/sinclairzx81/typebox?tab=readme-ov-file#typecheck-typecompiler]

src/services/item/schemas.ts
```diff
@@ -0,0 +1,575 @@
...
+    createdAt: customType.DateTime(),
+    updatedAt: customType.DateTime(),
+  },
+  {
+    // Schema Options
+    additionalProperties: false,
+  },
+);
+
+export const itemSchemaRef = registerSchemaAsRef('item', 'Item', itemSchema);
```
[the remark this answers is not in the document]
Defferrard: I can't because I'm using `itemSchema` for some other schemas by picking properties

pyphilia (approved):

spaenleh (approved):

PR #1521, opened 2024-10-15
refactor: use one build per tests

pyphilia:
ref #1516

### DIFF
--- a/src/services/item/plugins/importExport/test/index.test.ts
+++ b/src/services/item/plugins/importExport/test/index.test.ts
@@ -3,12 +3,18 @@ import fs from 'fs';
 import { StatusCodes } from 'http-status-codes';
 import fetch from 'node-fetch';
 import path from 'path';
+import { ILike } from 'typeorm';
 import waitForExpect from 'wait-for-expect';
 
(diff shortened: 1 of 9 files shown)


src/services/item/plugins/importExport/test/index.test.ts
```diff
@@ -3,12 +3,18 @@ import fs from 'fs';
 import { StatusCodes } from 'http-status-codes';
 import fetch from 'node-fetch';
 import path from 'path';
+import { ILike } from 'typeorm';
```
Defferrard, line 6: Unused import

Defferrard, line 6: Unused import

Defferrard (approved):
LGTM!

src/services/item/plugins/importExport/test/index.test.ts
spaenleh, line 1: I see that you have changed the behaviour of the test to import inside a folder. Does this mean we do not test importing at the root anymore ? we could check that the import is done at the root with a query on the accessible for the user.

spaenleh (approved):

ReidyT (approved):
LGTM

PR #1523, opened 2024-10-16
docs: complete mobile schema

pyphilia:
I started completing the schemas, let me know what you think ^^

### DIFF
--- a/src/services/auth/plugins/mobile/schemas.ts
+++ b/src/services/auth/plugins/mobile/schemas.ts
@@ -3,10 +3,31 @@ import { StatusCodes } from 'http-status-codes';
 
 import { FastifySchema } from 'fastify';
 
-import { customType } from '../../../../plugins/typebox';
+import { customType, registerSchemaAsRef } from '../../../../plugins/typebox';
+import { errorSchemaRef } from '../../../../schemas/global';
 import { SHORT_TOKEN_PARAM, TOKEN_PARAM } from '../passport';
 
+const authTokensPairSchemaRef = registerSchemaAsRef(
+  'tokensPair',
+  'Tokens Pair',
+  Type.Object(
+    {
+      authToken: Type.String(),
+      refreshToken: Type.String(),
+    },
+    {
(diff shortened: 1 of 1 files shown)


src/services/auth/plugins/mobile/schemas.ts
```diff
@@ -52,24 +83,46 @@ export const mPasswordLogin = {
...
 } as const satisfies FastifySchema;
 
 export const authWeb = {
+  tags: ['authentication', 'mobile'],
+  summary: 'Authentication on the web with mobile token',
+  description:
+    'Obtain session cookie to authenticate on the web given provided mobile JWT token. Redirect to given url.',
   querystring: Type.Object(
     {
       [TOKEN_PARAM]: Type.String({ format: 'jwt' }),
```
Defferrard, line 119: Also, we can add a small comment to each params to describe what it is. [Sources...](https://quobix.com/vacuum/rules/descriptions/oas3-parameter-description/)

I don't think we should put a lot of effort on this one, but can still be relevant.
I'm also wondering, what about querystrings/body.. We should probably also add comments, but didn't found any rules about it.

src/services/auth/plugins/mobile/schemas.ts
```diff
@@ -3,10 +3,31 @@ import { StatusCodes } from 'http-status-codes';
...
+  'tokensPair',
+  'Tokens Pair',
+  Type.Object(
+    {
+      authToken: Type.String(),
+      refreshToken: Type.String(),
+    },
+    {
+      // Schema Options
+      additionalProperties: false,
```
Defferrard, line 20: It is only a recommandation, but we can add a small description to each `SchemaRef`. [Source](https://quobix.com/vacuum/rules/descriptions/component-description/)

Defferrard (approved):
LGTM! I've check some vacuum's rules that we're not covering yet. Feel free to fix them in this PR on in a futur iteration

ReidyT (approved):
Looks good!

PR #1527, opened 2024-10-17
feat: lint openapi during git actions

Defferrard:


spaenleh (commented):
Looks good ! Should we also update the tests file ?

Defferrard:
> Looks good ! Should we also update the tests file ?

If you mean the `test/app.ts`, I'd say yes, but I did not be able to make it works... I've probably didn't put enough effort in it because I think it is out of the scope of this PR

package.json
```diff
@@ -127,6 +129,7 @@
     "@commitlint/config-conventional": "19.2.2",
     "@faker-js/faker": "9.0.3",
     "@jest/globals": "29.7.0",
+    "@quobix/vacuum": "0.13.6",
```
ReidyT, line 132: I think you can put it in dev dependencies?

package.json
```diff
@@ -159,6 +162,7 @@
     "eslint-config-prettier": "8.10.0",
     "eslint-import-resolver-typescript": "3.6.3",
     "eslint-plugin-import": "2.29.1",
+    "fastify-cli": "7.0.1",
```
ReidyT, line 165: same here?

src/server.ts
ReidyT, line 1: Is this file still useful? Why don't we just use the new `fastify` file? Also, why do we have moved the start function in the other file?

ReidyT (commented):
The code looks good, but I still have a few questions before approving your PR ^^.

src/server.ts
Defferrard: So we can retrieve the export fastify instance without running the function `start()`

ReidyT: Mmmh okay thanks for your explanation! But in this case, could we move the code in the server.ts directly where this file is imported? I mean, we have an entire file just to start a method we imported from another file 😅

ReidyT: My bad, I thought server.ts was used in app, but its the entry point ^^

ReidyT (approved):

package.json
```diff
@@ -44,7 +44,9 @@
     "migration:fake": "yarn migration:run:fake && migration:revert:fake",
     "migration:run:fake": "npx typeorm-ts-node-commonjs  migration:run --fake",
     "migration:revert:fake": "npx typeorm-ts-node-commonjs migration:revert --fake",
-    "migration:check": "npx typeorm-ts-node-commonjs migration:generate src/migrations/migrations --check -d src/plugins/datasource.ts"
+    "migration:check": "npx typeorm-ts-node-commonjs migration:generate src/migrations/migrations --check -d src/plugins/datasource.ts",
+    "openapi:generate": "node dist/scripts/generateOpenAPI.js",
+    "openapi:lint": "vacuum lint ./openapi.json"
```
pyphilia, line 49: Shouldn't it be `.openapi.json`?  I have a personal preference for `openapi.json` though.

```suggestion
    "openapi:lint": "vacuum lint ./.openapi.json"
```

pyphilia (changes requested):
LGMT! But I think the vaccum script does not use the right file?

pyphilia (approved):

PR #1533, opened 2024-10-17
fix: password reset token should be single use

### LINKED ISSUE #1532: [BUG] Token used for password reset request can be used multiple times
It is possible to use a reset password link multiple times.

The key is incorrectly removed from redis.

spaenleh:
close #1532

### DIFF
--- a/src/services/auth/plugins/password/service.ts
+++ b/src/services/auth/plugins/password/service.ts
@@ -85,7 +85,7 @@ export class MemberPasswordService {
     if (!id) {
       return;
     }
-    await this.redis.del(uuid);
+    await this.redis.del(`${REDIS_PREFIX}${uuid}`);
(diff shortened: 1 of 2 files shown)


src/services/auth/plugins/password/service.ts
```diff
@@ -85,7 +85,7 @@ export class MemberPasswordService {
     if (!id) {
       return;
     }
-    await this.redis.del(uuid);
+    await this.redis.del(`${REDIS_PREFIX}${uuid}`);
```
pyphilia, line 88: Wouldn't it be clearer if you have a private fn to build the prefix? So you can reuse at get, post and delete.

pyphilia (approved):
Thanks for the quick fix! Maybe my comment is relevant before merging.

PR #1535, opened 2024-10-17
fix: do not encode redirection url for password

### LINKED ISSUE #1530: Login with password does not redirect correctly
```
{"statusCode":400,"code":"FST_ERR_VALIDATION","error":"Bad Request","message":"querystring/url must match format \"uri\""}
```
For url
```
https://api.dev.graasp.org/auth?t=<token>&url=https%253A%252F%252Fbuilder.dev.graasp.org
```

pyphilia:
fix #1530

### DIFF
--- a/src/services/auth/plugins/password/index.test.ts
+++ b/src/services/auth/plugins/password/index.test.ts
@@ -791,7 +791,7 @@ describe('Flow tests', () => {
       payload: { email: member.email, password: pwd.password, captcha: MOCK_CAPTCHA },
     });
 
-    expect(loginResponse.statusCode).toEqual(StatusCodes.OK);
+    expect(loginResponse.statusCode).toEqual(StatusCodes.SEE_OTHER);
 
     const response = await app.inject({
       method: HttpMethod.Get,
--- a/src/services/auth/plugins/password/index.test.ts
+++ b/src/services/auth/plugins/password/index.test.ts
@@ -761,7 +761,7 @@ describe('GET members current password status', () => {
   });
 });
 
-describe('Regression tests', () => {
+describe('Flow tests', () => {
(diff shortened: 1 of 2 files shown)


src/services/auth/plugins/password/index.test.ts
```diff
@@ -745,3 +745,43 @@ describe('GET members current password status', () => {
     expect(response.json()).toEqual({ hasPassword: true });
   });
 });
+
+describe('Regression tests', () => {
```
spaenleh, line 749: Should this be named differently ? 

`Flow tests` maybe instead of singling it out as a `regression test` ?

src/services/auth/plugins/password/index.test.ts
```diff
@@ -745,3 +745,43 @@ describe('GET members current password status', () => {
...
+    const member = await saveMemberAndPassword(m, pwd);
+
+    // login
+    const loginResponse = await app.inject({
+      method: HttpMethod.Post,
+      url: '/login-password',
+      payload: { email: member.email, password: pwd.password, captcha: MOCK_CAPTCHA },
+    });
+
+    const response = await app.inject({
```
spaenleh, line 779: Should we check the status code before sending this request ?

spaenleh (changes requested):
Thank you for the fix, let me know what you think ?

Defferrard (approved):
I agree with what @spaenleh said, otherwise LGTM!

ReidyT (approved):
LGTM 🌨️

spaenleh (approved):
🎩

spaenleh:
@pyphilia just looks like the expected status code is not `200` but `303` as this is a `redirect`

PR #1539, opened 2024-10-17
feat: save last authenticated for guest

### LINKED ISSUE #1538: Save `lastAuthenticated` for guest

pyphilia:
Refresh last authenticated at date for guest.

close #1538


src/services/itemLogin/errors.ts
```diff
@@ -202,3 +202,16 @@ export class CannotEnrollFrozenItemLoginSchema extends GraaspItemLoginError {
...
   }
 }
+
+export class CannotRefreshLastAuthenticatedAt extends GraaspItemLoginError {
+  constructor(data?: unknown) {
+    super(
+      {
+        code: 'GILERR016',
+        statusCode: StatusCodes.FORBIDDEN,
+        message: 'CANNOT_REFRESH_LAST_AUTHENTICATED_AT',
```
pyphilia, line 212: I will add the message in translations on approve.

src/services/itemLogin/test/index.test.ts
spaenleh, line 1: Is it correct that the error you added is not tested ? Does this mean that it is not situated on a possible code-path ?

src/services/itemLogin/test/index.test.ts
```diff
@@ -559,9 +567,17 @@ describe('Item Login Tests', () => {
...
             expectItemLogin(member, m);
-            expect(res.statusCode).toBe(StatusCodes.OK);
+
+            // last authenticated is updated
+            const guestInDb = await rawItemLoginRepository.findBy({
+              name: payload.username,
+              itemLoginSchema,
+            });
+            expect(guestInDb).toHaveLength(1);
+            expect(guestInDb[0].lastAuthenticatedAt).toBeDefined();
```
spaenleh, line 580: Should we also check that it is different after we logged in ? So we can ensure it is actually updated ?
If the test relies on the fact that it is undefined before, this assumption is not clear and I think it should be avoided. I  mean we should avoid relying on default values. Any values that we rely on in the test should be explicitly defined.

src/services/itemLogin/repositories/guest.ts
```diff
@@ -29,4 +31,21 @@ export class GuestRepository extends AbstractRepository<Guest> {
       ...guestData,
     });
   }
+
+  async get(id: UUID) {
```
spaenleh, line 35: Should the return of this function be explicitly annotated ?

spaenleh (changes requested):
Thank you for fixing this issue. 🐞 

I have left comments related to some architectural choices. Let me know what you think. 
Also please check that service functions are private if they do not need to be used for outside.

src/services/itemLogin/test/index.test.ts
```diff
@@ -532,36 +538,57 @@ describe('Item Login Tests', () => {
...
               payload,
             });
 
-            const member = res.json();
-            expectItemLogin(member, m);
             expect(res.statusCode).toBe(StatusCodes.OK);
+            const member = res.json();
+            expectItemLogin(member, guest);
+
+            // last authenticated is updated
```
spaenleh, line 582: Is comment up-to-date ?

spaenleh (commented):
Thank you for the refactor ! Maybe the difference is a bit too small ? should we make it bigger and more reality grounded ? Like say the member is first authenticated on Monday and logs back in a day later ? I fear that with such small differences we would make the test flacky.

Defferrard (approved):
LGTM!

src/services/itemLogin/test/index.test.ts
```diff
@@ -532,36 +538,57 @@ describe('Item Login Tests', () => {
...
           it('Successfully reuse item login with username', async () => {
             assertIsDefined(anotherItem);
             const payload = USERNAME_LOGIN;
             // pre-create pseudonymized data
-            const { guest: m } = await saveItemLoginSchema({
+            const { guest, itemLoginSchema } = await saveItemLoginSchema({
               item: anotherItem as unknown as DiscriminatedItem,
               memberName: payload.username,
+              // purposely set lastAuthenticatedAt date to yesterday
+              lastAuthenticatedAt: new Date(new Date().getDate() - 1),
```
spaenleh, line 569: This results in `Thu Jan 01 1970 01:00:00 GMT+0100 (Central European Standard Time)` which is not what you want I guess. 

Maybe using date-fns would be easier ?

src/services/itemLogin/test/index.test.ts
```diff
@@ -532,36 +538,57 @@ describe('Item Login Tests', () => {
...
+
+            const guestInDb = await rawItemLoginRepository.findBy({
+              name: payload.username,
+              itemLoginSchema,
+            });
+            expect(guestInDb).toHaveLength(1);
+            // last authenticated got updated
+            expect(
+              new Date(guestInDb[0].lastAuthenticatedAt) > new Date(guest.lastAuthenticatedAt),
+            ).toEqual(true);
```
spaenleh, line 591: Should we also expect the new lastlogin date to be within the last minute ?

src/services/itemLogin/test/index.test.ts
spaenleh:
```suggestion
              lastAuthenticatedAt: new Date(new Date() - (24 * 60 * 60 * 1000)),
```

spaenleh (changes requested):
Sorry for being picky but I think it does not do what it is intended to 😅

spaenleh (approved):

PR #1542, opened 2024-10-18
fix: update schema for current account

### LINKED ISSUE #1534: GET /current fails for guest user
Logging in as a guest will throw /current because of schema validation.

pyphilia:
Based on  https://github.com/graasp/graasp/pull/1539

fix #1534

### DIFF
--- a/src/services/member/schemas.ts
+++ b/src/services/member/schemas.ts
@@ -51,39 +51,6 @@ export const nullableMemberSchemaRef = registerSchemaAsRef(
   customType.Nullable(memberSchema),
 );
 
-const currentMemberSchema = Type.Object(
-  {
-    // Object definition
-    id: customType.UUID(),
-    name: customType.Username(),
-    email: Type.Optional(Type.String({ format: 'email' })),
-    type: Type.Enum(AccountType),
-    createdAt: customType.DateTime(),
-    updatedAt: customType.DateTime(),
-    lastAuthenticatedAt: customType.DateTime(),
-    isValidated: Type.Boolean(),
-    userAgreementsDate: customType.DateTime(),
-    enableSaveActions: Type.Boolean(),
-    extra: Type.Object({}, { additionalProperties: true }),
-  },
-  {
-    // Schema options
-    additionalProperties: false,
-  },
-);
-
-export const currentMemberSchemaRef = registerSchemaAsRef(
-  'currentMember',
-  'Current Member',
-  currentMemberSchema,
-);
-
-export const nullableCurrentMemberSchemaRef = registerSchemaAsRef(
-  'nullableCurrentMember',
-  'Nullable Current Member',
-  customType.Nullable(currentMemberSchema),
-);
-
 export const updateMemberRequiredOneSchemaRef = registerSchemaAsRef(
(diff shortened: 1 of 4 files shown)


test/setup.ts
spaenleh, line 1: Should this file be in the `/test` folder or inside `/src/services/member/test` ?

src/services/member/test/index.test.ts
```diff
@@ -47,11 +57,28 @@ describe('Member routes tests', () => {
...
+      expect(m.name).toEqual(member.name);
+      expect(m.email).toEqual(member.email);
+      expect(m.id).toEqual(member.id);
       expect(m.password).toBeUndefined();
+    });
+
+    it('Returns successfully if signed in as guest', async () => {
+      const { guest } = await setupGuest(app);
+
+      mockAuthenticate(guest);
```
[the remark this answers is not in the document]
spaenleh: For this we should probably rely on making tests with support for a session (if we want to use the cookie based authentication). I think `app.inject` is better suited to test either individual endpoints or API that use JWT authentication that can be passed in a header.

src/services/member/schemas.ts
```diff
@@ -240,7 +281,7 @@ export const updateOne = {
   params: entityIdSchemaRef,
   body: updateMemberRequiredOneSchemaRef,
   response: {
-    [StatusCodes.OK]: currentMemberSchemaRef,
+    [StatusCodes.OK]: currentAccountSchemaRef,
```
spaenleh, line 284: I am thinking that the frontend usually does not make use of the return values from POST and PATCH calls (unless we use `mutateAsync` so maybe we could spare the returns and rely on the stale fetching from the hooks ? since typeorm is also not very useful when we want to get the value we have to perform a get after the update. Maybe something to think about for later.

spaenleh (changes requested):
Changes look good.

There is just something I don't quite get. For anyone on graasp `account` is supposed to be other users right ? So it should only contain `id, name`.

We only expose more of these properties when we are giving the **current** member. In these cases it can be an `individual` or a `guest`.

I think we should have these concepts clear. I think the issue is that in terms of entities the account is the most complete one. But in terms of what the API and types for clients are. Account is really the most simple `id,name` one. 
I think the schema names should represent this.

test/setup.ts
pyphilia: I envision it as a reusable setup, so that's why it's in test. But I can move it to member. What do you think?

pyphilia:
From my understanding:
- `Account` can either be a `Guest` or a `Member`. It's usually what we get for an `account` property (id, name)
- `CompleteX` is a type in `graasp/sdk`, sometimes used in the backend, but more for its type. But it is was what was used for typing `/current`. So I instead called it `CurrentX` in my PR.
- `AugmentedX` is what is used for memberships, since we needed more data to display the table (eg: email).

`Account` is not the most complete type. I think even when using typeorm, you only receive the common properties, since it's based on the entity.

Let me know if you need more explanations.

spaenleh:
@pyphilia Your explanation makes sens. I guess I was thrown off by the use of both `account` and `currentAccount`.

spaenleh (approved):

PR #1546, opened 2024-10-18
fix: set html as optional for link metadata

### LINKED ISSUE #1545: Metadata throws because of schema
Metadata link returns
```
"\"html\" is required!"
```
for a valid link such as `https://graasp.org`

pyphilia:
fix #1545

### DIFF
--- a/src/services/item/plugins/embeddedLink/test/controller.test.ts
+++ b/src/services/item/plugins/embeddedLink/test/controller.test.ts
@@ -1,3 +1,4 @@
+import { faker } from '@faker-js/faker';
 import { StatusCodes } from 'http-status-codes';
 import fetch, { Response } from 'node-fetch';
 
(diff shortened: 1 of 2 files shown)


src/services/item/plugins/embeddedLink/test/controller.test.ts
```diff
@@ -75,5 +76,37 @@ describe('Tests Embedded Link Controller', () => {
 
       expect(response.statusCode).toBe(StatusCodes.OK);
     });
+
+    it('Returns 200 Ok when a valid URL without html', async () => {
+      (fetch as jest.MockedFunction<typeof fetch>).mockImplementation(async () => {
```
spaenleh, line 81: You did not try with nock (😉) ?

spaenleh (approved):

PR #1552, opened 2024-10-21
fix: fix packed tag

### LINKED ISSUE #1547: Fix flacky test
```
 ● Item routes tests › GET /items/:id/descendants › Public › Returns successfully

    expect(received).toEqual(expected) // deep equality

    Expected: "ba32bc80-7199-4161-ab2f-2fa17fa55506"
    Received: "4ddbfcec-c5fa-49d8-9175-b2de27da92a4"

      304 |   if (pTag) {
      305 |     expect(newItem!.public!.type).toEqual(pTag.type);
    > 306 |     expect(newItem!.public!.id).toEqual(pTag.id);
          |                                 ^
      307 |     expect(newItem!.public!.item!.id).toEqual(pTag.item.id);
      308 |   }
      309 |   const hTag = tags?.find((t) => t.type === ItemTagType.Hidden);

      at toEqual (src/services/item/test/fixtures/items.ts:306:33)
      at src/services/item/test/index.read.test.ts:1579:27
          at Array.forEach (<anonymous>)
      at src/services/item/test/index.read.test.ts:1578:21
      at fulfilled (src/services/item/test/index.read.test.ts:28:58)
```

pyphilia:
Packed returns tags such as `public` and `hidden`. But sometimes it does not return the correct one: we want to return the highest inherited tag. It happens because it depends on the array order (usually fetched from db). I've also added sorting for `createPackedItems`.

That will hopefully fix the flacky test. 

I don't think it's the best fix, I would prefer to order from the db, or actually get the correct tag right away from the db. 
With the packed we actually removed the use of `GET /tags` and so we can remove `ResultOf` from getting tags.

I would apply this fix first, then I will deprecate getting tags, and finally remove `ResultOf`. 

fix #1547

### DIFF
--- a/src/services/account/repository.test.ts
+++ b/src/services/account/repository.test.ts
@@ -8,7 +8,7 @@ import { AccountRepository } from './repository';
 
 const accountRepository = new AccountRepository();
 
-describe('MemberRepository', () => {
+describe('AccountRepository', () => {
   let app: FastifyInstance;
 
   beforeEach(async () => {
--- a/src/services/item/ItemWrapper.test.ts
+++ b/src/services/item/ItemWrapper.test.ts
@@ -1,0 +1,44 @@
+import { FastifyInstance } from 'fastify';
+
+import { ItemTagType } from '@graasp/sdk';
+
+import build, { clearDatabase } from '../../../test/app';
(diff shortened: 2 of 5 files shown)


spaenleh (approved):
LGTM !

PR #1557, opened 2024-10-21
refactor: remove GET tags, simplify tags return value

pyphilia:
Remove GET /tags endpoints, as we now get tags only with the packed item.


spaenleh (approved):
Yeah !

PR #1566, opened 2024-10-25
docs: update item tag openapi doc

pyphilia:
ref #1528


src/services/item/ItemWrapper.ts
```diff
@@ -67,11 +67,18 @@ export class ItemWrapper {
     for (const i of items) {
       const { permission = null } = memberships.data[i.id] ?? {};
       const thumbnails = itemsThumbnails?.[i.id];
+
+      // sort tags to retrieve most relevant (highest) tag first
+      const thisTags = tags?.data?.[i.id];
```
ReidyT, line 72: I'm not a big fan of `thisTags` ^^, what do you think about my suggestion?
```suggestion
      const itemTags = tags?.data?.[i.id];
```

ReidyT (approved):
LGTM, I just add a suggestion, feel free to ignore it if you disagree ^^

PR #1567, opened 2024-10-25
feat: add migration from item category to item lang

### LINKED ISSUE #800: Add item's lang column in meilisearch filterable attributes
- back: inherit lang from parent item
- add lang in meilisearch
	- update child on change
- migration : remove lang from categories -> move category tags to item lang

pyphilia:
close  #800

### DIFF
--- a/src/migrations/1729857146775-item-category-to-item.ts
+++ b/src/migrations/1729857146775-item-category-to-item.ts
@@ -4,7 +4,7 @@ export class Migrations1729857146775 implements MigrationInterface {
   name = 'language-category-to-item-1729857146775';
 
   public async up(queryRunner: QueryRunner): Promise<void> {
-    // we ignore english category
+    // we ignore english category because default item lang is already english
     for (const [lang, key] of [
       ['french', 'fr'],
       ['german', 'de'],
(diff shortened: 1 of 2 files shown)


src/migrations/1729857146775-item-category-to-item.ts
```diff
@@ -0,0 +1,31 @@
...
+  name = 'language-category-to-item-1729857146775';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    // we ignore english category because default item lang is already english
+    for (const [lang, key] of [
+      ['french', 'fr'],
+      ['german', 'de'],
+      ['arabic', 'ar'],
+      ['italian', 'it'],
+    ]) {
```
spaenleh, line 13: The order of the loop will matter right ? If an item has both french and italian for example, it will receive italian as the final language ?

src/migrations/1729857146775-item-category-to-item.ts
```diff
@@ -0,0 +1,31 @@
...
+        )`,
+      );
+
+      // remove language category from db
+      // cascade delete relatede item_category
+      await queryRunner.query(`DELETE FROM category where type='language'`);
+    }
+  }
+
+  public async down(): Promise<void> {}
```
spaenleh, line 30: I guess it makes sens that we can't really have a down migration for this step since we are potentially loosing data. I am wondering however if we should still try to have a migration that at least adds back the languages in the categories ?

ReidyT (commented):
I agree with Basile, maybe we can discuss later?

spaenleh (approved):

PR #1571, opened 2024-10-25
docs: add documentation for item favorite, small refactor

pyphilia:
ref #1528

### DIFF
--- a/src/services/item/plugins/itemFavorite/schemas.ts
+++ b/src/services/item/plugins/itemFavorite/schemas.ts
@@ -45,8 +45,8 @@ const packedFavoriteSchemaRef = registerSchemaAsRef(
 export const getOwnFavorite = {
   operationId: 'getOwnFavorite',
   tags: ['favorite'],
-  summary: 'Get all own bookmarked instances',
-  description: 'Get all own bookmarked instances',
+  summary: 'Get all bookmarked instances of the current member',
+  description: 'Get all bookmarked instances of the current member',
 
   response: {
     [StatusCodes.OK]: Type.Array(packedFavoriteSchemaRef),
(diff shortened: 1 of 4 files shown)


src/services/item/plugins/itemFavorite/schemas.ts
```diff
@@ -35,38 +36,50 @@ export const packedFavoriteSchemaRef = registerSchemaAsRef(
...
-      },
-      {
-        additionalProperties: false,
-      },
-    ),
-  ),
+export const getOwnFavorite = {
+  operationId: 'getOwnFavorite',
+  tags: ['favorite'],
+  summary: 'Get all own bookmarked instances',
```
ReidyT, line 48: What about 
```suggestion
  summary: 'Get all bookmarked instances of the current member',
```

src/services/item/plugins/itemFavorite/schemas.ts
```diff
@@ -35,38 +36,50 @@ export const packedFavoriteSchemaRef = registerSchemaAsRef(
...
+  description: 'Get all own bookmarked instances',
```
ReidyT, line 49:
```suggestion
  description: 'Get all bookmarked instances of the current member',
```

ReidyT (approved):
LGTM! I just have a global question: is the description mandatory? Because I have the feelings that in some case, it only add redundancy. Maybe we should only add the description when we add a real new information?

Here is from the example of Swagger 3, we see a real interest of the description, and I think they don't use it for each route.
image

pyphilia:
Thanks for the review!
Openapi specs recommends to always have a summary and description for all endpoints!

ReidyT:
> Thanks for the review! Openapi specs recommends to always have a summary and description for all endpoints!

Okay, thanks for your answer!

PR #1572, opened 2024-10-28
refactor: remove fluent schema

pyphilia:
Remove fluent schema since tricks using it are not used anymore.
Use typebox definition for FileItemMetadata (get storage list).


ReidyT (approved):
🎃

PR #1582, opened 2024-10-30
docs: add docs for recycled item endpoints

pyphilia:
ref #1528

### DIFF
--- a/src/services/item/plugins/recycled/schemas.ts
+++ b/src/services/item/plugins/recycled/schemas.ts
@@ -1,55 +1,38 @@
-import { TRef, Type } from '@sinclair/typebox';
+import { Type } from '@sinclair/typebox';
 import { StatusCodes } from 'http-status-codes';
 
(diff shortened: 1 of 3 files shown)


src/services/item/plugins/recycled/schemas.ts
```diff
@@ -1,55 +1,38 @@
...
-      // Object definition
       id: customType.UUID(),
       item: itemSchemaRef,
       creator: nullableMemberSchemaRef,
       createdAt: customType.DateTime(),
     },
     {
-      // Schema options
-      additionalProperties: false,
+      description: 'Instance representing a deleted item and its deleted descendants',
```
ReidyT, line 25:
```suggestion
      description: "Instance representing a deleted item and its deleted descendants. A recycled item is not permanently deleted, it's still possible to recover it.",
```

ReidyT (approved):
LGTM!

PR #1584, opened 2024-10-30
docs: add docs for app data file endpoints

pyphilia:
ref #1528

### DIFF
--- a/src/services/item/plugins/app/appData/plugins/file/index.ts
+++ b/src/services/item/plugins/app/appData/plugins/file/index.ts
@@ -1,10 +1,9 @@
 import { fastifyMultipart } from '@fastify/multipart';
-import { FastifyPluginAsync } from 'fastify';
+import { FastifyPluginAsyncTypebox } from '@fastify/type-provider-typebox';
 
-import { HttpMethod, UUID } from '@graasp/sdk';
+import { HttpMethod } from '@graasp/sdk';
 
 import { resolveDependency } from '../../../../../../../di/utils';
-import { IdParam } from '../../../../../../../types';
 import { asDefined } from '../../../../../../../utils/assertions';
 import { Repositories, buildRepositories } from '../../../../../../../utils/repositories';
 import { guestAuthenticateAppsJWT } from '../../../../../../auth/plugins/passport';
@@ -15,6 +14,7 @@ import {
   UploadFileUnexpectedError,
 } from '../../../../../../file/utils/errors';
 import { Actor, Member } from '../../../../../../member/entities/member';
+import { addMemberInAppData } from '../../../legacy';
 import { AppData } from '../../appData';
 import { PreventUpdateAppDataFile } from '../../errors';
 import { AppDataService } from '../../service';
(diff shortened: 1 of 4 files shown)


src/services/item/plugins/app/appData/plugins/file/schema.ts
```diff
@@ -1,29 +1,42 @@
...
 
-const download = {
+export const download = {
+  operationId: 'downloadAppDataFile',
+  tags: ['app', 'app-data', 'file'],
+  summary: 'Download app data file',
+  description: 'Download app data file.',
+
   params: customType.StrictObject({
     id: customType.UUID(),
```
spaenleh, line 30: Same comment here, can we specify which `id` this is ?

src/services/item/plugins/app/appData/plugins/file/schema.ts
```diff
@@ -1,29 +1,42 @@
...
 
-const upload = {
-  querystring: {
-    type: 'object',
-    properties: {
-      id: customType.UUID(),
-    },
-    additionalProperties: false,
+  querystring: customType.StrictObject({
+    id: customType.UUID(),
```
spaenleh, line 15: Also here.

src/services/item/plugins/app/appData/plugins/file/index.ts
```diff
@@ -77,7 +77,7 @@ const basePlugin: FastifyPluginAsync<GraaspPluginFileOptions> = async (fastify,
   };
   appDataService.hooks.setPreHook('patch', patchPreHook);
 
-  fastify.route<{ Querystring: IdParam; Body: unknown }>({
+  fastify.route({
     method: HttpMethod.Post,
```
spaenleh, line 81: Same question here about the use of `fastify.route` instead of the shorthand.

spaenleh (changes requested):
Nice ! I think we should change some things before merging. 📜

src/services/item/plugins/app/appSetting/plugins/file/index.ts
```diff
@@ -88,7 +88,7 @@ const basePlugin: FastifyPluginAsync<GraaspPluginFileOptions> = async (fastify,
   };
   appSettingService.hooks.setPreHook('patch', patchPreHook);
 
-  fastify.route<{ Body: unknown }>({
+  fastify.route({
     method: HttpMethod.Post,
```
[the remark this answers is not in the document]
pyphilia: Not sure, but it was not the goal of the PR so I didn't change it.

spaenleh (approved):

src/services/item/plugins/app/appData/plugins/file/schema.ts
```diff
@@ -1,29 +1,41 @@
...
-    },
-    additionalProperties: false,
+  querystring: customType.StrictObject({
+    replyUrl: Type.Boolean({
+      default: false,
+    }),
+  }),
+
+  response: {
+    [StatusCodes.OK]: Type.String({ format: 'uri' }),
```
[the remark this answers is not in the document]
pyphilia: I think it does, it is for regular files or thumbnails that it doesn't.

https://github.com/graasp/graasp/blob/908a278f1116f8b34bf34ba1a9463f5ffccc3dcf/src/services/item/plugins/app/appData/plugins/file/service.ts#L96

PR #1587, opened 2024-10-30
refactor: remove item id schema ref

pyphilia:


src/services/itemMembership/schemas.ts
```diff
@@ -69,7 +71,9 @@ export const create = {
 
 // schema for creating many item memberships
 export const createMany = {
-  params: itemIdSchemaRef,
+  params: customType.StrictObject({
```
spaenleh, line 74: Why is it in the params here and in the query in the single ?

PR #1592, opened 2024-10-30
docs: add docs to item login endpoints

pyphilia:
ref #1528

### DIFF
--- a/src/services/itemLogin/schemas.ts
+++ b/src/services/itemLogin/schemas.ts
@@ -14,11 +14,11 @@ const itemLoginSchemaSchema = customType.StrictObject(
   {
     id: customType.UUID(),
     type: customType.EnumString(Object.values(ItemLoginSchemaType), {
-      description: 'Defined which credentials are necessary to login.',
+      description: 'Defines which credentials are necessary to login.',
     }),
     status: customType.EnumString(Object.values(ItemLoginSchemaStatus), {
       description:
-        'Whether the item login is enabled, freezed or disabled. An item login is never deleted, but disabled instead to prevent attached guests to be deleted.',
+        'Item login status, which can be enabled, frozen, or disabled. Item login cannot be deleted, an item login can be disabled instead to prevent deleting associated guest accounts.',
     }),
     item: Type.Optional(itemSchemaRef),
     createdAt: customType.DateTime(),
@@ -26,7 +26,7 @@ const itemLoginSchemaSchema = customType.StrictObject(
   },
   {
     description:
-      'Instance allowing to login without a member on related item and its descendants. The required credentials are defined given the status.',
+      'Instance allowing to login without a member on related item and its descendants. The required credentials are defined given the type.',
   },
 );
 
@@ -36,20 +36,18 @@ export const itemLoginSchemaSchemaRef = registerSchemaAsRef(
   itemLoginSchemaSchema,
 );
 
-export const loginOrRegister = {
-  operationId: 'itemLogin',
+export const loginOrRegisterAsGuest = {
+  operationId: 'loginOrRegisterAsGuest',
   tags: ['item-login'],
-  summary: 'Login or Register in item',
-  description: `Login in item with necessary credentials depending on item login's status. If the username does not exist, a guest account is created and the user is logged in.`,
+  summary: 'Login or Register in item as guest',
+  description: `Log in to an item with necessary credentials depending on item login's type. If the username does not exist, a guest account is created and is given access.`,
 
   params: customType.StrictObject({
     id: customType.UUID(),
   }),
   body: customType.StrictObject({
     username: customType.Username(),
-    password: Type.Optional(
-      Type.String({ minLength: 3, maxLength: 50, pattern: '^\\S+( \\S+)*$' }),
-    ),
+    password: Type.Optional(Type.String({ minLength: 3, maxLength: 50 })),
   }),
   response: {
     [StatusCodes.OK]: accountSchemaRef,
@@ -76,7 +74,7 @@ export const getLoginSchemaType = {
   },
 } as const satisfies FastifySchema;
 
-export const getLoginSchema = {
+export const getItemLoginSchema = {
   operationId: 'getItemLoginSchema',
   tags: ['item-login'],
   summary: 'Get item login data',
--- a/src/services/itemLogin/schemas.ts
+++ b/src/services/itemLogin/schemas.ts
@@ -1,4 +1,5 @@
 import { Type } from '@sinclair/typebox';
+import { StatusCodes } from 'http-status-codes';
 
 import { FastifySchema } from 'fastify';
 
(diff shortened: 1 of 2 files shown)


src/services/itemLogin/schemas.ts
```diff
@@ -9,89 +10,100 @@ import { errorSchemaRef } from '../../schemas/global';
...
   {
     id: customType.UUID(),
-    type: customType.EnumString(Object.values(ItemLoginSchemaType)),
-    status: customType.EnumString(Object.values(ItemLoginSchemaStatus)),
+    type: customType.EnumString(Object.values(ItemLoginSchemaType), {
+      description: 'Defined which credentials are necessary to login.',
+    }),
+    status: customType.EnumString(Object.values(ItemLoginSchemaStatus), {
+      description:
+        'Whether the item login is enabled, freezed or disabled. An item login is never deleted, but disabled instead to prevent attached guests to be deleted.',
```
ReidyT, line 21:
```suggestion
        'Item login status, which can be enabled, frozen, or disabled. While never deleted, an item login can be disabled instead to prevent deleting associated guest accounts.',
```

src/services/itemLogin/schemas.ts
```diff
@@ -9,89 +10,100 @@ import { errorSchemaRef } from '../../schemas/global';
...
+    }),
     item: Type.Optional(itemSchemaRef),
     createdAt: customType.DateTime(),
     updatedAt: customType.DateTime(),
   },
-  { additionalProperties: false },
+  {
+    description:
+      'Instance allowing to login without a member on related item and its descendants. The required credentials are defined given the status.',
```
ReidyT, line 29: Not sure, but: 
```suggestion
      'Instance allowing to login without a member on related item and its descendants. The required credentials are defined given the type.',
```

src/services/itemLogin/schemas.ts
```diff
@@ -9,89 +10,100 @@ import { errorSchemaRef } from '../../schemas/global';
...
   'Item Login Schema',
   itemLoginSchemaSchema,
 );
 
-export const login = {
+export const loginOrRegister = {
+  operationId: 'itemLogin',
+  tags: ['item-login'],
+  summary: 'Login or Register in item',
+  description: `Login in item with necessary credentials depending on item login's status. If the username does not exist, a guest account is created and the user is logged in.`,
```
ReidyT, line 43: still not sure ^^
```suggestion
  description: `Login in item with necessary credentials depending on item login's type. If the username does not exist, a guest account is created and the user is logged in.`,
```

src/services/itemLogin/schemas.ts
```diff
@@ -9,89 +10,100 @@ import { errorSchemaRef } from '../../schemas/global';
...
+
   params: customType.StrictObject({
     id: customType.UUID(),
   }),
-  body: credentialsSchemaRef,
+  body: customType.StrictObject({
+    username: customType.Username(),
+    password: Type.Optional(
+      Type.String({ minLength: 3, maxLength: 50, pattern: '^\\S+( \\S+)*$' }),
```
ReidyT, line 51: Should we extract this pattern to name it and simplify the readability? 
```suggestion
      Type.String({ minLength: 3, maxLength: 50, pattern: 'SingleSpacedWords' }),
```

ReidyT (changes requested):
LGTM, I just let you some suggestions. 🎃

src/services/itemLogin/schemas.ts
```diff
@@ -9,89 +10,100 @@ import { errorSchemaRef } from '../../schemas/global';
...
-      password: Type.Optional(
-        Type.String({ minLength: 3, maxLength: 50, pattern: '^\\S+( \\S+)*$' }),
-      ),
-    },
-    { additionalProperties: false },
-  ),
-);
-
-const itemLoginSchemaSchema = Type.Object(
+const itemLoginSchemaSchema = customType.StrictObject(
```
spaenleh, line 13: Haha the name is a bit weird, but 🤷

src/services/itemLogin/schemas.ts
```diff
@@ -9,89 +10,100 @@ import { errorSchemaRef } from '../../schemas/global';
...
   {
     id: customType.UUID(),
-    type: customType.EnumString(Object.values(ItemLoginSchemaType)),
-    status: customType.EnumString(Object.values(ItemLoginSchemaStatus)),
+    type: customType.EnumString(Object.values(ItemLoginSchemaType), {
+      description: 'Defined which credentials are necessary to login.',
```
spaenleh, line 17:
```suggestion
      description: 'Defines which credentials are necessary to login.',
```

src/services/itemLogin/schemas.ts
```diff
@@ -9,89 +10,100 @@ import { errorSchemaRef } from '../../schemas/global';
...
+
 export const itemLoginSchemaSchemaRef = registerSchemaAsRef(
   'itemLoginSchema',
   'Item Login Schema',
   itemLoginSchemaSchema,
 );
 
-export const login = {
+export const loginOrRegister = {
+  operationId: 'itemLogin',
```
spaenleh, line 40:
```suggestion
  operationId: 'loginOrRegisterAsGuest',
```

src/services/itemLogin/schemas.ts
```diff
@@ -9,89 +10,100 @@ import { errorSchemaRef } from '../../schemas/global';
...
+  tags: ['item-login'],
+  summary: 'Login or Register in item',
```
spaenleh, line 42:
```suggestion
  summary: 'Login or Register in item as guest',
```

src/services/itemLogin/schemas.ts
spaenleh, line 43: Here you use `item login` but isn't it an `itemLoginSchema` ? If the `schema` part is not necessary, then we should remove it. Maybe we should discuss to better define the naming of these elements.

spaenleh, line 43:
```suggestion
  description: `Log in to an item with necessary credentials depending on item login's type. If the username does not exist, a guest account is created and the user is given access.`,
```

src/services/itemLogin/schemas.ts
spaenleh: Is this regex really useful ? what guaranties does it give us ? can't the password simply be a string ?

src/services/itemLogin/schemas.ts
```diff
@@ -9,89 +10,100 @@ import { errorSchemaRef } from '../../schemas/global';
...
+  tags: ['item-login'],
+  summary: 'Get type of item login',
+  description: `Get type of item login. Return null if the item does not allow item login.`,
+
   params: customType.StrictObject({
     id: customType.UUID(),
   }),
   response: {
-    '2xx': customType.Nullable(customType.EnumString(Object.values(ItemLoginSchemaType))),
+    [StatusCodes.OK]: customType.Nullable(
```
spaenleh, line 71: Should we use a different status code for when the result is actually nothing ?

src/services/itemLogin/schemas.ts
```diff
@@ -9,89 +10,100 @@ import { errorSchemaRef } from '../../schemas/global';
...
+      customType.EnumString(Object.values(ItemLoginSchemaType)),
+    ),
     '4xx': errorSchemaRef,
     '5xx': errorSchemaRef,
   },
 } as const satisfies FastifySchema;
 
 export const getLoginSchema = {
```
spaenleh, line 79:
```suggestion
export const getItemLogin = {
```

spaenleh (approved):
maybe there is way to improve the naming:
loginSchema -> itemLogin
itemLogin (the actual credentials) -> guest account ?

src/services/itemLogin/schemas.ts
pyphilia: It was called `schema` to differentiate with `item login` that was the link between member and item login schema. But now we don't have this link anymore (the link is the guests themselves) so we can use `item login` for the whole namenclature.
But let's rename in another PR, so I will keep the `itemLoginSchema` nomenclature for now.

ReidyT (approved):
Thanks for your changes, LGTM 🤖

PR #1593, opened 2024-10-30
docs: add docs for zip import and export endpoints

pyphilia:
ref #1528

### DIFF
--- a/src/services/item/plugins/importExport/schema.ts
+++ b/src/services/item/plugins/importExport/schema.ts
@@ -10,7 +10,8 @@ export const zipImport = {
   operationId: 'importZip',
   tags: ['item', 'import'],
   summary: 'Import ZIP content',
-  description: 'Import content of a ZIP, creating corresponding structure and items.',
+  description:
+    'Import and extract the content of a ZIP, creating the corresponding structure and items.',
 
   querystring: customType.StrictObject({
     parentId: Type.Optional(
--- a/src/services/item/plugins/importExport/schema.ts
+++ b/src/services/item/plugins/importExport/schema.ts
@@ -1,24 +1,41 @@
 import { Type } from '@sinclair/typebox';
+import { StatusCodes } from 'http-status-codes';
 
 import { FastifySchema } from 'fastify';
 
 import { customType } from '../../../../plugins/typebox';
+import { errorSchemaRef } from '../../../../schemas/global';
 
 export const zipImport = {
-  querystring: Type.Partial(
-    Type.Object(
-      {
-        parentId: customType.UUID(),
-      },
-      {
-        additionalProperties: false,
-      },
+  operationId: 'importZip',
+  tags: ['item', 'import'],
+  summary: 'Import ZIP content',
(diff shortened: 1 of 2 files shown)


src/services/item/plugins/importExport/schema.ts
```diff
@@ -1,24 +1,41 @@
...
-      {
-        parentId: customType.UUID(),
-      },
-      {
-        additionalProperties: false,
-      },
+  operationId: 'importZip',
+  tags: ['item', 'import'],
+  summary: 'Import ZIP content',
+  description: 'Import content of a ZIP, creating corresponding structure and items.',
```
ReidyT, line 13:
```suggestion
  description: 'Import and extract the content of a ZIP, creating the corresponding structure and items.',
```

ReidyT (approved):
LGTM, just have a look at my suggestion 🫠

src/services/item/plugins/importExport/schema.ts
```diff
@@ -1,24 +1,41 @@
...
+
+  querystring: customType.StrictObject({
+    parentId: Type.Optional(
```
spaenleh, line 16: Interesting that the name of this parameter is not consistent with the file upload one.

src/services/item/plugins/importExport/schema.ts
```diff
@@ -1,24 +1,41 @@
...
-      },
-      {
-        additionalProperties: false,
-      },
+  operationId: 'importZip',
+  tags: ['item', 'import'],
+  summary: 'Import ZIP content',
+  description: 'Import content of a ZIP, creating corresponding structure and items.',
+
+  querystring: customType.StrictObject({
```
spaenleh, line 15: There is no handling of "after this element" here ?

src/services/item/plugins/importExport/schema.ts
```diff
@@ -1,24 +1,41 @@
...
+  summary: 'Export content',
+  description:
+    'Export content. Return raw file for single item, or a ZIP with structure and items for a folder.',
+
   params: customType.StrictObject({
     itemId: customType.UUID(),
   }),
+  response: {
+    // return a stream
+    [StatusCodes.OK]: { content: Type.String() },
```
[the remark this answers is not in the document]
spaenleh: Maybe this: https://swagger.io/docs/specification/v3_0/describing-responses/#response-that-returns-a-file

spaenleh (approved):
Looks good, I have left a few comments here and there

src/services/item/plugins/importExport/schema.ts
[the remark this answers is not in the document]
pyphilia: I tried already, also it looks like a pb is that it seems we need to define a format? And also `byte` or `binary` are not handled.. We can have an issue for that.

src/services/item/plugins/importExport/schema.ts
pyphilia: No, I think it was a bit of work to handle it. Not everything handles the `previousItemId`.

PR #1594, opened 2024-11-01
Optimize slow GET own recycled items

### LINKED ISSUE #1551: Optimize GET /recycled recycle bin
Query is very slow.

Add index on item_path similarly to item memberships?

Probably we also want to move to item id instead of item path.

pyphilia:
I played a long time with indices and queries. My main problem was reordering and limit, for some reasons it was slowing down the original queries (and all similar queries I wrote). But what I learned in the end is "start with the smallest table", which seems obvious, tho.

So instead of starting with `recycled_item_data`, I started with item membership which could be filtered "right away" and become a smaller table for later joins. This helped a lot. I supported the query with an index `ItemMembership(accountId, permission)`. For the purpose of this GET, I return items directly since we don't really use any data from recycled item data. I will change the frontend for this.

I found a way to use another repository from a given repository. Maybe it's a bit mixed up, but I prefer having the purpose in the same repository than mixing with itemmembership juste because it comes from that repo. Let me know what you think.

close #1551

### DIFF
--- a/src/services/item/plugins/recycled/repository.ts
+++ b/src/services/item/plugins/recycled/repository.ts
@@ -41,17 +41,16 @@ export class RecycledItemDataRepository extends MutableRepository<RecycledItemDa
     const query = this.repository
       .createQueryBuilder('recycledItem')
       .withDeleted()
-      .leftJoinAndSelect('recycledItem.creator', 'creator')
-      .leftJoinAndSelect('recycledItem.item', 'item')
-      .leftJoinAndSelect('item.creator', 'itemMember')
       .innerJoin(
         'item_membership',
         'im',
-        `im.item_path @> item.path 
-        AND im.permission = :permission 
-        AND im.account_id = :accountId`,
+        `im.account_id = :accountId
+        AND im.permission = :permission
+        AND im.item_path @> recycledItem.item_path`,
         { permission: PermissionLevel.Admin, accountId: account.id },
       )
+      .leftJoinAndSelect('recycledItem.item', 'item')
+      .leftJoinAndSelect('item.creator', 'itemMember')
       .orderBy('recycledItem.created_at', 'DESC');
 
     const [data, totalCount] = await query.offset(skip).limit(limit).getManyAndCount();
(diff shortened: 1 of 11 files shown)


src/services/item/plugins/recycled/test/index.test.ts
```diff
@@ -158,14 +158,46 @@ describe('Recycle Bin Tests', () => {
...
+          const res = await app.inject({
+            method: HttpMethod.Get,
+            url: '/items/recycled',
+          });
+
+          const response = res.json();
+          expect(res.statusCode).toBe(StatusCodes.OK);
+
+          // should not return child item for actor
+          // does not return parent because actor has read permissio
```
ReidyT, line 195:
```suggestion
          // does not return parent because actor has read permission
```

ReidyT (approved):
Very nice optimisation! It looks good to me. As already discussed, we could create an issue to check whether we can one day delete the RecycledItem entity if we don't use it (it's not a priority, but it could simplify the code, if there is no performance issues).

For me, it's fine what you've done with the use of the ItemMembershipRepository in the RecycledRepository.

Great job 🎇

src/services/item/plugins/recycled/repository.ts
```diff
@@ -33,29 +34,41 @@ export class RecycledItemDataRepository extends MutableRepository<RecycledItemDa
...
+        permission: PermissionLevel.Admin,
+      })
+      // show most recently deleted items first
+      .orderBy('item.deleted_at', 'DESC')
+      .offset(skip)
+      .limit(limit);
 
-    const [data, totalCount] = await query.offset(skip).limit(limit).getManyAndCount();
-    return { data, totalCount, pagination };
+    const [data, totalCount] = await query.getManyAndCount();
```
spaenleh, line 70: I think we will have issues with getting the count with this query too. It does not really make sense to optimize a query with pagination if we perform a query that materializes all rows just to get their count. 

If we get the count just for the frontend to show the last page, I think it is not worth it. I guess having an infinite query would be enough in the frontend. What do you think ?

spaenleh (changes requested):
Thank you for your work on this query. Your justification makes sense and this is in my experience what works best. 

I have left a comment about the use of the `getmanyAndCount`. I think we should not count on this query. It can have a huge performance penalty and is not worth it. 

Let me know. We can validate this and refactor the removal of the count in a different PR.

spaenleh (approved):
As discussed, I approve now, so these improvements can be tested IRL. However I think we still need to find a way to not do the count query on each page, if at all.

PR #1598, opened 2024-11-01
docs: remove useless ref, create ref

pyphilia:
- Remove useless refs.
- Also added necessary ones for better generation.
- Added handling of component's def names


src/plugins/swagger.ts
```diff
@@ -14,6 +14,11 @@ export default async function (instance: FastifyInstance): Promise<void> {
         description: 'Graasp Backend API used to serve data to frontend applications',
       },
     },
+    refResolver: {
+      buildLocalReference(json, baseUri, fragment, i) {
+        return json.$id?.toString() || `my-fragment-${i}`;
+      },
+    },
```
ReidyT, line 21: I'm not sure to understand this part, maybe it could be good to add a comment for clarity?

ReidyT (changes requested):
I didn't understand the refResolver, so it would be nice to add a comment, but the rest of the code looks good to me!

src/plugins/swagger.ts
```diff
@@ -14,6 +14,12 @@ export default async function (instance: FastifyInstance): Promise<void> {
         description: 'Graasp Backend API used to serve data to frontend applications',
       },
     },
+    refResolver: {
+      // define generated name of the definition
+      buildLocalReference(json, baseUri, fragment, i) {
+        return json.$id?.toString() || `my-fragment-${i}`;
+      },
+    },
```
spaenleh, line 22: I am not sure to understand the purpose of this bit of code ... ?

spaenleh (dismissed):
I am not sure to understand the purpose of the building ref segment. Can you explain how it works and why it is necessary ? In the current state it is hard to evaluate whether it is the best solution.

pyphilia:
The openapi spec is generated by swagger and for example a minimal account schema ref is generated as follow:
```
"def-1": {
        "additionalProperties": false,
        "type": "object",
        "properties": {
          "id": {
            "format": "uuid",
            "type": "string"
          },
          "name": {
            "format": "username",
            "minLength": 2,
            "maxLength": 60,
            "type": "string"
          }
        },
        "required": [
          "id",
          "name"
        ],
        "title": "Minimal Account"
      },
```
Notice `def-1`. It's the default name generation. It's then referred as follow:
```
"hidden": {
            "$ref": "#/components/schemas/def-12"
          },
```

With this bit of code, we can define a name given the id we defined. For minimal account, it will be
```
"minimalAccount": { ... }
"$ref": "#/components/schemas/minimalAccount"
```

Not sure how to explain with comments only, do you want me to add more details?

ReidyT (approved):
Thanks for this change! LGTM 🪐

PR #1608, opened 2024-11-04
docs: add documentation for actions

pyphilia:
I didn't add too much description on aggregate, since I think the goal is to have an endpoint per graph. Or just not have such graphs.

ref #1528


src/services/item/plugins/action/base-analytics.ts
```diff
@@ -10,7 +10,22 @@ import { Item } from '../../entities/Item';
...
+    id: { type: 'string' },
+    name: { type: 'string' },
+    email: { type: 'string' },
+    extra: {
+      type: 'object',
+      additionalProperties: false,
+      properties: { lang: { type: 'string' } },
+    },
+  },
+};
```
pyphilia, line 28: I moved this here without changes, since we use ajv for validation, which is different from the endpoint schemas.

ReidyT (approved):
Looks really good!

PR #1612, opened 2024-11-04
feat(refactor): simplify the short links by using new SDK types

ReidyT:


src/services/item/plugins/shortLink/schemas.ts
```diff
@@ -43,57 +46,49 @@ export const getAllByItem = {
...
   params: customType.StrictObject({
     itemId: customType.UUID(),
   }),
-  response: { [StatusCodes.OK]: Type.Array(shortLinkPayload), '4xx': errorSchemaRef },
+  response: {
+    [StatusCodes.OK]: Type.Object({
+      [ShortLinkPlatform.Builder]: Type.Optional(Type.String()),
+      [ShortLinkPlatform.Player]: Type.Optional(Type.String()),
+      [ShortLinkPlatform.Library]: Type.Optional(Type.String()),
+    }),
```
pyphilia, line 60: The description does not match the schema. 
- Missing returned empty object. If empty does it return with another status code?
- Also does it always return all three platforms?

pyphilia (changes requested):
LGTM! I just have a question about the schema, it's a nice refactor!

src/services/item/plugins/shortLink/repository.ts
```diff
@@ -88,26 +91,29 @@ export class ShortLinkRepository extends MutableRepository<ShortLink, UpdateShor
...
 
-    if (!shortLink) {
-      throw new ShortLinkNotFound(alias);
-    }
+    try {
+      await this.repository.update({ alias }, entity);
 
-    return shortLink;
-  }
+      const updatedEntity = await this.getOne(entity.alias);
```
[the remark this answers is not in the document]
ReidyT: As discussed earlier, I have to return it to invalidate the query cache correctly.

spaenleh (approved):
Very clean. I like how you defined the DTO class, which does the explicit mapping to the exit type. Maybe we should annotate the function type explicitly ? Anyway I think it is ver clear and readable. Well done ! 👍

src/services/item/plugins/shortLink/schemas.ts
```diff
@@ -43,57 +46,55 @@ export const getAllByItem = {
...
+  description:
+    'Get all short links created for an item. The response could be an empty object or a key-value with at least one platform and the alias.',
 
   params: customType.StrictObject({
     itemId: customType.UUID(),
   }),
-  response: { [StatusCodes.OK]: Type.Array(shortLinkPayload), '4xx': errorSchemaRef },
+  response: {
+    [StatusCodes.OK]: Type.Union([
+      Type.Object(
```
pyphilia, line 57: You could use `customType.StrictObject` to avoid repeating `additionalProperties: false`

src/services/item/plugins/shortLink/schemas.ts
```diff
@@ -43,57 +46,55 @@ export const getAllByItem = {
...
+        {
+          [ShortLinkPlatform.Builder]: Type.Optional(Type.String()),
+          [ShortLinkPlatform.Player]: Type.Optional(Type.String()),
+          [ShortLinkPlatform.Library]: Type.Optional(Type.String()),
+        },
+        { additionalProperties: false },
+      ),
+      Type.Object({}, { additionalProperties: false }),
```
pyphilia, line 65: Same as above.

src/services/item/plugins/shortLink/schemas.ts
```diff
@@ -43,57 +46,55 @@ export const getAllByItem = {
...
-  response: { [StatusCodes.OK]: Type.Array(shortLinkPayload), '4xx': errorSchemaRef },
+  response: {
+    [StatusCodes.OK]: Type.Union([
+      Type.Object(
+        {
+          [ShortLinkPlatform.Builder]: Type.Optional(Type.String()),
+          [ShortLinkPlatform.Player]: Type.Optional(Type.String()),
+          [ShortLinkPlatform.Library]: Type.Optional(Type.String()),
+        },
+        { additionalProperties: false },
```
pyphilia, line 63: I think you can add `minProperties: 1` to apply "at least one"

pyphilia (approved):
LGTM! I added suggestions to improve even more your schema ^^

PR #1616, opened 2024-11-05
docs: add documentation to item endpoints

pyphilia:
ref #1528


src/services/item/schemas.create.ts
```diff
@@ -62,87 +54,94 @@ function itemCreateSchemaFactoryWithSettings<
...
-  Type.Object({ target: customType.UUID() }, { additionalProperties: false }),
+  customType.StrictObject({ target: customType.UUID() }),
 );
 
 export const create = {
+  operationId: 'createItem',
+  tags: ['item'],
+  summary: 'Create item',
+  description:
+    'Create item, whose possible types are folder, app, document, embeddedLink, localFile, s3File, etherpad, h5p and shortcut.',
```
spaenleh, line 141: This could be generated from the ItemType enum ?

src/services/item/schemas.packed.ts
```diff
@@ -36,35 +34,62 @@ export const packedItemSchemaRef = registerSchemaAsRef(
       thumbnails: Type.Optional(Type.Object({}, { additionalProperties: true })),
     },
     {
-      // Schema Options
-      additionalProperties: false,
+      description: 'Item with additional information',
```
spaenleh, line 37: Should we be more precise ?

src/services/item/schemas.ts
```diff
@@ -72,168 +44,145 @@ export const settingsSchema = Type.Partial(
...
-    }),
-    '4xx': errorSchemaRef,
-  },
-} as const satisfies FastifySchema;
-
 export const reorder = {
+  operationId: 'reorderItem',
+  tags: ['item'],
+  summary: 'Reorder item',
+  description: 'Reorder item within its parent given previous item id.',
```
spaenleh, line 157:
```suggestion
  description: 'Reorder item inside its parent. Place it after item denoted by the `previousItemId` parameter.',
```

src/services/item/schemas.ts
```diff
@@ -72,168 +44,145 @@ export const settingsSchema = Type.Partial(
...
+
   params: customType.StrictObject({
-    id: customType.UUID(),
+    id: customType.UUID({ description: 'Item to reorder' }),
+  }),
+  body: customType.StrictObject({
+    previousItemId: customType.UUID({
+      description: 'Item which the item defined in params should go after',
```
spaenleh, line 164: Too many `item` in this sentence.

src/services/item/schemas.ts
```diff
@@ -72,168 +44,145 @@ export const settingsSchema = Type.Partial(
...
-        uniqueItems: true,
-      }),
-    },
-    { additionalProperties: false },
-  ),
+  operationId: 'deleteManyItems',
+  tags: ['item'],
+  summary: 'Delete many items',
+  description:
+    'Delete many items given their ids. This endpoint is asynchronous and a feedback is returned through websockets.',
```
spaenleh, line 178:
```suggestion
    'Delete many items given their ids. This endpoint is asynchronous (response is given via HTTP before completion of the task) and a feedback is returned through websockets.',
```

spaenleh (approved):
Thank you for these improvements. I have added some comments. I know writing is hard, maybe we can find better descriptions for some of these.

ReidyT (approved):
Nice work! LGTM, I agree with the Basile's comments 😶‍🌫️

src/services/item/schemas.create.ts
pyphilia: Potentially we will remove it, so I let it be.

src/services/item/schemas.packed.ts
pyphilia: I will add more details on each prop instead.

PR #1628, opened 2024-11-06
feat: add temperature in query params for chatbot API

swouf:


src/services/item/plugins/app/chatBot/openAICompletion.ts
```diff
@@ -15,11 +24,20 @@ export const openAICompletion = async (body: Array<ChatBotMessage>, gptVersion:
...
   return completion;
 };
 
+export function validateTemperature(temperature?: number) {
+  if (temperature) {
+    if (temperature < OPENAI_MIN_TEMP || temperature > OPENAI_MAX_TEMP) {
+      throw new OpenAIBadTemperature(temperature, OPENAI_MAX_TEMP, OPENAI_MIN_TEMP);
+    }
+  }
+}
```
spaenleh, line 39: This could be handled by a validation in the schema.

src/services/item/plugins/app/chatBot/schemas.ts
```diff
@@ -32,6 +32,7 @@ export const create = {
     gptVersion: Type.Optional(
       Type.Enum(GPTVersion, { description: 'Model to use', default: OPENAI_GPT_VERSION }),
     ),
+    temperature: Type.Optional(Type.Number({ maximum: OPENAI_MAX_TEMP, minimum: OPENAI_MIN_TEMP })),
```
spaenleh, line 35: Ah so you already have it. I guess this is sufficient.

swouf: Yes, I found that after implementing the validation manually. I'll keep just that then.

src/utils/config.ts
```diff
@@ -414,6 +414,10 @@ export const OPENAI_GPT_VERSION = getGptVersion();
 export const OPENAI_API_KEY = process.env.OPENAI_API_KEY;
 export const OPENAI_ORG_ID = process.env.OPENAI_ORG_ID;
 
+export const OPENAI_MAX_TEMP = 2;
+export const OPENAI_MIN_TEMP = 0;
```
pyphilia, line 418:
```suggestion
export const OPENAI_MAX_TEMPERATURE = 2;
export const OPENAI_MIN_TEMPERATURE = 0;
```

src/utils/errors.ts
```diff
@@ -423,6 +423,13 @@ export class OpenAIBadVersion extends OpenAIBaseError {
   }
 }
 
+export class OpenAIBadTemperature extends OpenAIBaseError {
+  constructor(temperature: number, maxTemperature: number, minTemperature: number) {
+    const message = `The temperature must be between ${minTemperature} and ${maxTemperature}. You set ${temperature}.`;
```
pyphilia, line 428: Reuse your constants here?

pyphilia (approved):
Thanks for the PR! I've added minor suggestions, otherwise LGTM! 🪜

spaenleh (approved):
Looks good to me !

src/utils/errors.ts
swouf: Good question. In the end, I'm not even using this error anymore. I'm doing the validation in the schema. Maybe I'll just remove it.

PR #1633, opened 2024-11-08
docs: add docs to member endpoints

pyphilia:
I added docs on member endpoints.

I also removed get many by id, and get by email endpoints.

ref #1528


spaenleh (changes requested):

ReidyT (commented):
Looks good! I agree with Basile's suggestions

spaenleh (approved):

PR #1634, opened 2024-11-08
refactor: refactor enroll endpoint

pyphilia:
Create EnrollService to be able to use repositories when needed (get item without permission, create membership without permission).

Next steps would be to remove `throwOnForbiddenPermission` for get items in ItemService.

Added some `private`.

### DIFF
--- a/src/services/item/plugins/enroll/service.ts
+++ b/src/services/item/plugins/enroll/service.ts
@@ -1,0 +1,54 @@
+import { singleton } from 'tsyringe';
+
+import { ItemLoginSchemaStatus, PermissionLevel, UUID } from '@graasp/sdk';
+
+import { Repositories } from '../../../../utils/repositories';
+import { hasPermission } from '../../../authorization';
+import {
+  CannotEnrollFrozenItemLoginSchema,
+  CannotEnrollItemWithoutItemLoginSchema,
+} from '../../../itemLogin/errors';
+import { ItemLoginService } from '../../../itemLogin/service';
+import { ItemMembershipAlreadyExists } from '../../../itemMembership/plugins/MembershipRequest/error';
+import { Member } from '../../../member/entities/member';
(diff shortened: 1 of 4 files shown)


MatveyK (approved):
LGTM ! I really like what I'me seeing here. I much prefer this division of responsibilities between service and the controller. Also the removal of the `throwOnForbiddenPermission` is very pleasant to me.

src/services/item/plugins/enroll/service.ts
```diff
@@ -0,0 +1,48 @@
...
+  CannotEnrollItemWithoutItemLoginSchema,
+} from '../../../itemLogin/errors';
+import { ItemLoginService } from '../../../itemLogin/service';
+import { ItemMembershipAlreadyExists } from '../../../itemMembership/plugins/MembershipRequest/error';
+import { Member } from '../../../member/entities/member';
+import { ItemService } from '../../service';
+
+@singleton()
+export class EnrollService {
+  private itemService: ItemService;
```
ReidyT, line 18:
```suggestion
```

src/services/item/plugins/enroll/service.ts
```diff
@@ -0,0 +1,48 @@
...
+  private itemLoginService: ItemLoginService;
+
+  constructor(itemService: ItemService, itemLoginService: ItemLoginService) {
+    this.itemService = itemService;
+    this.itemLoginService = itemLoginService;
+  }
```
ReidyT, line 24:
```suggestion
  constructor(itemLoginService: ItemLoginService) {
    this.itemLoginService = itemLoginService;
  }
```

ReidyT (approved):
Really nice changes, LGTM 🦄

spaenleh (approved):
Looks good like this. Just apply suggestion by @ReidyT to remove the dependency on the itemService if you don't use it.

PR #1636, opened 2024-11-08
refactor: remove member hooks

pyphilia:
- Remove member hooks: only one use is to create memberships from invitations for the new registered emails
- Create invitation utils for tests.
- Move /register tests in its own file, with regular tests and related invitation creation tests.

### DIFF
--- a/src/services/member/service.ts
+++ b/src/services/member/service.ts
@@ -14,14 +14,12 @@ import {
   EMAIL_CHANGE_JWT_SECRET,
 } from '../../utils/config';
 import { MemberAlreadySignedUp } from '../../utils/errors';
-import HookManager from '../../utils/hook';
 import { Repositories } from '../../utils/repositories';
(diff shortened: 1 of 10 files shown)


src/services/member/service.ts
```diff
@@ -14,14 +14,12 @@ import {
...
 } from '../../utils/config';
 import { MemberAlreadySignedUp } from '../../utils/errors';
-import HookManager from '../../utils/hook';
 import { Repositories } from '../../utils/repositories';
 import { NEW_EMAIL_PARAM, SHORT_TOKEN_PARAM } from '../auth/plugins/passport';
 import { Actor, Member } from './entities/member';
 
 @singleton()
 export class MemberService {
-  hooks = new HookManager();
```
ReidyT, line 24: 🥳

ReidyT (approved):
🚀

PR #1645, opened 2024-11-15
feat: add delete tag for item

### LINKED ISSUE #1644: Implement DELETE /items/<itemid>/tags/<tagId>

pyphilia:
close #1644


src/services/item/plugins/tag/ItemTag.repository.test.ts
```diff
@@ -105,4 +105,54 @@ describe('ItemTag Repository', () => {
...
+      });
+      expect(result1.tag.id).toEqual(tag1.id);
+    });
+    it('does not throw if tag is not associated with item', async () => {
+      const tag = await tagRawRepository.save(TagFactory());
+      const item = await itemRawRepository.save(FolderItemFactory({ creator: null }));
+      await itemTagRawRepository.save({ tag, item });
+
+      expect(await repository.delete(item.id, tag.id)).toBeUndefined();
+    });
```
spaenleh, line 156: Not sure to understand this test ? Maybe add a comment when you save the tag with the item why this is different from the "normal" behaviour.

src/services/item/plugins/tag/controller.test.ts
```diff
@@ -21,9 +22,11 @@ const tagRawRepository = AppDataSource.getRepository(Tag);
 const itemTagRawRepository = AppDataSource.getRepository(ItemTag);
 
 const createTagsForItem = async (item: Item, tags: Tag[]) => {
```
spaenleh, line 24: Should we explicitly type this function's return ?

src/services/item/plugins/tag/controller.test.ts
```diff
@@ -203,6 +207,67 @@ describe('Tag Endpoints', () => {
...
+
+    describe('Signed out', () => {
+      it('Cannot delete tag', async () => {
+        const response = await app.inject({
+          method: HttpMethod.Delete,
+          url: `/items/${v4()}/tags/${v4()}`,
+        });
+        expect(response.statusCode).toBe(StatusCodes.UNAUTHORIZED);
+      });
+    });
```
spaenleh, line 238: How does this differ from the two previous tests ? I do not see explicitly that you deactivate login mocks. Maybe you forgot them on the other two tests ?

src/services/item/plugins/tag/controller.test.ts
```diff
@@ -203,6 +207,67 @@ describe('Tag Endpoints', () => {
...
+      });
+      expect(response.statusCode).toBe(StatusCodes.BAD_REQUEST);
+    });
+    it('Throw for invalid tag id', async () => {
+      const response = await app.inject({
+        method: HttpMethod.Delete,
+        url: `/items/${v4()}/tags/invalid}`,
+      });
+      expect(response.statusCode).toBe(StatusCodes.BAD_REQUEST);
+    });
```
spaenleh, line 228: Should these be in the signed in section ?

src/services/item/plugins/tag/schemas.ts
```diff
@@ -58,3 +58,20 @@ export const createTagForItem = {
     '5xx': errorSchemaRef,
   },
 } as const satisfies FastifySchema;
+
+export const deleteTagForItem = {
+  operationId: 'deleteTagForItem',
+  tags: ['tag'],
+  summary: 'Delete tag for item',
```
spaenleh, line 65:
```suggestion
  summary: 'Delete tag associated with item',
```

src/services/item/plugins/tag/service.ts
```diff
@@ -39,4 +39,13 @@ export class ItemTagService {
 
     return await itemTagRepository.getByItemId(itemId);
   }
+
+  async delete(actor: Actor, repositories: Repositories, itemId: UUID, tagId: UUID) {
```
spaenleh, line 43: I think the actor can be typed more strictly, as a member for example.

src/services/item/plugins/tag/controller.ts
```diff
@@ -37,7 +37,23 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+    '/:itemId/tags/:tagId',
+    {
+      schema: deleteTagForItem,
+      preHandler: [isAuthenticated, matchOne(validatedMemberAccountRole)],
+    },
+    async ({ user, params: { itemId, tagId } }, reply) => {
+      await db.transaction(async (manager) => {
+        const repositories = buildRepositories(manager);
+
+        await itemTagService.delete(user?.account, repositories, itemId, tagId);
```
spaenleh, line 56: Here you should probably use the `assertDefined(user?.account)` so you can be sure that the actor is defined. and later use the `Member` type. This makes it more clear to the folowing services that we are dealing with an authentication required service function.

spaenleh (changes requested):
Thank you for your work on this feature ! It looks good and clean. 

I have added somme comments and maybe we need to check for the tests. So I would be happy to do another review once you have fixed them ! 💪 

🚆

MatveyK (approved):
LGTM ! I share some of the remarks of @spaenleh

src/services/item/plugins/tag/controller.test.ts
pyphilia: Previous tests have incorrect body, and it throws because of that even before checking for an authenticated user. This test can go through the authentication check because the request is correct.

src/services/item/plugins/tag/controller.test.ts
pyphilia: It does not even check authentication, so it's valid for both unauthenticated and authenticated users.

spaenleh: As discussed just add more context before the test.

src/services/item/plugins/tag/controller.ts
```diff
@@ -37,7 +38,23 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+  fastify.delete(
+    '/:itemId/tags/:tagId',
+    {
+      schema: deleteTagForItem,
+      preHandler: [isAuthenticated, matchOne(validatedMemberAccountRole)],
+    },
+    async ({ user, params: { itemId, tagId } }, reply) => {
+      await db.transaction(async (manager) => {
+        const repositories = buildRepositories(manager);
+        assertIsMember(user?.account);
```
spaenleh, line 56: Should we do this before the transaction ? It would not change the behaviour and is actually just a was to force TS to narrow the type since we have the auth handler before.

PR #1651, opened 2024-11-20
feat: add tag count search

### LINKED ISSUE #1647: Implement GET Tags with occurrence count for a search

pyphilia:
I've added `/tags?search=<search>&category=<category>`where search is mandatory, and category is optional.

close #1647

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -42,8 +42,7 @@ export default async function (instance: FastifyInstance): Promise<void> {
     .register(fp(metaPlugin))
     .register(fp(passportPlugin))
     // need to be defined before member and item for auth check
-    .register(fp(authPlugin))
-    .register(tagPlugin);
+    .register(fp(authPlugin));
 
   instance.register(async (instance) => {
     // core API modules
@@ -64,7 +63,8 @@ export default async function (instance: FastifyInstance): Promise<void> {
       })
       .register(fp(MemberServiceApi))
       .register(fp(ItemServiceApi))
-      .register(fp(ItemMembershipServiceApi));
+      .register(fp(ItemMembershipServiceApi))
+      .register(tagPlugin);
   });
 }
 
--- a/src/app.ts
+++ b/src/app.ts
@@ -15,6 +15,7 @@ import { plugin as passportPlugin } from './services/auth/plugins/passport';
 import ItemServiceApi from './services/item';
 import ItemMembershipServiceApi from './services/itemMembership';
 import MemberServiceApi from './services/member';
+import tagPlugin from './services/tag/controller';
 import websocketsPlugin from './services/websockets';
 import {
   DATABASE_LOGS,
@@ -41,7 +42,8 @@ export default async function (instance: FastifyInstance): Promise<void> {
     .register(fp(metaPlugin))
     .register(fp(passportPlugin))
     // need to be defined before member and item for auth check
-    .register(fp(authPlugin));
+    .register(fp(authPlugin))
+    .register(tagPlugin);
 
   instance.register(async (instance) => {
     // core API modules
--- a/src/services/item/plugins/tag/ItemTag.repository.ts
+++ b/src/services/item/plugins/tag/ItemTag.repository.ts
@@ -1,11 +1,12 @@
 import { EntityManager } from 'typeorm';
 
-import { UUID } from '@graasp/sdk';
+import { TagCategory, UUID } from '@graasp/sdk';
 
 import { AbstractRepository } from '../../../../repositories/AbstractRepository';
 import { IllegalArgumentException } from '../../../../repositories/errors';
 import { isDuplicateEntryError } from '../../../../utils/typeormError';
 import { Tag } from '../../../tag/Tag.entity';
+import { TagCount } from '../../../tag/schemas';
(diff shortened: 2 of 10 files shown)


src/app.ts
```diff
@@ -41,7 +42,8 @@ export default async function (instance: FastifyInstance): Promise<void> {
     .register(fp(metaPlugin))
     .register(fp(passportPlugin))
     // need to be defined before member and item for auth check
-    .register(fp(authPlugin));
+    .register(fp(authPlugin))
+    .register(tagPlugin);
```
spaenleh, line 46: Is there a reason why you register it here instead of next to the `memberServiceApi` ?

src/services/item/plugins/tag/ItemTag.repository.ts
```diff
@@ -24,6 +25,38 @@ export class ItemTagRepository extends AbstractRepository<ItemTag> {
...
+
+    if (category) {
+      q.where('category = :category', { category });
+    }
+
+    const res = await q.groupBy('t.id').orderBy('count', 'DESC').getRawMany<{
+      name: Tag['name'];
+      category: Tag['category'];
+      count: string; // sql returns string by default
+    }>();
```
spaenleh, line 54: I think in this case the use of the getRawMany is legitimate, but it feels very "raw" with typeorm... 😓

src/services/item/plugins/tag/ItemTag.repository.ts
```diff
@@ -24,6 +25,38 @@ export class ItemTagRepository extends AbstractRepository<ItemTag> {
...
+      .select(['t.name AS name', 't.category AS category', 'count(t.id) as count'])
+      .leftJoinAndSelect('tag', 't', 't.id = itemTag.tag_id AND t.name ILIKE :search', {
+        search: `%${search}%`,
+      });
+
+    if (category) {
+      q.where('category = :category', { category });
+    }
+
+    const res = await q.groupBy('t.id').orderBy('count', 'DESC').getRawMany<{
```
spaenleh, line 50: Should we add a limit ? like max 10 results ?

src/services/tag/controller.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+      expect(response.statusCode).toBe(StatusCodes.OK);
+      const data = response.json();
+      expect(data).toHaveLength(2);
+      for (const d of data) {
+        expect(tags.map((t) => t.name)).toContain(d.name);
+        expect(d.category).toEqual(TagCategory.Discipline);
+        expect(d.count).toBeGreaterThanOrEqual(1);
+      }
+    });
```
spaenleh, line 85: Why does this look like it just loosely checks for the results ? Is it to limit breakage ? We should be able to exactly match the return of the search no ?

spaenleh (approved):
Looks good. 🌈 

I have a few questions and remarks, otherwise looks good !

src/services/tag/controller.test.ts
pyphilia: This test is more loose because it will depend on the other tags added from other tests. 
I check more precisely at the repository level. Here it's more about checking the return values exists and are kinda correct. I could have stricter tests, but I think it could break sometimes (mainly because it will collide on the search string)

src/services/item/plugins/tag/ItemTag.repository.ts
```diff
@@ -24,6 +25,38 @@ export class ItemTagRepository extends AbstractRepository<ItemTag> {
...
+    search: string;
+    category?: TagCategory;
+  }): Promise<TagCount[]> {
+    if (!search) {
+      throw new IllegalArgumentException(`search is invalid: "${search}"`);
+    }
+
+    const q = this.repository
+      .createQueryBuilder('itemTag')
+      .select(['t.name AS name', 't.category AS category', 'count(t.id) as count'])
```
ReidyT, line 41: I didn't remember but are the alias required in TypeORM for the columns name?

Also, concerning the parsing of the count from string to number, you could try this (it works in pure sql):

```suggestion
      .select(['t.name AS name', 't.category AS category', 'count(t.id)::integer AS count'])
```

src/services/item/plugins/tag/ItemTag.repository.ts
```diff
@@ -24,6 +25,38 @@ export class ItemTagRepository extends AbstractRepository<ItemTag> {
...
+    }
+
+    const res = await q.groupBy('t.id').orderBy('count', 'DESC').getRawMany<{
+      name: Tag['name'];
+      category: Tag['category'];
+      count: string; // sql returns string by default
+    }>();
+
+    // transform into integer
+    return res.map((value) => ({ ...value, count: parseInt(value.count) }));
```
ReidyT, line 57: If my suggestion below works correctly, you don't need to transform in the code.

```suggestion
    return res;
```

ReidyT (approved):
Looks good to me, thanks for this PR! Maybe just have a look to my suggestion about parsing to integer directly in the SQL code 😆

src/services/tag/controller.test.ts
spaenleh: okay I see

PR #1653, opened 2024-11-20
fix: sanitize html on upload and import

### LINKED ISSUE #1652: Sanitize html on upload

pyphilia:
close #1652


src/services/file/sanitize.ts
```diff
@@ -0,0 +1,65 @@
...
+    },
+  });
+}
+
+export function sanitizeHtml(content: string): string {
+  return sanitize(content);
+}
+
+export function createSanitizedFile(
+  file,
```
spaenleh, line 41: Can we set a type for this parameter ?

src/services/file/sanitize.ts
```diff
@@ -0,0 +1,65 @@
...
+      rejects(err);
+    });
+
+    file.on('close', async function () {
+      const content = await readFile(tmpFile, {
+        encoding: 'utf8',
+        flag: 'r',
+      });
+
+      const readable = Readable.from(sanitizeFn(content));
```
spaenleh, line 59: Is it possible to stream the sanitization ? Maybe not.

ReidyT (approved):
LGTM really good fix! I agree with the comments of Basile 😶‍🌫️

PR #1664, opened 2024-11-28
fix: use lower limit for consecutive exports

spaenleh:


src/services/action/constants/constants.ts
```diff
@@ -7,6 +7,6 @@ export const MAX_ACTIONS_SAMPLE_SIZE = 10000;
 
 export const ZIP_MIMETYPE = 'application/zip';
 
-export const DEFAULT_REQUEST_EXPORT_INTERVAL = 3600 * 1000 * 24; // 1 day - used for timestamp
+export const DEFAULT_REQUEST_EXPORT_INTERVAL = 3600 * 1000 * 2; // 2hours - used for timestamp
```
ReidyT, line 10:
```suggestion
export const DEFAULT_REQUEST_EXPORT_INTERVAL = 3600 * 1000 * 2; // 2 hours is the default interval for requesting exports, in milliseconds.
```

ReidyT (approved):
LGTM 🚀

PR #1665, opened 2024-11-29
feat: integrate tags to meilisearch, remove categories, migration

pyphilia:


.devcontainer/docker-compose.yml
```diff
@@ -119,8 +119,8 @@ services:
       - MEILI_LOG_LEVEL
     # uncomment below to develop on the meilisearch interface from your local machine
     # by default ports do not need to be exposed as all queries are proxied by the backend
-    # ports:
-    #   - 7700:7700
```
pyphilia, line 123: To revert.

src/migrations/1733808052981-item-tags-to-tag.ts
```diff
@@ -0,0 +1,134 @@
...
+                            substr(
+                                jsonb_array_elements(settings::jsonb->'tags')::text,
+                                2,
+                                length(jsonb_array_elements( settings::jsonb->'tags' )::text) - 2
+                            ),','
+                        )
+                    ),'\n',' '
+                )
+            )
+        ), 'discipline'::tag_category_enum from item
```
spaenleh, line 27: This means current tags are inserted as `disciplines` ? I am ok with that, just wanted to highlight this change.

src/migrations/1733808052981-item-tags-to-tag.ts
```diff
@@ -0,0 +1,134 @@
...
+                replace(
+                    unnest(
+                        -- break comma-separated tags into multiple tags 
+                        string_to_array(
+                            -- remove resulting quotes (") 
+                            substr(
+                                jsonb_array_elements(settings::jsonb->'tags')::text,
+                                2,
+                                length(jsonb_array_elements( settings::jsonb->'tags' )::text) - 2
+                            ),','
```
spaenleh, line 22: Is there a difference when using `->>` and `->` with json fields ? maybe one or the other will return the value without quotes ?

Edit:  Indeed there is https://arc.net/l/quote/fdgxpxhw

TLDR: `->` returns JSON (so strings will be `"string"`) and `->>` returns TEXT (so strings will be `string` without quotations)

So I think you should be able to get the direct text value with the `->>` operator and simplify your query.

src/migrations/1733808052981-item-tags-to-tag.ts
```diff
@@ -0,0 +1,134 @@
...
+          'B1',
+          'CDD course. B2',
+          'collège',
+          'ELE. A1.',
+          'formación profesional',
+          'Lehrplan 21',
+          'MA.2.A.1.i',
+          'MA.2.C.2.h',
+          'Sek1',
+          'Sekundarstufe 1',
```
spaenleh, line 78: These are from the current levels in the database ?

src/migrations/1733808052981-item-tags-to-tag.ts
```diff
@@ -0,0 +1,134 @@
...
+        ],
+      ],
+      [
+        'resource-type',
```
spaenleh, line 82: Do we have a specific definition for these ? if yes maybe it would be interesting to have a little tooltip explaining what to put into this field.

src/services/item/plugins/publication/published/plugins/search/meilisearch.ts
```diff
@@ -88,7 +97,13 @@ export class MeiliSearchWrapper {
     this.logger = logger;
 
     // create index in the background if it doesn't exist
-    this.getIndex();
+    this.getIndex().then(async () => {
+      // set facetting order to count for tag categories
+      await this.meilisearchClient.index(ACTIVE_INDEX).updateFaceting({
+        maxValuesPerFacet: 50,
```
spaenleh, line 103: Can this be changed ? Why is it set to 50 ? Does it have a link to pagination ?

src/services/item/plugins/publication/published/plugins/search/meilisearch.ts
```diff
@@ -144,17 +163,26 @@ export class MeiliSearchWrapper {
 
   // WORKS ONLY FOR PUBLISHED ITEMS
   async search(queries: MultiSearchParams) {
-    const searchResult = await this.meilisearchClient.multiSearch(queries);
-
+    // type should match schema
+    const searchResult =
+      await this.meilisearchClient.multiSearch<
+        Static<(typeof search)['response']['200']>['hits'][0]
```
spaenleh, line 169: Should this type be constructed in the schema file and exported from there ?

src/services/item/plugins/publication/published/test/meilisearch.test.ts
```diff
@@ -326,25 +333,31 @@ describe('MeilisearchWrapper', () => {
...
+          // }),
+          // expect.objectContaining({
+          //   id: descendant2.id,
+          //   discipline: [],
+          //   level: [],
+          //   'resource-type': [],
+          //   content: '',
+          //   isPublishedRoot: true,
+          //   isHidden: false,
+          // }),
```
spaenleh, line 360: Remove ?

spaenleh (changes requested):
Looks good ! Great work ! I have added some comments. Let me know. 
🔍

src/migrations/1733808052981-item-tags-to-tag.ts
pyphilia: Yes.

src/migrations/1733808052981-item-tags-to-tag.ts
pyphilia: Yes, from the current tags.

src/migrations/1733808052981-item-tags-to-tag.ts
pyphilia: I close here, I will do so in builder.

src/services/item/plugins/publication/published/plugins/search/meilisearch.ts
pyphilia: This is the number of facets you get when requesting faceting. Default value is 10, so you usually receive the 10 most used facets. For Library I'd like to show more than 10, so I increased this number. I will add a comment, close this conversation if it is okay for you.

tsconfig.json
```diff
@@ -1,13 +1,13 @@
...
-    "module": "Node16",
+    "lib": ["es2021", "es6", "esnext.asynciterable"],
+    "target": "es2021",
+    "module": "NodeNext",
     // "strict": true, // TODO: enable
     "esModuleInterop": true /* Emit additional JavaScript to ease support for importing CommonJS modules. This enables `allowSyntheticDefaultImports` for type compatibility. */,
     "checkJs": false,
     "allowJs": false,
-    "moduleResolution": "node16",
+    "moduleResolution": "nodenext",
```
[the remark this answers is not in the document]
pyphilia: I wanted to use `string.replaceAll` but I think I can do the same with regex. I will revert, but I believe this change is something we want to go forward with one day, don't we?

src/services/tag/controller.ts
```diff
@@ -1,22 +1,33 @@
...
 import { TagService } from './service';
 
 const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
   const tagService = resolveDependency(TagService);
+  fastify.register(
+    async function (fastify: FastifyInstanceTypebox) {
+      // add CORS support
+      if (fastify.corsPluginOptions) {
+        fastify.register(fastifyCors, fastify.corsPluginOptions);
+      }
```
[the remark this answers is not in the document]
spaenleh: Ah so this is why ... could we move the cors higher up and not include the apps there ? or are the apps too dependent on that ? Maybe we need to shift a few routes. Boundaries seem a bit lacking in this context.

tsconfig.json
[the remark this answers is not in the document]
spaenleh: Yes this is something we want to change. But I would say it requires more changes. For string.replaceAll, you should add to the `lib` key in `compilerOptions` the `"ES2020"` value.

src/services/tag/controller.ts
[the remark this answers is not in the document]
spaenleh: You can close for now, but maybe we should try to schedule that change.

src/migrations/1733808052981-item-tags-to-tag.ts
pyphilia: I tried but actually `->>` returns a `TEXT` which is not an array. Furthermore it has the following syntax "["a","b"]" which is not the native syntax for array in sql. So I think I better keep my current implementation, as I will anyway have to deal with quotes anyway.

src/services/item/plugins/publication/published/plugins/search/meilisearch.ts
pyphilia: https://www.meilisearch.com/docs/reference/api/settings#faceting-object
I think the exact definition of `facet` is more the **set** of used tag + count.

src/services/tag/controller.ts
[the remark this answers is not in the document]
pyphilia: You're right, I created an issue for that.

src/migrations/1733808052981-item-tags-to-tag.ts
spaenleh: Ah... yes if you are dealing with an array as the value, then my suggestion does not make a lot of sens. I am sorry. Looks good like this.

spaenleh (approved):

PR #1679, opened 2024-12-10
fix: return no content for no avatar

pyphilia:
Hopefully it will solve this problem
```
GET /members/:id/avatar/:size
'S3_FILE_NOT_FOUND' captured as exception with message 'S3_FILE_NOT_FOUND'
environment: production  transaction: GET /members/:id/avatar/:size
```

- [x] test on dev
- [x] write tests


src/services/member/plugins/thumbnail/schemas.ts
```diff
@@ -13,21 +13,22 @@ export const upload = {
...
+  params: customType.StrictObject({
+    id: customType.UUID(),
+    size: Type.Enum(ThumbnailSize, { default: ThumbnailSize.Medium }),
+  }),
+  querystring: customType.StrictObject({
+    replyUrl: Type.Boolean({ default: false }),
+  }),
   response: {
-    [StatusCodes.OK]: errorSchemaRef,
+    [StatusCodes.OK]: {},
```
pyphilia, line 30: Check if this is okay.

src/services/member/plugins/thumbnail/schemas.ts
```diff
@@ -27,7 +27,7 @@ export const download = {
     replyUrl: Type.Boolean({ default: false }),
   }),
   response: {
-    [StatusCodes.OK]: {},
+    [StatusCodes.OK]: Type.String({ description: 'Url string of the avatar if replyUrl is true' }),
```
spaenleh, line 30: Not to be changed now, but do we really need to support replying with the file as a blob ? i.e. can we remove the `replyUrl` param and always rely with the url instead ?

src/services/member/plugins/thumbnail/test/index.test.ts
```diff
@@ -50,87 +56,133 @@ jest.mock('@aws-sdk/lib-storage', () => {
...
+
+    it('Get member avatar', async () => {
+      const member = await saveMember(MemberFactory({ extra: { hasAvatar: true } }));
 
       const response = await app.inject({
         method: HttpMethod.Get,
         url: `members/${member.id}/avatar/${ThumbnailSize.Small}`,
       });
-
+      console.log(response);
```
spaenleh, line 78: remove ?

src/services/member/plugins/thumbnail/schemas.ts
pyphilia: Yes, I think we never use one of the case.

PR #1681, opened 2024-12-10
fix: add item in app action

### LINKED ISSUE #1680: Add item id in app actions on export
This code is missing the item relations to include the item infos on export.
```

  async getForItem(itemId: string, filters: SingleItemGetFilter) {
    if (!itemId) {
      throw new IllegalArgumentException('The itemId must be defined');
    }

    const { accountId } = filters;
    return await this.repository.find({
      where: { item: { id: itemId }, account: { id: accountId } },
      relations: { account: true },
    });
  }
```

pyphilia:
- [x] check it is okay to add the relations
- [x] write tests

close #1680

### DIFF
--- a/src/services/item/plugins/action/service.test.ts
+++ b/src/services/item/plugins/action/service.test.ts
@@ -9,6 +9,7 @@ import {
   AggregateMetric,
   Context,
   CountGroupBy,
+  ItemType,
   PermissionLevel,
 } from '@graasp/sdk';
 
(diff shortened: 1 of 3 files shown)


spaenleh (approved):
Change look good. But we did not even get the itemId ? Now we will get the full item data for each action right ?

pyphilia:
Yes, before we didn't even include the item id. But you're right, I'll see if it's okay with @swouf but it should be ok to send only the item id.

PR #1689, opened 2024-12-13
docs: add set default settings sql for postgres

pyphilia:


bootstrapDB.sql
```diff
@@ -10,3 +10,8 @@ create database etherpad with owner etherpad;
 -- Test database
 create user test with password 'test';
 create database test with owner test;
+
+-- set timeout settings for postgres
+-- should avoid transactions to hang for too long
+set idle_in_transaction_session_timeout = '3600000';
+set statement_timeout = '3600000';
```
spaenleh, line 17: I guess these are in miliseconds but we could say it in the comment. Also say what this is supposed to be (1h).

spaenleh (approved):

PR #1703, opened 2024-12-18
fix: analytics query schema error

spaenleh:
In this PR: 
- fix some schema issues with the aggregation and action endpoints used by analytics.

Question:
@pyphilia Can you give me feedback on these issues:
- I have modified the appActionSchema definition in the analytics return since we do not return the deprecated `member` attribute in the data. Do you think I should instead create a "standard appAction schema" and then extend that one with the legacy `member` prop for use in the appAction endpoint ? So all other schemas that may build on top of appAction can use the "standard" instead of having to each time strip the legacy prop ?
- I have removed the `itemMemberships` props from the return. But we still retrieve it from the DB and process it. Ideally we would use a different function.Should this be done now, or later ? 
- I haven't implemented specific tests for this, and it looks like there were no such tests before since the issue was present in production. How much coverage do you think we should have on this ? If we plan on improving these endpoints, do we need to fully cover their current functionality ?

### DIFF
--- a/src/plugins/typebox.ts
+++ b/src/plugins/typebox.ts
@@ -70,7 +70,7 @@ export const customType = {
   Username: (options?: StringOptions) =>
     Type.String({
       ...options,
-      format: 'username',
+      format: 'graaspUsername',
       minLength: MIN_USERNAME_LENGTH,
       maxLength: MAX_USERNAME_LENGTH,
     }),
@@ -81,7 +81,7 @@ export const customType = {
     }),
   EnumString: <T extends string[]>(values: [...T], options?: SchemaOptions) =>
     Object.assign(
-      /* 
+      /*
       Object Assign is used so the return type contains the intersection with `{ type: 'string' }`,
       and so can be used in combination with `customType.Nullable(...)`
        */
--- a/src/services/action/entities/action.ts
+++ b/src/services/action/entities/action.ts
@@ -42,7 +42,7 @@ export class Action extends BaseEntity {
     nullable: false,
     enum: Object.values(Context),
   })
-  view: Context | 'Unknown';
+  view: Context;
 
   @Column({
     nullable: false,
--- a/src/services/item/plugins/action/base-analytics.ts
+++ b/src/services/item/plugins/action/base-analytics.ts
@@ -1,6 +1,6 @@
 import { Ajv } from 'ajv';
 
-import { UUID } from '@graasp/sdk';
+import { MemberConstants, UUID } from '@graasp/sdk';
 
 import { Account } from '../../../account/entities/account';
 import { Action } from '../../../action/entities/action';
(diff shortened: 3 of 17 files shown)


src/services/action/entities/action.ts
```diff
@@ -42,7 +42,7 @@ export class Action extends BaseEntity {
     nullable: false,
     enum: Object.values(Context),
   })
-  view: Context | 'Unknown';
+  view: Context;
```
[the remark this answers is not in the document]
pyphilia: From my old dump 'unknown' is indeed the value we save, so I guess it's safe to remove it.

src/services/item/plugins/action/base-analytics.ts
```diff
@@ -68,13 +62,17 @@ export class BaseAnalytics {
     // TODO: all other schemas
 
     // validate and remove additional properties from member
-    const ajv = new Ajv({ removeAdditional: 'all' });
+    const ajv = new Ajv({ removeAdditional: 'all' }).addFormat(
+      'username',
```
pyphilia, line 66: Maybe add a comment why we need it. It might mean that we should add all the other definitions? But for now if it solves your problem it's alright.

pyphilia (changes requested):
LGTM, I'll let you fix what we've discussed.
I'll also let you open issues for things we should refactor (ie. item memberships, action types, etc.)

As for tests, I think it would be safer to have one or two tests to be sure it returns successfully. Something like a test that should return both member and guest, a test that as both types of actions. It doesn't have to be precise, but just make sure it returns 200.

src/services/item/plugins/action/base-analytics.ts
spaenleh: Actually this did not solve the issue of the un-recognized format since the error was not originating from here.
I have removed this change and instead found a fix for the format.

spaenleh:
@pyphilia I have made the tests, I will rebase to resolve the conflict. I have created an issue about the splitting of concerns for the analytics with baseAnalytics.

src/plugins/typebox.ts
```diff
@@ -70,7 +70,7 @@ export const customType = {
   Username: (options?: StringOptions) =>
     Type.String({
       ...options,
-      format: 'username',
+      format: 'graaspUsername',
```
pyphilia, line 73: Was there a reason for renaming? Was `username` an existing format?

pyphilia (approved):
LGTM! Nice splitting 💯

src/plugins/typebox.ts
spaenleh: From my experiments, it looks like `username` might have already been defined by a different extension. Replacing the usage by `graaspUsername` seemed to solve the "count not find format `username`. So if it is ok for you to change the name of the format I think it would be better and solve an issue we had.

PR #1707, opened 2024-12-19
fix: add migration for downgrading guest memberships

### LINKED ISSUE #1706: downgrade guest memberships
Guests cannot have admin or write memberships.
Because of legacy implementation, some of them might have those rights.

pyphilia:
close #1706


spaenleh (approved):
Very nice !

PR #1713, opened 2025-01-08
feat: allow readers to edit etherpad

### LINKED ISSUE #724: [etherpad] Pseudonymized members see read-only pad in builder mode
Pseudonymized members obtain a read-only pad in builder mode

- Is that the expected behaviour?
- Otherwise, ensure that a pseudonymized logged in user can still edit the pad (i.e. is member but does not have membership)

pyphilia:
close #724


MatveyK (approved):
LGTM.

Issue #1714, opened 2025-01-09
Include tags in liked collections, recent collections, etc...


pyphilia:
Should we use meilisearch to get such items? That would be faster and keep the same type.

PR #1720, opened 2025-01-15
feat(published-search): allow creator id filter

pyphilia:
Allow creator id in filter for search in meilisearch. The goal is to use this call in library.

I didn't remove the current `/collections/<memberId>` because it is used by builder (packed collections). Because it requires packed info, it won't be so easy to replace with the search call. So it will require some changes from the front first to remove the call definitely.

ref #1714

### DIFF
--- a/src/services/item/plugins/publication/published/plugins/search/meilisearch.ts
+++ b/src/services/item/plugins/publication/published/plugins/search/meilisearch.ts
@@ -65,6 +65,7 @@ const FILTERABLE_ATTRIBUTES: (keyof IndexItem)[] = [
   'isHidden',
   'lang',
   'likes',
+  'creator',
   ...Object.values(TagCategory),
 ];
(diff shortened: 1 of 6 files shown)


src/services/item/plugins/publication/published/plugins/search/meilisearch.ts
```diff
@@ -65,6 +65,7 @@ const FILTERABLE_ATTRIBUTES: (keyof IndexItem)[] = [
   'isHidden',
   'lang',
   'likes',
+  'creator',
```
spaenleh, line 68: Is this a typo ?
```suggestion
  'creatorId',
```

spaenleh (changes requested):
Looks good, I am not sure about the `creatorId` paroperty. Does the return contain the full `creator` ? of is it a typo ?

spaenleh (approved):
Looks good !

PR #1725, opened 2025-01-16
refactor: improve typings, get folder item

pyphilia:
- Fix some typings.
- Add `get` a Folder Item

ref #1693

### DIFF
--- a/src/services/item/plugins/publication/validation/index.ts
+++ b/src/services/item/plugins/publication/validation/index.ts
@@ -18,10 +18,10 @@ import {
   ItemOpFeedbackEvent,
   memberItemsTopic,
 } from '../../../ws/events';
+import { FolderItemService } from '../../folder/service';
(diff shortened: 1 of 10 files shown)


src/services/item/plugins/publication/validation/index.ts
```diff
@@ -83,7 +84,12 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
       db.transaction(async (manager) => {
         const repositories = buildRepositories(manager);
         // get item and check permission
-        const item = await itemService.get(member, repositories, itemId, PermissionLevel.Admin);
+        const item = await folderItemService.get(
```
spaenleh, line 87: Should we add a comment that lets us know we rely on querying the folderService to only allow folder items to be published ?

spaenleh (approved):
It is nice to see improved typings ! Maybe at some point we will be able to set strict ? 😄

Issue #995, opened 2024-04-26
Generate mock data from a succinct object


Defferrard:
Currently, mock data can be pushed on the database from the `test/mock/index.ts/seed(...)` function. But the parameter is too verbose to be used in integration tests.

We could implement a function that generate the parameters needed to this function from a succinct object.

By example : 
```json
{
   "items":[
      {
         "name":"1st folder"
         "type": "folder"
         "$readers":[
            {
               "$count":2
            }
         ]
      }
   ]
}
```
This object should be transformed to the parameter needed to generate a folder `1st folder` that include 2 members as readers.

PR #1728, opened 2025-01-22
fix(seed): seed from json data

pyphilia:
First version for using the seed in the tests with a succinct format.

I've added a new `ItemFactory` as the one in sdk clashes a lot with the backend types (dates, discriminated item, etc)

ref #995


src/services/item/service.test.ts
```diff
@@ -2,13 +2,13 @@ import { v4 } from 'uuid';
 
 import { FastifyInstance } from 'fastify';
 
-import { FolderItemFactory, ItemVisibilityType } from '@graasp/sdk';
+import { FolderItemFactory, ItemType, ItemVisibilityType, buildPathFromIds } from '@graasp/sdk';
```
spaenleh, line 5: Remove unused ?

test/factories/item.factory.ts
```diff
@@ -0,0 +1,38 @@
...
+ * This factory does not guarantee valid items given their type. But they are acceptable when seed into the db.
+ * Create by default a folder
+ * @param item partial properties
+ * @returns complete item valid to be fed into the database
+ */
+export const ItemFactory = (item: Partial<Item>) => ({
+  type: ItemType.FOLDER,
+  order: null,
+  name: faker.word.words(2),
+  description: item.description ?? faker.lorem.text(),
```
spaenleh, line 17: Why use a nullish here when you spread the item below ? It is not sufficient to spread below ?

test/factories/item.factory.ts
```diff
@@ -0,0 +1,38 @@
...
+        showChatbox: faker.datatype.boolean(),
+        hasThumbnail: false,
+        isResizable: faker.datatype.boolean(),
+        isCollapsible: faker.datatype.boolean(),
+        enableSaveActions: faker.datatype.boolean(),
+        tags: faker.lorem.words({ min: 1, max: 10 }).split(' '),
+        displayCoEditors: faker.datatype.boolean(),
+        ccLicenseAdaption: faker.helpers.enumValue(CCLicenseAdaptions),
+      },
+    ]),
```
spaenleh, line 34: This means you have to provide a full setting, otherwise you will get ill-defined settings ? should we allow to provide a partial setting that can be filled with default values ?

test/mocks/seed.ts
```diff
@@ -0,0 +1,226 @@
...
+    )[];
+  })[];
+};
+
+const replaceActorInItems = (createdActor?: Actor, items?: DataType['items']) => {
+  if (!items?.length) {
+    return [];
+  }
+
+  return items?.map((i) => ({
```
spaenleh, line 76: The optional should not be necessary here.

spaenleh (approved):
Code looks good ! I did not have a super deep look, I think it should be used and adapted if needed, I like how you have decomposed it in smaller chunks. Very nice !

test/factories/item.factory.ts
pyphilia: I would rather say that it forces you to set your necessary settings during tests. So if the default changes, it won't break your test. As we discussed I think both approaches are correct, depends on what you want to achieve. I personally prefer a bit of chaos.

PR #1736, opened 2025-01-29
fix: return 200 and null for empty member profile

pyphilia:
Return 200 instead of 204 in order to return `null` for the frontend.

- Complete `seedFromJson` with `memberProfile`
- Use `seedFromJson` in `controller.test.ts` of member profile

### DIFF
--- a/test/mocks/seed.ts
+++ b/test/mocks/seed.ts
@@ -2,7 +2,13 @@ import { faker } from '@faker-js/faker';
 import { BaseEntity, DataSource } from 'typeorm';
 import { v4 } from 'uuid';
 
-import { ItemType, MemberFactory, PermissionLevel, buildPathFromIds } from '@graasp/sdk';
+import {
+  CompleteMember,
+  ItemType,
+  MemberFactory,
+  PermissionLevel,
+  buildPathFromIds,
+} from '@graasp/sdk';
 
 import { AppDataSource } from '../../src/plugins/datasource';
 import { Item } from '../../src/services/item/entities/Item';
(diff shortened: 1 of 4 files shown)


test/mocks/seed.ts
```diff
@@ -157,6 +171,18 @@ const processItemMemberships = (items: DataType['items'] = []) => {
     }));
 };
 
+/**
+ * Generate ids for members, necessary to further references (for example when creating profiles)
+ * @param members
+ * @returns
```
pyphilia, line 177: complete

test/mocks/seed.ts
```diff
@@ -171,28 +197,35 @@ export async function seedFromJson(data: DataType = {}) {
     items: Item[];
```
pyphilia, line 197: complete docs

test/mocks/seed.ts
```diff
@@ -93,6 +94,7 @@ const replaceActorInItems = (createdActor?: Actor, items?: DataType['items']) =>
 const processActor = async ({ actor, items, members }: DataType) => {
   // create actor if not null
   let createdActor;
+  let memberProfiles;
```
spaenleh, line 97: Why is it plural ? can there be multiple member profiles for the same user ?

test/mocks/seed.ts
```diff
@@ -105,12 +107,24 @@ const processActor = async ({ actor, items, members }: DataType) => {
         },
       })
     ).actor[0];
+
+    // a profile is defined
+    if (actor !== 'actor' && actor?.profile) {
+      memberProfiles = (
```
spaenleh, line 113: Should this be a push operation instead to handle multiple profiles ?

test/mocks/seed.ts
spaenleh: Still TODO ?

spaenleh (approved):
Looks good !

PR #1738, opened 2025-01-30
feat: remove unnecessary host variables

### LINKED ISSUE #1737: refactor env variables to use only one domain graasp.org
We can infer ACCOUNT_HOST, BUILDER_HOST, etc from one domain.

pyphilia:
Remove unnecessary client host variables:
- builder
- player
- auth
- account
- analytics

Get used of the ClientManager to generate client links.

close #1737

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,11 +18,7 @@ env:
   # random keys
   APPS_JWT_SECRET: ${{ vars.MOCK_APPS_JWT_SECRET }}
   APPS_PUBLISHER_ID: 9c9cea73-f3b7-48a3-aa6e-ead82c0685e7 # mock uuid
-  AUTH_CLIENT_HOST: 'localhost:3001'
   AUTH_TOKEN_JWT_SECRET: ${{ vars.MOCK_AUTH_TOKEN_JWT_SECRET }}
-  BUILDER_CLIENT_HOST: 'http://localhost:3111'
-  ACCOUNT_CLIENT_HOST: 'http://localhost:3114'
-  ANALYTICS_CLIENT_HOST: 'http://localhost:3005'
   COOKIE_DOMAIN: localhost
   DB_PASSWORD: docker
   DB_USERNAME: docker
@@ -44,7 +40,6 @@ env:
   MAILER_CONFIG_PASSWORD: password
   MAILER_CONFIG_SMTP_HOST: localhost
   MAILER_CONFIG_USERNAME: username
-  PLAYER_CLIENT_HOST: 'http://localhost:3112'
   RECAPTCHA_SECRET_ACCESS_KEY: 'iamamockkey'
   REDIS_HOST: localhost
   REDIS_PORT: 6379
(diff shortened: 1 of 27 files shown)


src/services/item/plugins/html/h5p/index.ts
```diff
@@ -54,7 +53,8 @@ const plugin: FastifyPluginAsyncTypebox<H5PPluginOptions> = async (fastify) => {
       const html = renderHtml(
         DEFAULT_H5P_ASSETS_ROUTE,
         DEFAULT_H5P_CONTENT_ROUTE,
-        CLIENT_HOSTS.map(({ url }) => url.hostname) ?? ['localhost'],
+        // todo: temporary value
```
MatveyK, line 56: Is this temporary value supposed to stay ?

.github/workflows/test.yml
spaenleh, line 1: Define `CLIENT_HOST` ?

src/services/item/plugins/html/h5p/index.ts
spaenleh: I have some changes that I will push that will change this. We can keep it like this. It is not used in production anyways.

src/utils/config.ts
```diff
@@ -56,43 +47,14 @@ export const TEST = ENVIRONMENT === Environment.test;
...
-export const LIBRARY_HOST = {
-  name: Context.Library,
-  url: new URL(process.env.LIBRARY_CLIENT_HOST ?? DEFAULT_HOST),
-};
-
-export const ACCOUNT_HOST = {
-  name: Context.Account,
-  url: new URL(process.env.ACCOUNT_CLIENT_HOST ?? DEFAULT_HOST),
-};
+export const CLIENT_HOST = process.env.CLIENT_HOST ?? 'http://localhost:3000';
```
spaenleh, line 50: I am not sure about the default for this ... should it not be `http://localhost:3114` or whatever we decide to run the front on ?

src/utils/config.ts
```diff
@@ -56,43 +47,14 @@ export const TEST = ENVIRONMENT === Environment.test;
...
 
-export const ANALYTICS_HOST = {
-  name: Context.Analytics,
-  url: new URL(process.env.ANALYTICS_CLIENT_HOST ?? DEFAULT_HOST),
-};
+export const LIBRARY_HOST = process.env.LIBRARY_CLIENT_HOST ?? CLIENT_HOST;
 
-export const CLIENT_HOSTS = [BUILDER_HOST, PLAYER_HOST, LIBRARY_HOST, ACCOUNT_HOST, ANALYTICS_HOST];
+export const ALLOWED_ORIGINS = [new URL(CLIENT_HOST).origin, new URL(LIBRARY_HOST).origin];
```
spaenleh, line 54: This can potentially fail if the CLIENT_HOST and LIBRARY_HOST are not correct URLs, should we catch and throw a better error ? I know by experience that such issues are hard to debug and find the root ...

spaenleh (approved):
Changes look good, I think the interface is bit more clear that way !

PR #1742, opened 2025-02-03
refactor: use seed for password tests

pyphilia:
- Merge two password controller tests together.
- Use `seedFromJson` to replace utility save functions
- Add `password` prop to actor for `seedFromJson`
- Add `MemberFactory`


spaenleh (changes requested):
Very nice work ! 💪 

I however think that providing a hashed password to the seed function is not the right API. I would prefer to be able to send the password for the user in clear text and the seed function hashes it internally at the level where we transform from the "nice API" to the "database level" the seed data. What do you think ? Also it would make calling the api with a password actually less strange than using await inside the call.

spaenleh (approved):

PR #1751, opened 2025-02-10
refactor: use seed in membership tests

pyphilia:
Adds membership management for `seedFromJson`.
You can specify an `account` with a name only for a membership, that will generate a member correctly from it. Account are distinct by their name, so you can reuse the same name across one seed. 
This management does not handle `guest` yet.

### DIFF
--- a/src/services/itemMembership/test/index.test.ts
+++ b/src/services/itemMembership/test/index.test.ts
@@ -6,7 +6,8 @@ import { FastifyInstance } from 'fastify';
 
 import { DiscriminatedItem, HttpMethod, PermissionLevel } from '@graasp/sdk';
 
-import build, { clearDatabase } from '../../../../test/app';
+import build, { clearDatabase, mockAuthenticate, unmockAuthenticate } from '../../../../test/app';
+import { seedFromJson } from '../../../../test/mocks/seed';
 import { resolveDependency } from '../../../di/utils';
 import { AppDataSource } from '../../../plugins/datasource';
 import { MailerService } from '../../../plugins/mailer/service';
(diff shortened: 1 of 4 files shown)


src/services/itemMembership/test/index.test.ts
```diff
@@ -63,16 +69,17 @@ describe('Membership routes tests', () => {
...
-        const member = await saveMember();
-        const membership = await testUtils.saveMembership({ item, account: member });
-
-        const memberships = [itemMembership, membership];
+        const {
+          actor,
+          items: [item],
+          itemMemberships,
+        } = await seedFromJson({
+          items: [{ memberships: [{ account: 'actor' }, { account: { name: 'bob' } }] }],
```
spaenleh, line 78: Would it be possible to have 
```suggestion
          items: [{ memberships: [{ account: 'actor' }, { account: 'bob' }] }],
```

or even: 
```suggestion
          items: [{ memberships: [ 'actor', 'bob' ] }],
```

As it would be more concise.

test/factories/member.factory.ts
```diff
@@ -19,14 +19,22 @@ function BaseAccountFactory<T extends AccountType>(baseAccount: Partial<Account>
...
+      { lang: faker.helpers.arrayElement(['en', 'fr', 'de']) },
+      {},
+    ]),
+    enableSaveActions: m.enableSaveActions ?? true,
+    isValidated,
+    lastAuthenticatedAt: isValidated ? (m.lastAuthenticatedAt ?? faker.date.anytime()) : null,
+    userAgreementsDate: m.userAgreementsDate ?? faker.date.anytime(),
+    ...BaseAccountFactory({ type: AccountType.Individual }),
+    ...m,
+  } as Member;
```
spaenleh, line 36: Why is the cast necessary ? maybe add a comment so we know when we can remove it.

spaenleh (approved):
Very nice, maybe you can ignore my suggestion as It makes it harder to provide the permission and only simplifies the expression when specifying the name.

PR #1757, opened 2025-02-14
refactor: refactor file service for app data

pyphilia:
I moved from 
```
.
├── item/plugins/app/appData         
│   ├── index.ts   
│   ├── schemas.ts   
│   ├── service.ts   
│   ├── plugins/file
│       ├── index.ts
│       ├── service.ts
├── ...
```
to 
```
.
├── item/plugins/app/appData         
│   ├── index.ts   
│   ├── schemas.ts   
│   ├── service.ts    
│   ├── appDataFileServiceAdapter.ts    
├── ...
```

This change is motivated by removing hooks. 

The original idea was to have a separated plugin, so that it is easy to remove if we don't need it.
We now take the assumption file for app data is a core feature, so I merged both plugins.
`AppDataService` now __has__ a `AppDataFileServiceAdapter`. This service is in charge of reusing a `FileService` for the purpose of app-data. 

ref #1635


src/services/item/plugins/app/appData/appData.controller.ts
```diff
@@ -104,6 +106,63 @@ const appDataPlugin: FastifyPluginAsyncTypebox = async (fastify) => {
         return appData.map(addMemberInAppData);
       },
     );
+
+    fastify.route({
+      method: HttpMethod.Post,
+      url: '/app-data/upload',
+      schema: upload,
```
spaenleh, line 113: Use `fastify.post` shorthand ?

src/services/item/plugins/app/appData/appData.controller.ts
```diff
@@ -104,6 +106,63 @@ const appDataPlugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+            // necessary to get file size -> can use stream busboy only otherwise
+            // only one file is uploaded
+            const file = await request.file();
+            if (!file) {
+              throw new UploadEmptyFileError();
+            }
+            return addMemberInAppData(await appDataService.upload(tx, member, file, app.item));
+          })
+          .catch((e) => {
+            console.error(e);
```
spaenleh, line 132: Use logger instance ?

src/services/item/plugins/app/appData/appData.controller.ts
```diff
@@ -104,6 +106,63 @@ const appDataPlugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+      async (request) => {
+        const {
+          user,
+          params: { id: appDataId },
+        } = request;
+        const member = asDefined(user?.account);
+        const app = asDefined(user?.app);
+
+        return appDataService.download(db, member, { item: app.item, appDataId }).catch((e) => {
+          if (e.code) {
```
spaenleh, line 159: No log ?

src/services/item/plugins/app/appData/appData.schemas.ts
```diff
@@ -131,3 +131,38 @@ export const getForOne = {
     '4xx': errorSchemaRef,
   },
 } as const satisfies FastifySchema;
+
+export const upload = {
+  operationId: 'createAppDataFile',
+  tags: ['app', 'app-data', 'file'],
+  summary: 'Create app data file',
+  description: `Upload a file to create a corresponding app data. The created app data will have a type file and visibility ${AppDataVisibility.Member}. The data property will contain the file properties.`,
```
spaenleh, line 139: Is it possible / do we want to allow a file to be uploaded as a Visibility.Item ?

src/services/item/plugins/app/appData/appData.service.ts
```diff
@@ -285,4 +266,36 @@ export class AppDataService {
...
+
+  async upload(
+    dbConnection: DBConnection,
+    account: AuthenticatedUser,
+    file: MultipartFile,
+    item: ItemRaw,
+  ) {
+    const appDataValue = await this.appDataFileService.upload(account, file, item);
+
+    const { id: appDataId } = await this.appDataRepository.addOne(dbConnection, {
```
spaenleh, line 278: Do we set the default visibility here ? is it set in the SQL ?

src/services/item/plugins/app/appData/appData.service.ts
```diff
@@ -285,4 +266,36 @@ export class AppDataService {
...
+      itemId: item.id,
+      actorId: account.id,
+      appData: appDataValue,
+    });
+
+    // get relations
+    const appData = await this.appDataRepository.getOne(dbConnection, appDataId);
```
spaenleh, line 285: Why do we have to make a get afterwards, can't we return the created appData from the addOne ? This looks like legacy behaviour from TypeORM. Do we want to change it, maybe in another one.

spaenleh (approved):
Changes look good ! Thank you for this refactor !

src/services/item/plugins/app/appData/appData.service.ts
pyphilia: We need item, creator and account to be returned for compatibility, `returning` cannot give this back. Whether we really need them is another question, we can simplify afterwards.

PR #1760, opened 2025-02-17
refactor: remove reply url from get item content and thumbnail

### LINKED ISSUE #1683: Remove replyUrl for GET /avatar
Not to be changed now, but do we really need to support replying with the file as a blob ? i.e. can we remove the `replyUrl` param and always rely with the url instead ?

_Originally posted by @spaenleh in https://github.com/graasp/graasp/pull/1679#discussion_r1880097807_

### LINKED ISSUE #1758: Remove replyUrl for get thumbnails

pyphilia:
Remove `replyUrl` for
- GET items/thumbnails
- GET members/avatar
- GET item/download (item file)

We still keep the query param to prevent schema error, but it does nothing. 
The only and default behavior is to return a string which is the url of the file (s3 url).

close #1758 
close #1683

### DIFF
--- a/src/services/item/plugins/file/index.ts
+++ b/src/services/item/plugins/file/index.ts
@@ -1,5 +1,3 @@
-import { StatusCodes } from 'http-status-codes';
-
 import { fastifyMultipart } from '@fastify/multipart';
 import { FastifyPluginAsyncTypebox } from '@fastify/type-provider-typebox';
 
@@ -193,7 +191,7 @@ const basePlugin: FastifyPluginAsyncTypebox<GraaspPluginFileOptions> = async (fa
       schema: download,
       preHandler: optionalIsAuthenticated,
     },
-    async (request, reply) => {
+    async (request) => {
       const {
         user,
         params: { id: itemId },
@@ -203,7 +201,7 @@ const basePlugin: FastifyPluginAsyncTypebox<GraaspPluginFileOptions> = async (fa
         itemId,
       });
 
-      reply.status(StatusCodes.OK).send(url);
+      return url;
     },
   );
 };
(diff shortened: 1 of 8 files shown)


src/services/item/plugins/file/index.ts
```diff
@@ -195,13 +197,13 @@ const basePlugin: FastifyPluginAsyncTypebox<GraaspPluginFileOptions> = async (fa
...
         params: { id: itemId },
-        query: { replyUrl },
       } = request;
 
       const url = await fileItemService.getUrl(user?.account, buildRepositories(), {
         itemId,
       });
-      fileService.setHeaders({ url, reply, replyUrl, id: itemId });
+
+      reply.status(StatusCodes.OK).send(url);
```
spaenleh, line 206: Isn't it easier to just use `return` ?

src/services/item/plugins/file/schema.ts
```diff
@@ -39,7 +39,9 @@ export const download = {
   params: customType.StrictObject({
     id: customType.UUID(),
   }),
-  querystring: customType.StrictObject({ replyUrl: Type.Boolean({ default: false }) }),
+  querystring: customType.StrictObject({
+    replyUrl: Type.Boolean({ default: true, deprecated: true }),
```
spaenleh, line 43: Add a comment here about why it is deprecated and the plans for the future ?

src/services/member/plugins/thumbnail/index.ts
```diff
@@ -94,7 +92,7 @@ const plugin: FastifyPluginAsyncTypebox<GraaspThumbnailsOptions> = async (fastif
       if (!url) {
         reply.status(StatusCodes.NO_CONTENT);
       } else {
-        fileService.setHeaders({ reply, replyUrl, url, id: memberId });
+        reply.status(StatusCodes.OK).send(url);
```
spaenleh, line 95: Same comment here, use the simple `return url` ? Or is it a better practice to use the `send` ?

spaenleh (approved):

PR #1765, opened 2025-02-18
feat(item): add postMany and refactor post

MatveyK:


src/services/item/service.ts
```diff
@@ -108,103 +112,276 @@ export class ItemService {
...
+    // no previous item adds at the beginning
+    // else define order from given previous item id
+    let order;
+    if (!previousItemId) {
+      order = await repositories.itemRepository.getFirstOrderValue(parentItem.path);
+    } else {
+      order = await repositories.itemRepository.getNextOrderCount(parentItem.path, previousItemId);
+    }
+    for (let i = 0; i < items.length; i++) {
+      items[i] = { ...items[i], order: order++ };
```
pyphilia, line 271: We prefer to increase by `DEFAULT_ORDER` instead of 1

pyphilia, line 271: I'm wondering if the order computation shouldn't be done at the repository level?

src/services/item/service.ts
```diff
@@ -108,103 +112,276 @@ export class ItemService {
...
+
+    // rescale the item ordering, if there's more than one item
+    if (items.length > 1) {
+      repositories.itemRepository.rescaleOrder(member, parentItem);
+    }
+
+    return createdItems;
+  }
+
+  private async createItemsAndMemberships(
```
pyphilia, line 290: Add comments about this function and the other one, how it is different from the one above

src/services/item/service.ts
```diff
@@ -108,103 +112,276 @@ export class ItemService {
...
+  ) {
+    return Promise.all(
+      Object.keys(geolocations).map(async (itemPath) => {
+        const geolocation = geolocations[itemPath];
+
+        // lat and lng should exist together
+        const { lat, lng } = geolocation || {};
+        if ((lat && !lng) || (lng && !lat)) {
+          throw new PartialItemGeolocation({ lat, lng });
+        }
```
pyphilia, line 339: Should we perhaps lower this down in the geolocation repository?

src/services/item/service.ts
```diff
@@ -108,103 +112,276 @@ export class ItemService {
...
-      createdItem.settings = { hasThumbnail: true };
+  /**
+   * Upload the item thumbnails.
+   * @param thumbnails Key-value map with the item ID as key
+   * @returns Items with updated `hasThumbnail` property
+   */
+  private async uploadThumbnails(
+    member: Member,
+    repositories: Repositories,
+    thumbnails: { [key: string]: Readable | undefined },
```
pyphilia, line 353: Do we need to handle `undefined`?

src/repositories/ImmutableRepository.ts
```diff
@@ -83,6 +91,20 @@ export abstract class ImmutableRepository<T extends BaseEntity> extends Abstract
     return entity;
   }
 
+  public getMultiple(
```
pyphilia, line 94:
```suggestion
  public getMany(
```

src/services/item/repository.test.ts
```diff
@@ -729,6 +730,69 @@ describe('ItemRepository', () => {
...
   });
+  describe('postMany', () => {
+    it('post many', async () => {
+      const items: Item[] = [];
+      for (let i = 0; i < 15; i++) {
+        const newItem = itemRepository.createOne({
+          name: `item${i}`,
+          type: ItemType.FOLDER,
+          creator: actor,
+        });
```
pyphilia, line 741: We have an ItemFactory (the one in sdk!) you can use to create your item structure. So we could simplify:
```js
const items = Array.from(length: 15}, (idx) => 
ItemFactory({name: `item${idx}`, creator}))
```
---> I see creator is not handled by the factor currently, but I think you can easily update.

src/services/item/repository.test.ts
```diff
@@ -729,6 +730,69 @@ describe('ItemRepository', () => {
...
+
+      const items: Item[] = [];
+      for (let i = 0; i < 15; i++) {
+        const newItem = itemRepository.createOne({
+          name: `item${i}`,
+          type: ItemType.FOLDER,
+          creator: actor,
+        });
+        items.push(newItem);
+      }
```
pyphilia, line 773: same as above.

src/services/item/repository.test.ts
```diff
@@ -729,6 +730,69 @@ describe('ItemRepository', () => {
...
+        where: { name: In(insertedItemNames) },
+        relations: { creator: true },
+      });
+      const itemNamesInDB = itemsInDB.map((i) => i.name);
+      const itemTypesInDB = insertedItems.map((i) => i.type);
+      const itemCreatorIdsInDB = insertedItems.map((i) => i.creator?.id);
+
+      expect(itemNamesInDB.sort()).toEqual(insertedItemNames.sort());
+      expect(itemTypesInDB.sort()).toEqual(insertedItemTypes.sort());
+      expect(itemCreatorIdsInDB.sort()).toEqual(insertedItemCreatorIds.sort());
```
pyphilia, line 760: I would have an additional check on the path, that is supposed to be a root path.

src/services/item/repository.test.ts
```diff
@@ -729,6 +730,69 @@ describe('ItemRepository', () => {
...
+      });
+      const itemNamesInDB = itemsInDB.map((i) => i.name);
+      const itemTypesInDB = insertedItems.map((i) => i.type);
+      const itemCreatorIdsInDB = insertedItems.map((i) => i.creator?.id);
+      const itemPathsInDB = insertedItems.map((i) => i.path);
+
+      expect(itemNamesInDB.sort()).toEqual(insertedItemNames.sort());
+      expect(itemTypesInDB.sort()).toEqual(insertedItemTypes.sort());
+      expect(itemCreatorIdsInDB.sort()).toEqual(insertedItemCreatorIds.sort());
+      expect(itemPathsInDB.sort()).toEqual(insertedItemPaths.sort());
```
pyphilia, line 793: I would have an additional check on the paths, for the level as well as it should contain the parent's path.

src/services/item/service.test.ts
```diff
@@ -77,7 +80,162 @@ describe('Item Service', () => {
...
+        where: { id: In(newItems.map((i) => i.id)) },
+      });
+
+      expect(itemsInDB.every((i) => i.settings.hasThumbnail === true)).toBeTruthy();
+    });
+    it('Registers the geolocations', async () => {
+      const actor = await saveMember();
+      const repositories = buildRepositories();
+      const geoMock = jest.spyOn(repositories.itemGeolocationRepository, 'put');
+      const parentItem = await itemService.post(actor, repositories, {
```
pyphilia, line 220: We've implemented a `seedFromJson` function, you can use to init your data, I believe you can use it in service's tests as well. So you don't have to rely on the service itself to create your data.

pyphilia (changes requested):
Thanks for this PR! It's very clean, very understandable^^ I've left some suggestions, let me know if you have any questions!

src/services/item/service.ts
```diff
@@ -108,103 +112,276 @@ export class ItemService {
...
+    const createdItems = await this.createItems(
+      member,
+      repositories,
+      [item],
+      parentId,
+      previousItemId,
+    );
+
+    // take the first and (must be) the only item
+    const createdItem = createdItems[0];
```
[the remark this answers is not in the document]
MatveyK: Beau !

src/services/item/service.ts
```diff
@@ -108,103 +112,276 @@ export class ItemService {
...
+    const { id, path } = createdItem;
+
+    // geolocation
+    if (geolocation) {
+      await this.registerGeolocations(itemGeolocationRepository, { [path]: geolocation });
```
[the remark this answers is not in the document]
MatveyK: Fine by me.

src/services/item/service.ts
MatveyK: Indeed.

src/services/item/service.ts
MatveyK: I removed the undefined, since it is not really supposed to happen, as the map passed in the parameters must contain only the relevant thumbnails.

src/services/item/repository.test.ts
MatveyK: Awesome.

src/services/item/service.test.ts
MatveyK: Awesome.

src/services/item/service.ts
```diff
@@ -108,103 +112,276 @@ export class ItemService {
...
-      else {
-        item.order = await repositories.itemRepository.getNextOrderCount(
-          parentItem.path,
-          args.previousItemId,
-        );
+    // register geolocations
+    await this.registerGeolocations(itemGeolocationRepository, geolocations);
+
+    // upload thumbnails
+    await this.uploadThumbnails(member, repositories, thumbnails);
```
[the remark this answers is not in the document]
MatveyK: The refactor looks alright for `postMany`:
```
    const { items: inputItems, parentId } = args;

    // create items
    const itemsToInsert = inputItems.map((i) => i.item);
    const createdItems = await this.createItems(member, repositories, itemsToInsert, parentId);

    await this.saveGeolocationsAndThumbnails(member, repositories, inputItems, createdItems);

    return createdItems;
```

But I think that it looks quite ugly for the `post` function:
```
    const { item, parentId, previousItemId, geolocation, thumbnail } = args;

    // item
    // take the first and (must be) the only item
    const [createdItem] = await this.createItems(
      member,
      repositories,
      [item],
      parentId,
      previousItemId,
    );

    await this.saveGeolocationsAndThumbnails(
      member,
      repositories,
      [{ item, geolocation, thumbnail }],
      [createdItem],
    );

    return createdItem;
```

What do you think ?

src/services/item/service.ts
MatveyK: This logic seems a bit too complex to be included in the repository. I think that the repository level should really focus on the database interactions and this logic seems to be more of "business" logic to me.

src/repositories/ImmutableRepository.ts
MatveyK: Renaming this function to `getMany` creates a conflict with the already present `getMany` in the `ItemRepository`. I think that we should probably merge the two functions, but it seems to be beyond the scope of this PR.

src/services/item/service.ts
MatveyK: I've reworked the refactor a bit and now it seems quite satisfying. Let me know what you think.

src/services/item/service.ts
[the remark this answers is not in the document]
MatveyK: I've reworked the refactor a bit and now it seems quite satisfying. Let me know what you think.

pyphilia (approved):

PR #1771, opened 2025-02-23
ci: replace vars with mock values for test workflow

pyphilia:
Replace vars value in test workflow so it can be run outside the organisation. 

TODO

- [ ] remove vars from repository's variables


.github/workflows/test.yml
```diff
@@ -35,23 +35,22 @@ env:
...
-  REFRESH_TOKEN_JWT_SECRET: ${{ vars.MOCK_REFRESH_TOKEN_JWT_SECRET }}
+  REFRESH_TOKEN_JWT_SECRET: 'refresh-token-jwt-secret'
   S3_FILE_ITEM_ACCESS_KEY_ID: graasp-user
   S3_FILE_ITEM_BUCKET: graasp
   S3_FILE_ITEM_PLUGIN: true
   S3_FILE_ITEM_REGION: us-east-1
   S3_FILE_ITEM_SECRET_ACCESS_KEY: graasp-pwd
-  SECURE_SESSION_SECRET_KEY: ${{ vars.MOCK_SECURE_SESSION_SECRET_KEY }}
-  STRIPE_SECRET_KEY: ${{ secrets.STRIPE_SECRET_KEY }}
+  SECURE_SESSION_SECRET_KEY: 'secure-session-secret-key'
```
spaenleh, line 53: It looks like the key has to be `key must be 32 bytes` for fastify to allow it. So maybe we should use values that are closer to "real" values.

spaenleh (changes requested):
Thank you for taking the time to do this change ! I think the keys can not be random values like these as they need to have a specific length.

spaenleh (approved):

PR #1780, opened 2025-02-27
feat(etherpad): allow readerPermission on create, item props in update

### LINKED ISSUE #1779: Allow POST /etherpad to set `readerPermission`, allow PATCH /etherpad to update item's properties

pyphilia:
- POST /etherpad: allow to set `readerPermission`
- PATCH /etherpad/id: allow to set item's props (name, settings, lang)

close #1779

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "@fastify/type-provider-typebox": "4.1.0",
     "@fastify/websocket": "7.2.0",
     "@graasp/etherpad-api": "2.1.1",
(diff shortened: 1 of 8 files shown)


MatveyK (approved):
Looks good.

I can't get over the `EtherpadReaderPermission` naming, though. This name implies that the actor is a reader, but the permission can be `Write`. When you read it, it seems that the reader is also a writer. It's not wrong, but it's hard to parse.

Maybe `EtherpadPermission` would be a better naming ?

PR #1781, opened 2025-02-28
test: update seed for item visibility

pyphilia:
ref #995

### DIFF
--- a/src/services/item/plugins/itemVisibility/test/index.test.ts
+++ b/src/services/item/plugins/itemVisibility/test/index.test.ts
@@ -10,10 +10,9 @@ import build, {
   mockAuthenticate,
   unmockAuthenticate,
 } from '../../../../../../test/app';
+import { seedFromJson } from '../../../../../../test/mocks/seed';
 import { AppDataSource } from '../../../../../plugins/datasource';
 import { ITEMS_ROUTE_PREFIX } from '../../../../../utils/config';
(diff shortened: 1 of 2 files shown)


src/services/item/plugins/itemVisibility/test/index.test.ts
```diff
@@ -139,92 +141,111 @@ describe('Item Visibility', () => {
         expect(res.statusCode).toBe(StatusCodes.BAD_REQUEST);
       });
     });
-  });
+    // });
```
spaenleh, line 144: Remove ?

test/mocks/seed.ts
```diff
@@ -322,6 +331,30 @@ async function processMembers({
...
+    (acc, { path, isHidden, isPublic }) => {
+      if (isHidden) {
+        acc.push({ item: { path }, type: ItemVisibilityType.Hidden });
+      }
+      if (isPublic) {
+        acc.push({ item: { path }, type: ItemVisibilityType.Public });
+      }
+      return acc;
+    },
+    [] as { item: { path: string }; type: ItemVisibilityType }[],
```
spaenleh, line 345: To avoid this, you can add this type to the generic on the `reduce`. It should allow you to not have to cast and TD will be happy.

spaenleh (approved):
Looks good !

PR #1785, opened 2025-03-04
fix(file): keep extension on copy, remove content-disposition and cache control

pyphilia:
I think it never worked in the first place, but duplicating pdf would trigger a download of the file once displayed. The error only happens for pdf, probably because we use a viewer.

The solution was to remove the set content-disposition (attachment) in s3.

Alongside I removed the set cache control policy, and fixed the build path that was wrong on copy. It would result in `xxxundefined`. It doesn't impact anything tho (except the download name of the file).

### DIFF
--- a/src/services/item/plugins/file/service.ts
+++ b/src/services/item/plugins/file/service.ts
@@ -50,7 +50,7 @@ class FileItemService extends ItemService {
     this.storageService = storageService;
   }
 
-  public buildFilePath(extension?: string) {
+  public buildFilePath(extension: string = '') {
     // TODO: CHANGE ??
     const filepath = `${randomHexOf4()}/${randomHexOf4()}/${randomHexOf4()}-${Date.now()}${extension}`;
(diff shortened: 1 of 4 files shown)


MatveyK (approved):
LGTM.

src/services/item/plugins/file/service.ts
```diff
@@ -209,9 +209,8 @@ class FileItemService extends ItemService {
 
   async copyFile(member: Member, repositories: Repositories, { copy }: { original; copy }) {
     const { id, extra } = copy; // full copy with new `id`
-    const { path: originalPath, mimetype } = extra[this.fileService.fileType];
-    // filenames are not used
-    const newFilePath = this.buildFilePath();
```
spaenleh, line 214: Ach...

spaenleh (approved):
Great fix ! Happy you were able to identify the underlying issue. 
Also nice that this allowed us to remove a dependency. 💪

PR #1786, opened 2025-03-04
feat(file-upload): add the possibility to delete and upload many files

MatveyK:


src/services/caching/__mocks__/service.ts
```diff
@@ -28,4 +28,8 @@ export class CachingService implements ICachingService {
   async delete(filePath: string) {
     this.cache.delete(filePath);
   }
+
+  async deleteMany(filePaths: string[]): Promise<void> {
+    filePaths.forEach((filepath) => this.cache.delete(filepath));
+  }
```
pyphilia, line 34: Is `this.cache.delete` synchronous? Do we need it to be `async`? Or should we wait somewhere?

src/services/file/service.ts
```diff
@@ -40,30 +40,50 @@ class FileService {
...
+    data: { file: Readable; filepath: string; mimetype?: string }[],
+  ) {
+    data.forEach((fileInput) => {
+      if (!fileInput.file || !fileInput.filepath) {
+        throw new UploadFileInvalidParameterError({
+          file: fileInput.file,
+          filepath: fileInput.filepath,
+        });
+      }
+    });
```
pyphilia, line 58: I think you can use `if(data.some(...)) { throw... }`

src/services/file/service.ts
```diff
@@ -40,30 +40,50 @@ class FileService {
...
+
+    const filesToUpload: FileUpload[] = [];
+    for (let i = 0; i < data.length; i++) {
+      filesToUpload.push({
+        fileStream: sanitizedFiles[i],
+        filepath: data[i].filepath,
+        memberId: account.id,
+        mimetype: data[i].mimetype,
       });
     }
```
pyphilia, line 74: Should this be merged? I think the for-loop is like a map, so both could be done at the same time? Or do you think it is clearer to split them?

src/services/file/service.ts
```diff
@@ -119,11 +139,18 @@ class FileService {
...
-      throw new DeleteFileInvalidPathError(filepath);
+    return this.deleteMany([filepath]);
+  }
+
+  async deleteMany(filepaths: string[]) {
+    for (let i = 0; i < filepaths.length; i++) {
+      if (!filepaths[i].length) {
+        throw new DeleteFileInvalidPathError(filepaths[i]);
+      }
     }
```
pyphilia, line 150: Use `array.some(...)`

pyphilia (changes requested):
LGTM! I suggested some refactor, but the general logic looks good! 🍨

src/services/file/service.ts
MatveyK: Merged !

pyphilia (approved):

PR #1790, opened 2025-03-13
fix: return deleted child items in own recycled

pyphilia:
Quick fix/patch for the following problem: deleted nested children are not returned for GET /recycled because they don't have a direct membership. The test was working because it created a membership for the child.

An error happens because typeorm automatically adds `item_path = item.path` which is not what we want. And thus ignores children.

The proposed solution works (returns the children) but the `totalCount` is false because the filtering happens after getting items. The filtering is necessary because the *raw* sql returns duplicated "membership/item".

A true solution would be to "unique" results with sql. But since we are currently refactoring the repositories with drizzle, we are not going to change the full query now.

### DIFF
--- a/src/services/item/plugins/recycled/repository.ts
+++ b/src/services/item/plugins/recycled/repository.ts
@@ -45,14 +45,15 @@ export class RecycledItemDataRepository extends MutableRepository<RecycledItemDa
       .createQueryBuilder('im')
       // we want to join on recycled item
       .withDeleted()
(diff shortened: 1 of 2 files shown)


spaenleh (approved):
Looks good to me ! As discussed, there are ways to fix the totalCount, but we agreed that we will not invest more efforts in this since it will be changed. Also he total count is wrong only if you have duplicated memberships which should not happen that often in production (it can still happen).

PR #1791, opened 2025-03-14
fix: sql migration

spaenleh:
In this PR I fix/create the 2 migration files that are necessary to 1) migrate from typeorm to drizzle and 2) create a database from scratch

These migration files have been tested on:
- a production database dump (scenario 1)
- a clean database generated by typeorm migrations (scenario 1)
- an empty database for creation from scratch by drizzle (scenario 2)

### Clean install with empty database

You should ensure that the `src/drizzle/0000_typeorm_state.sql` migration file is not commented out.

Run the migrations with the command: 
```sh
yarn drizzle-kit migrate
```

You can optionally specify connection parameters via env variables by prepending them to the command: 

```sh
DB_HOST=localhost DB_NAME=graasp DB_USERNAME=graasper DB_PORT=5432 DB_PASSWORD=graasper yarn drizzle-kit migrate
```

### Migrating an existing typeorm database

You should ensure that the `src/drizzle/0000_typeorm_state.sql` migration file is completely commented out. Use `--` style comments, not `/* ... */` style as drizzle does not support the later.

Only the last two lines should be kept, creating a fake table and dropping it right after, so the migration has "something to do".

You can run the migration with the same command as above. 

```sh
DB_HOST=localhost DB_NAME=graasp DB_USERNAME=graasper DB_PORT=5432 DB_PASSWORD=graasper yarn drizzle-kit migrate
```


src/drizzle/0000_typeorm_state.sql
```diff
@@ -0,0 +1,447 @@
...
+CREATE INDEX "IDX_item_like_item" ON "item_like" USING btree ("item_id" uuid_ops);--> statement-breakpoint
+CREATE INDEX "IDX_item_tag_item" ON "item_tag" USING btree ("item_id" uuid_ops);
+
+-- -- -- -- -- -- -- -- -- -- -- -- -- --
+-- TYPEORM MIGRATION - Comment up to here
+-- -- -- -- -- -- -- -- -- -- -- -- -- --
+
+-- Keep this is you migrate from typeorm
+create table "test" (id text not null);--> statement-breakpoint
+drop table "test";
```
pyphilia, line 447: creating a table and dropping it right away? 
Should we have this in another sql migration?

pyphilia (approved):

src/drizzle/0000_typeorm_state.sql
```diff
@@ -0,0 +1,447 @@
+-- -- -- -- -- -- -- -- -- -- -- -- -- --
+-- TYPEORM MIGRATION
+-- If migrating from Typeorm, comment from here to the next comment
```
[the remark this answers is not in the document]
spaenleh: I guess we should discuss how we will manage the migration. I suggested to comment and un-comment this part only for the deployment to "legacy" applications, In those cases maybe it would be easier to deploy the code to the EC2 and push the migrations from there, as we will have more control. For local development, commenting and un-commenting is OK i think.

What would you suggest to do instead ? Copy, paste the migration sql into a psql prompt and then inserting the required lines into the migration tables ? Those tables do not exist yet, how will we create them ? Are we sure we will correctly create those entries ?

src/drizzle/0000_typeorm_state.sql
spaenleh: Yes this is just some dummy code for when the whole migration is commented, otherwise drizzle complains and does not register the migration file.

spaenleh:
As discussed together, the idea for deploying would be to manually run these migrations using the `drizzle-kit migrate` command from within the EC2. In this view, we will push the code to main with the first migration not commented-out. On the EC2 we will comment the irrelevant parts.

Issue #429, opened 2023-06-23
[import-export] Handle file ordering


pyphilia:

MatveyK:
I have updated the issue description. What do you think ? @pyphilia @ReidyT @spaenleh

PR #1796, opened 2025-03-18
feat(graasp-export): export in graasp format

MatveyK:


src/services/item/plugins/importExport/service.ts
```diff
@@ -329,40 +332,131 @@ export class ImportExportService {
...
+    const { item, archive, itemManifest } = args;
+
+    // assign the uuid to the exported items
+    const exportItemId = v4();
+    const itemPath = path.join(path.dirname('./'), exportItemId);
+
+    // ignore the shortcuts
+    if (isItemType(item, ItemType.SHORTCUT)) {
+      return itemManifest;
+    }
```
pyphilia, line 360: In the end we will keep them 👼

src/services/item/plugins/importExport/service.ts
```diff
@@ -33,6 +33,17 @@ import {
...
 
+export type GraaspExportItem = {
+  id: string;
+  name: string;
+  type: ItemTypeUnion;
+  description: string | null;
+  settings: ItemSettings;
+  thumbnailFilename?: string;
+  children?: GraaspExportItem[];
+  mimetype?: string;
```
pyphilia, line 44: So these are different if it's for a folder or a file right, maybe let's have some comments.

pyphilia (changes requested):

src/services/item/plugins/importExport/index.ts
```diff
@@ -132,15 +132,49 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
...
+      const item = await itemService.get(member, repositories, itemId);
+
+      // trigger download action for a collection
+      const action = {
+        item,
+        type: ActionTriggers.ItemDownload,
+        extra: { itemId: item?.id },
+      };
+      await actionService.postMany(member, repositories, request, [action]);
+
```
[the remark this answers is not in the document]
MatveyK: It is already present in the other endpoint, I've just copied it wholesale. I will remove it from the `graasp-export` endpoint as it does not seem very appropriate.

src/services/item/plugins/importExport/service.ts
MatveyK: Yes, I'll create a separate issue for the shortcuts.

MatveyK: I have updated #429 to describe how the shortcuts should behave in import/export.

pyphilia (approved):
LGTM!

PR #1805, opened 2025-04-01
feat: add full text search

pyphilia:
Adds full text search to accessible, children and map items endpoints.

- No index, it doesn't seem to use it. Probably because the join removes a lot of entries beforehand.
- Reusable search operation functions

### DIFF
--- a/src/drizzle/operations.ts
+++ b/src/drizzle/operations.ts
@@ -2,12 +2,6 @@ import { SQL, ilike } from 'drizzle-orm';
 import { PgColumn } from 'drizzle-orm/pg-core';
 import { sql } from 'drizzle-orm/sql';
 
-import { DEFAULT_LANG } from '@graasp/translations';
-
-import { isMember } from '../services/authentication';
-import { MaybeUser } from '../types';
-import { ALLOWED_SEARCH_LANGS } from '../utils/config';
-
 export function isAncestorOrSelf(column: PgColumn, path: string | PgColumn): SQL {
   return sql`${column} @> ${path}`;
 }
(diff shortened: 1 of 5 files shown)


src/drizzle/operations.ts
```diff
@@ -11,3 +17,64 @@ export function isDescendantOrSelf(column: PgColumn, parentPath: string | PgColu
...
+ * @returns
+ */
+export function keywordSearch(column: PgColumn, allKeywords: string[]) {
+  return allKeywords.map((k) => {
+    return ilike(column, `%${k}%`);
+  });
+}
+
+/**
+ * return full text search query using the actor lang. If the lang is the default one or is not defined, return undefined.
```
spaenleh, line 58: Why return `undefined` if the lang is the default one ?

src/drizzle/schema.ts
```diff
@@ -729,56 +729,15 @@ export const itemsRaw = pgTable(
...
   (table) => [
     index('IDX_bdc46717fadc2f04f3093e51fd').using(
       'btree',
       table.creatorId.asc().nullsLast().op('uuid_ops'),
     ),
-    // index('IDX_gin_item_search_document').using(
-    //   'gin',
-    //   table.searchDocument.asc().nullsLast().op('tsvector_ops'),
-    // ),
+    // --------------
```
spaenleh, line 740: Remove ?

src/services/item/item.repository.ts
```diff
@@ -56,6 +63,7 @@ import {
 } from '../../drizzle/types';
 import { IllegalArgumentException } from '../../repositories/errors';
 import { AuthenticatedUser, MaybeUser, MinimalMember } from '../../types';
+import { ALLOWED_SEARCH_LANGS } from '../../utils/config';
```
spaenleh, line 66: Remove since it is not used.

src/services/item/item.repository.ts
```diff
@@ -65,6 +73,7 @@ import {
   TooManyDescendants,
   UnexpectedError,
 } from '../../utils/errors';
+import { isMember } from '../authentication';
```
spaenleh, line 76: Remove too ?

src/services/item/item.repository.ts
```diff
@@ -281,35 +290,24 @@ export class ItemRepository {
...
+    const allKeywords = params?.keywords?.filter((s) => s && s.length);
+    if (allKeywords && allKeywords.length) {
+      const keywordsString = allKeywords.join(' ');
+
+      andConditions.push(
+        or(
+          // search without lang
+          itemFullTextSearch(items, 'simple', keywordsString),
+          // search in english
+          itemFullTextSearch(items, 'english', keywordsString),
```
spaenleh, line 303: Why is english treated as a special language that is always included ? Is this why we return undefined if the user has the default lang set ? 

I am not sure I understand the reason why english should be treated differently.

spaenleh (approved):
Changes look good, I have left a question about the treatment of `english` as a special case.

src/services/item/item.repository.ts
pyphilia: Actually we use english for default for kinda "item lang". In the case the user has german as language, but is browsing an english item, the search will only happen in german if we don't add "english". But I will refactor to take into account the item's lang instead of the english only.

PR #1806, opened 2025-04-04
feat: add build and push workflow to public ECR

spaenleh:


drizzle.config.ts
```diff
@@ -7,15 +7,16 @@ export default defineConfig({
...
+    url: process.env.DB_CONNECTION!,
+    // host: process.env.DB_HOST!,
+    // port: Number(process.env.DB_PORT!),
+    // user: process.env.DB_USERNAME!,
+    // password: process.env.DB_PASSWORD!,
+    // database: process.env.DB_NAME!,
+    // ssl:
+    //   process.env.NODE_ENV === 'production' &&
+    //   // disable SSL in local, only set the var for it to disable the use of SSL
+    //   !process.env.DB_DISABLE_SSL,
```
pyphilia, line 19: Remove?

src/drizzle/db.ts
```diff
@@ -1,19 +1,21 @@
...
-  port: Number(process.env.DB_PORT!),
-  user: process.env.DB_USERNAME!,
-  password: process.env.DB_PASSWORD!,
-  database: process.env.DB_NAME!,
+  connectionString: process.env.DB_CONNECTION,
+  // host: process.env.DB_HOST!,
+  // port: Number(process.env.DB_PORT!),
+  // user: process.env.DB_USERNAME!,
+  // password: process.env.DB_PASSWORD!,
+  // database: process.env.DB_NAME!,
```
pyphilia, line 14: Remove?

src/fastify.ts
```diff
@@ -57,7 +49,7 @@ const start = async () => {
   await registerAppPlugins(instance);
 
   try {
-    await instance.listen({ port: PORT, host: HOSTNAME });
+    await instance.listen({ port: PORT, host: '0.0.0.0' });
```
pyphilia, line 52: Are we sure we want to hardcode this?

pyphilia (approved):
LGTM, I added some suggestions about default values and comments. Thanks 🩵

.github/workflows/release-please.yml
```diff
@@ -24,6 +24,14 @@ jobs:
 
       - uses: actions/checkout@v4
 
+      - name: Build images
+        uses: ./.github/actions/build-images
+        with:
+          tag: ${{ steps.release.outputs.tag_name }}
+          aws-ecr-uri: public.ecr.aws/i3k4i6f8 # should this be a var ?
```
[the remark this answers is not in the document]
spaenleh: I don't think there is anything secret in there. But using a var would allow to change it in the future and reference it in multiple places

docker/compose.yml
```diff
@@ -0,0 +1,209 @@
...
+  meilisearch_data:
+
+x-shared_environment: &shared_environment
+  # shape of the connection string:
+  # postgres://user:password@host:port/db_name?sslmode=disable
+  DB_CONNECTION: postgres://graasper:graasper@db:5432/graasp?sslmode=disable # update with your own value
+
+services:
+  core:
+    # image: public.ecr.aws/i3k4i6f8/graasp:core-0.0.1 # local image or image from aws ECR public
```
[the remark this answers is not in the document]
spaenleh: I am not sure why we should not put it here ? This is supposed to be a compose file that people can use to run Graasp no ? Do you mean that we should not advertise the public ECR ? Are we expecting interested users to build the image from source ?

src/fastify.ts
spaenleh: I can add another variable that controls the host address that the server listens to. But I think in general this should be fine to use in most situations. Instead I will set it as the default.

PR #1809, opened 2025-04-15
feat: add image build and release image to ECR

### LINKED ISSUE #1808: [CI] remove second postgres container once all tests pass
In the `test.yml` workflow, inside the `test-node` job, we declare 2 database services, with the same credentials, the second being only different on the port allocation (`5433` instead of `5432`).

## Reason

The second DB is there because in CI we run a maximum of 2 workers with jest to speed up the test execution.
Since we used to clear the DB after nearly all tests, we needed to have a database per worker, so they would not interfere.

## Proposed change

Since we are rewriting the tests to not rely on flushing the database content, we should be able to rely on a single database with concurrent test runs. 

We should probably wait until the tests are fully functional again before removing the second database from the workflow.

spaenleh:
In this PR:
- use `localhost` for the `HOSTNAME` and add `HOST_LISTEN_ADDRESS` to specify an address to listen to (default `0.0.0.0`)
- add a version endpoint at `/version` that returns the version of the backend (given inside the image)
- use a `DB_CONNECTION` string instead of the 5 env vars
- add a build `core` and `migrate` action so we can push the images to ECR on release
- remove `DATABASE_LOGS` env var
- remove `TOKEN_BASED_AUTH` env var
- fix #1808 remove second postgres DB from test job 
- update backend image to follow best practices and reduce size
- add a `compose.yml` file to run graasp for self hosting
- update deps


.github/actions/build-images/action.yml
```diff
@@ -0,0 +1,26 @@
+name: Build backend and migration image
```
pyphilia, line 1:
```suggestion
name: Build backend and migration images
```

docker/compose.yml
```diff
@@ -0,0 +1,209 @@
...
+      CORS_ORIGIN_REGEX: '^http?:\/\/(localhost)?:[0-9]{4}$'
+
+      FILE_STORAGE_ROOT_PATH: /tmp
+      FILE_STORAGE_HOST: http://localhost:1081
+
+      # endpoint of the nudenet model
+      IMAGE_CLASSIFIER_API: http://nudenet:8080/sync
+
+      # replace all these with fresh values !
+      # to Generate values use:
```
pyphilia, line 45:
```suggestion
      # to generate values use:
```

package.json
```diff
@@ -96,84 +95,82 @@
...
     "ts-node": "10.9.2",
     "tsx": "4.19.3",
-    "typescript": "5.5.4",
+    "typescript": "5.8.3",
     "wait-for-expect": "3.0.2"
   },
   "resolutions": {
     "sodium-native": "4.2.0"
-  },
-  "packageManager": "yarn@4.5.3"
```
pyphilia, line 178: Set this back?

pyphilia (approved):
LGTM!

PR #1822, opened 2025-04-23
fix: fix action export and improve download name

pyphilia:
Because the dates changed, `archive.timestamp` resolves in a different value as `requestExport.createdAt`. A `Z` was missing. I solved it by first creating a request export, and assign it's date to the archive data.

I used this opportunity to improve the download name of the file.


spaenleh (approved):
LGTM !

PR #1828, opened 2025-04-23
fix(export-actions): use timestamp for filenames, infer zip

### LINKED ISSUE #1823: File names in the download user data have a strange format
The filenames used inside the ZIP created when users download their data has filenames like:
`actions_2025-04-23T07/55/45.342Z.json`

It is not legal to use `:` inside filenames but the `/` look really out of place here too.

Maybe we could use a different format for this ?

pyphilia:
Remove strange slashes in the filename by using timestamp of the date.

Adds '.zip' format to download name.

fix #1823


spaenleh (approved):
LGTM !

PR #1829, opened 2025-04-23
fix: date utc format

spaenleh:


spaenleh:
Closing as we will implement timestamp with timezones instead.

PR #1830, opened 2025-04-24
fix: convert timestamp columns to include timezone

spaenleh:
In this PR, I alter the type of the timestamp columns for `item`, `account` and `membership_request` to include the timezone. This solves and issue in the serialization where the timestamp was not timezone aware and the frontend would interpret it as local time leading to wrong dates.

We can change other tables as well, and in the end it would be better to have all tables using `timestamp with time zone` for consistency.


pyphilia (approved):

spaenleh (commented):
Good !

PR #1840, opened 2025-04-25
feat: unify file type

pyphilia:
- I removed `s3File` to `file` only. This simplifies the handling of files. 
- There is a migration to move the `type` and `extra` properties correctly. Since we still use `file`, there's no changes required for the frontend. It concerns items, appSettings and appData.
- I introduced a `FileStorage` type that handles the server type. 
- I changed the legacy `S3_FILE_ITEM_PLUGIN=false` to use a more specific variable `FILE_STORAGE_TYPE=s3`

 
TODO:
- [ ] test on dev
- [ ] update deploy vars
- [ ] Verify no app is using the `data` of an appdata/appsetting file. But it shouldn't...

WAIT FOR:
- mobile does not handle "file" but only "s3File": https://github.com/graasp/graasp-mobile/blob/3bb87cf249bea3ae2b738c99e7ac199abc933636/src/navigation/useNavigateToPlayer.tsx#L42 https://github.com/graasp/graasp-mobile/blob/3bb87cf249bea3ae2b738c99e7ac199abc933636/src/screens/ItemScreen.tsx#L73


src/services/file/types.ts
```diff
@@ -1 +1,10 @@
+import { UnionOfConst } from '@graasp/sdk';
+
 export type BuildFilePathFunction = (itemId: string, filename: string) => string;
+
+export const FileStorage = {
+  Local: 'local',
+  S3: 's3',
+};
+Object.freeze(FileStorage);
```
spaenleh, line 9: Interesting. I read that the difference between `as const` and `Object.freeze()` is that `as const` is compile time and `Object.freeze` is runtime. Usually I use:
```ts
export const FileStorage = {
  Local: 'local',
  S3: 's3',
} as const;
```
But you are right that `Object.freeze` also ensures that at runtime nothing modifies the object, whereas `as const` does not give such guaranties. But it comes with some overhead.

src/utils/config.ts
```diff
@@ -188,11 +189,18 @@ export const MAILER_CONFIG_PASSWORD = process.env.MAILER_CONFIG_PASSWORD;
...
+/**
+ * GRAASP FILE STORAGE CONFIG
+ */
 // TODO: should this be here?
 export const FILE_STORAGE_ROOT_PATH = process.env.FILE_STORAGE_ROOT_PATH || process.env.TMPDIR;
 export const FILE_STORAGE_HOST = process.env.FILE_STORAGE_HOST;
 
+export const FILE_STORAGE_TYPE = process.env.FILE_STORAGE_TYPE ?? FileStorage.Local;
+if (!Object.values(FileStorage).includes(FILE_STORAGE_TYPE)) {
+  throw new Error(`File Storage type is not handled: '${FILE_STORAGE_TYPE}'`);
```
spaenleh, line 201:
```suggestion
  throw new Error(`File Storage type is not handled: '${FILE_STORAGE_TYPE}'. Should be one of: ${Object.values(FileStorage)}`);
```

src/utils/config.ts
```diff
@@ -223,9 +231,13 @@ const getS3FilePluginConfig = () => {
   };
 };
 // the varaible is undefined when `S3_FILE_ITEM_PLUGIN` is false
```
spaenleh, line 233: Update comment ?

spaenleh (approved):
Looks good ! thank you !

PR #1844, opened 2025-04-29
fix(dev-container): fix the init.sql file volume

MatveyK:


.devcontainer/docker-compose.yml
```diff
@@ -61,7 +61,7 @@ services:
       - 5432:5432
     volumes:
       - postgres-data:/var/lib/postgresql/data
-      - ${PWD}/postgresql/init.sql:/docker-entrypoint-initdb.d/init.sql
+      - ./postgresql:/docker-entrypoint-initdb.d
```
spaenleh, line 64: So you can't mount a single file, you need to mount the whole folder as a volume ?

spaenleh (approved):
Thank you for the fix !

Issue #1682, opened 2024-12-11
Move CORS


pyphilia:
Ah so this is why ... could we move the cors higher up and not include the apps there ? or are the apps too dependent on that ? Maybe we need to shift a few routes. Boundaries seem a bit lacking in this context.

_Originally posted by @spaenleh in https://github.com/graasp/graasp/pull/1665#discussion_r1880113246_

pyphilia:
duplicate of https://github.com/graasp/graasp/issues/1358

PR #1845, opened 2025-04-29
feat: add get maintenance endpoint

### LINKED ISSUE #1843: Create endpoint for maintenance information
Should it be a table with from-end dates, pulled from the db for the frontend to show the value?
The values would be stackable from the "admin panel".

pyphilia:
close #1843

### DIFF
--- a/src/drizzle/schema.ts
+++ b/src/drizzle/schema.ts
@@ -1066,3 +1066,13 @@ export const itemTagsTable = pgTable(
     unique('UQ_item_tag').on(table.tagId, table.itemId),
   ],
 );
+
+export const maintenanceTable = pgTable(
+  'maintenance',
+  {
+    slug: varchar({ length: 100 }).notNull(),
+    startAt: timestamp('start_at', { withTimezone: true, mode: 'string' }).notNull(),
+    endAt: timestamp('end_at', { withTimezone: true, mode: 'string' }).notNull(),
+  },
+  (table) => [unique('maintenance_slug_key').on(table.slug)],
+);
--- a/src/drizzle/schema.ts
+++ b/src/drizzle/schema.ts
@@ -1067,12 +1067,8 @@ export const itemTagsTable = pgTable(
   ],
 );
 
-export const maintenanceTable = pgTable(
-  'maintenance',
-  {
-    slug: varchar({ length: 100 }).notNull(),
-    startAt: timestamp('start_at', { withTimezone: true, mode: 'string' }).notNull(),
-    endAt: timestamp('end_at', { withTimezone: true, mode: 'string' }).notNull(),
-  },
-  (table) => [unique('maintenance_slug_key').on(table.slug)],
-);
+export const maintenanceTable = pgTable('maintenance', {
+  slug: varchar({ length: 100 }).notNull().primaryKey().unique('UQ_maintenance_slug'),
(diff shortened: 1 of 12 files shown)


src/drizzle/schema.ts
```diff
@@ -1066,3 +1066,13 @@ export const itemTagsTable = pgTable(
     unique('UQ_item_tag').on(table.tagId, table.itemId),
   ],
 );
+
+export const maintenanceTable = pgTable(
+  'maintenance',
+  {
```
spaenleh, line 1072: Should this table have a primary key even if we do not really need it ?
Something like a generated always identity https://orm.drizzle.team/docs/latest-releases/drizzle-orm-v0320#-postgresql-identity-columns ?

src/drizzle/schema.ts
```diff
@@ -1066,3 +1066,13 @@ export const itemTagsTable = pgTable(
...
+    slug: varchar({ length: 100 }).notNull(),
+    startAt: timestamp('start_at', { withTimezone: true, mode: 'string' }).notNull(),
+    endAt: timestamp('end_at', { withTimezone: true, mode: 'string' }).notNull(),
+  },
+  (table) => [unique('maintenance_slug_key').on(table.slug)],
```
spaenleh, line 1077: Any specific reason why you added the constraint as a table constraint instead of directly on the column definition ? 
I think the following is valid:
```ts
slug: varchar({ length: 100 }).notNull().unique(),
```

src/services/maintenance/maintenance.schemas.ts
```diff
@@ -0,0 +1,36 @@
...
+import { errorSchemaRef } from '../../schemas/global';
+
+const maintenanceSchema = customType.StrictObject(
+  {
+    slug: Type.String(),
+    startAt: customType.DateTime(),
+    endAt: customType.DateTime(),
+  },
+  { description: 'Entry of a maintenance period, usually to perform a migration.' },
+);
```
spaenleh, line 16: Should this be inlined in the ref definition ?

spaenleh (changes requested):
I have left a few comments. Great work, I have tried to give everything a good look. Let me know ⛑️

src/drizzle/schema.ts
pyphilia: I see, I didn't know. I did the same as for all the others tables. I changed it.

src/services/maintenance/maintenance.controller.ts
```diff
@@ -0,0 +1,36 @@
...
+import { MaintenanceService } from './maintenance.service';
+
+export const maintenancePlugin: FastifyPluginAsyncTypebox = async (fastify) => {
+  const maintenanceService = resolveDependency(MaintenanceService);
+  fastify.register(
+    async (fastify) => {
+      // add CORS support
+      if (fastify.corsPluginOptions) {
+        await fastify.register(fastifyCors, fastify.corsPluginOptions);
+      }
```
spaenleh, line 18: We should really address #1682

spaenleh (approved):
Thank you for the changes !

src/drizzle/schema.ts
spaenleh: No worries, I read it in the docs, but I think it is equivalent. You have to use the table version If you want to devine a unique constraint on multiple columns. But for single columns I feel using it inline on the column def is easier.

Issue #1358, opened 2024-08-28
Refactoring CORS


Defferrard:
We should register Cors at only 2 places
- For the base of the API
- For the App Plugin

PR #1856, opened 2025-05-01
fix: allow cors for patch, put and delete

pyphilia:
Allow back patch, put, delete and options in cors.

Given #1358  it would be better to not decorate the options. For simplicity I just added the necessary option.


pyphilia:
Because of one library that is in esm, the build for jest was failing. I downgraded ts-jest. Another solution would be to use `isolatedModules: true`.

spaenleh (approved):
Very nice ! Indeed we could investigate using isolated modules. But I think it requires some refactoring of incorrect code.

PR #1858, opened 2025-05-02
refactor: get item from authorized service

pyphilia:
This PR is still a draft, but the main idea is explained below:

The current architecture is as follow: A basic item service have been created to diminish the cycling dependency between some services and item service. This works because most of the services only just need to get an item. That's why basic item service has only getters. 
However some services still use `authorizationService` because they don't always need the item itself, but rather the permission check.

A 

```mermaid
classDiagram
direction LR
    class ItemService {
[anyService] 
	    basicItemService
    }

    class AnyService {
	    [authorizationService]
	    [basicItemService]
	    [itemService]
    }

    class BasicItemService {
	    itemRepository
	    authorizationService
	    get()
	    getMany()
    }

    class AuthorizationService {
        validatePermission()
    }

    ItemService <|-- AnyService
    AuthorizationService <|-- AnyService
    ItemService --|> AnyService
    BasicItemService <|-- AnyService
    AuthorizationService <|-- AnyService
    BasicItemService <|-- ItemService
    AuthorizationService <|-- BasicItemService

```

Instead I suggest to change the `AuthorizationService` into a `AuthorizedItemService`, that will be used to get an item (and it's related data - to be confirmed) and check it's permission. This will make all service use the AuthorizedItemService directly, instead of two services.

```mermaid
classDiagram
direction LR
    class ItemService {
	    authorizedItemService
        [anyService]
	    basicItemService.get()
    }

    class AnyService {
	    authorizedItemService
	    itemService.hooks()
    }

    class BasicItemService {
	    itemRepository
	    authorizedItemService
	    get()
	    getMany()
    }

    class AuthorizedItemService {
	    itemRepository
        getItem()
        getManyItems()
        getItemById()
        getManyItemsById()
    }

    ItemService <|-- AnyService
    ItemService --|> AnyService 
    ItemService --|> AuthorizedItemService 
    AuthorizedItemService <|-- AnyService
    AuthorizedItemService <|-- BasicItemService


```

item service is used for:
- setting hooks, we want to remove that
- itemActions: get descendants (we could use the repository)
- by extended services (childService extends itemService) to reuse the logic (post, patch, copy, delete)
- rescaleOrderForParent
- get packed. This could be a separated service.
- import/export to post/create items and get children
- invitationService: copy a template and patch its name
- item thumbnail: patch settings (hasThumbnail) - this is the main infinite loop


src/services/authorizedItem.service.getItemWithProperties.test.ts
```diff
@@ -0,0 +1,1516 @@
...
+
+      // any other member shouldn't access
+      await expect(
+        authorizationService.getItemWithProperties(MOCK_DB, {
+          permission: PermissionLevel.Admin,
+          actor: OTHER_MEMBER,
+          item: ITEM,
+        }),
+      ).rejects.toBeInstanceOf(MemberCannotAccess);
+    });
```
spaenleh, line 133: Can these 3 tests be parameterized, so we do not duplicate the logic ? Something like 
`it.each([Permission.Read, Permission.Write, Permission.Admin])(... same code, but use the input value)`

src/services/authorizedItem.service.getItemWithProperties.test.ts
spaenleh, line 1: Would it be possible, using parameterized tests to reduce the complexity of this file ? The idea would be to define the input parameters and the expected output for the cases and then have a smaller function test body making it easier to read and understand. It looks like a lot of the tests are duplicated with only the permission changed. What do you think ?

src/services/authorizedItem.service.ts
```diff
@@ -136,26 +136,141 @@ export class AuthorizationService {
...
       return true;
-    } catch (err: unknown) {
+    } catch (_e) {
       return false;
     }
   }
 
-  public async validatePermission(
+  /**
+   * Returns whether the actor can access an item.
```
spaenleh, line 175: It does not look like it returns anything. Update phrasing to make it clear that the function passes if the actor has rights, if not, it throws.

src/services/authorizedItem.service.ts
spaenleh, line 1: In this file there are functions that are public `assertAccess` that just call other public functions. Why would I use one instead of the other ?

src/services/chat/chatMessage.controller.ts
spaenleh, line 1: Fix unused imports.

src/services/chat/chatMessage.service.ts
```diff
@@ -4,30 +4,30 @@ import { PermissionLevel } from '@graasp/sdk';
...
 import { ChatMessageRepository } from './chatMessage.repository';
 import { ChatMessageNotFound, MemberCannotDeleteMessage, MemberCannotEditMessage } from './errors';
 import { MentionService } from './plugins/mentions/chatMention.service';
 
 @singleton()
 export class ChatMessageService {
   private readonly mentionService: MentionService;
-  private readonly basicItemService: BasicItemService;
   private readonly chatMessageRepository: ChatMessageRepository;
+  authorizedItemService: AuthorizedItemService;
```
spaenleh, line 16: Should it be `private readonly` too ?

src/services/item/plugins/action/requestExport/service.ts
```diff
@@ -22,36 +22,33 @@ import {
...
     fileService: FileService,
     mailerService: MailerService,
     memberService: MemberService,
     actionRequestExportRepository: ActionRequestExportRepository,
   ) {
     this.itemActionService = itemActionService;
-    this.basicItemService = basicItemService;
-    this.authorizationService = authorizationService;
+    this.authorizedItemService = authorizedItemService;
+    this.authorizedItemService = authorizedItemService;
```
spaenleh, line 51: Remove one line ?

spaenleh (changes requested):
The function `getPropertiesForItems` and `getItemWithProperties` are confusing in the naming. I did not notice the difference is one gets a single item, the other gets multiple items... I was focusing on the fact that it looked like the names were just twisted for no reason.

I like that it reduces the usage of services to mostly the `authorizedItemService`

src/services/authorizedItem.service.ts
pyphilia: `assertAccess` makes clear you only want to check the permission and does not need any return values. While the other `public` functions return some necessary values. It is to split the usages of the functions, so it's easier to continue the refactor. For example, `getPropertiesForItem` is only used twice outside of its own service, so it means we can refactor to prevent its use to make it private. Or on the other hand we can optimize `assertAccess` with let's say only one db call without caring about return values.
TLDR they can diverge at some time.

pyphilia:
Initially, the reason for having `getPropertiesForItems` and `getItemWithProperties` a bit different is that `getPropertiesForItems` does not return items, while `getItemWithProperties` does. 
But it looks possible to have `getPropertiesForItem` instead, so I'll adjust.

src/services/authorizedItem.service.getPropertiesForItem.test.ts
spaenleh, line 1: I see you applied my comment. 
Maybe we can find a sweet-spot that is in between the previous test structure and the current one.

I would say we should group tests together when they present some sort of parameterizable behaviour.
So if given some initial conditions I get the same outcome for a set of parameter values (A, B and C) I should create the test function for those initial conditions and expectations and plug in the A, B and C parameters. 

I would recommend against making complex branching inside the test function body. 
I would ensure that the inputs of the `it.each` have the same shape. If they do not have the same shape, it may mean that they should not be part of the same test.

The test inputs as well as the test body should read very naturally. If it does not read naturally, we should refactor it to make it easier to read, eventually splitting or grouping differently.

src/services/authorizedItem.service.getPropertiesForItem.test.ts
```diff
@@ -0,0 +1,443 @@
...
+            throw new Error('should throw');
+          }
+          expect(result).toEqual(returns);
+        })
+        .catch((e) => {
+          if (returns) {
+            throw new Error('should return');
+          }
+          expect(e).toBeInstanceOf(rejects);
+        });
```
spaenleh, line 150: I admit that this is a clever way to do it, but I do not think that it is easy to understand or update in case we need to change. Also I am not a fan to throwing errors inside tests in order to make the test fail ? that seems counter-productive.

src/services/authorizedItem.service.getPropertiesForItem.test.ts
```diff
@@ -0,0 +1,500 @@
...
+
+      await expect(
+        authorizationService.getPropertiesForItem(MOCK_DB, {
+          permission,
+          actor: MEMBER,
+          item: ITEM,
+        }),
+      ).rejects.toBeInstanceOf(rejects);
+    });
+  });
```
spaenleh, line 80:
```suggestion
    it.each([
      { permission: PermissionLevel.Read },
      { permission: PermissionLevel.Write },
      { permission: PermissionLevel.Admin },
    ])('request $permission should throw', async ({ permission }) => {
      jest.spyOn(itemVisibilityRepository, 'getByItemPath').mockResolvedValue([]);
      jest.spyOn(itemMembershipRepository, 'getInherited').mockImplementation(async () => null);

      await expect(
        authorizationService.getPropertiesForItem(MOCK_DB, {
          permission,
          actor: MEMBER,
          item: ITEM,
        }),
      ).rejects.toBeInstanceOf(MemberCannotAccess);
    });
  });
```

spaenleh: Seeing the rest of the tests I understand that having the rejects allows to expect different errors based on the permission requested. Question is: Should they be different errors, or should we simply add the requested permission in a generic `Requested permission not allowed` error ?

spaenleh (approved):
Nice refactor !

I am satisfied with the current implementation of the tests. I think it is a real step-up from the previous implementation. Great Job 💪 !

PR #1864, opened 2025-05-05
fix: allow to post twice a like on an item

### LINKED ISSUE #1863: POST item like if already liked throws a 500 Internal on duplicate key
We could simply update the createdAt date if this happens.

spaenleh:
In this PR:
- fix #1863 
- add return data to the `POST item like` 
- allow to post twice 
- add a test that ensures this is allowed

### DIFF
--- a/src/services/item/plugins/itemLike/itemLike.controller.test.ts
+++ b/src/services/item/plugins/itemLike/itemLike.controller.test.ts
@@ -325,6 +325,37 @@ describe('Item Like', () => {
         expect(savedLikes[1].creatorId).toEqual(actor.id);
       });
 
+      it('Allows to override like if already exist', async () => {
+        const {
+          actor,
+          items: [item],
+        } = await seedFromJson({
+          items: [
+            {
+              likes: ['actor'],
+              memberships: [{ account: 'actor', permission: PermissionLevel.Read }],
+            },
+          ],
+        });
+        assertIsDefined(actor);
+        assertIsMember(actor);
+        mockAuthenticate(actor);
+
+        const res = await app.inject({
+          method: HttpMethod.Post,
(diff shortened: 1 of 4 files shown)


src/services/item/plugins/itemLike/itemLike.schemas.ts
```diff
@@ -60,6 +60,22 @@ export const getLikesForItem = {
...
+  'Raw Item Like',
+  customType.StrictObject(
+    {
+      id: customType.UUID(),
+      itemId: customType.UUID(),
+      creatorId: customType.UUID(),
+      createdAt: customType.DateTime(),
+    },
+    {
+      description: 'Raw Like entry created when a member likes a published collection.',
```
pyphilia, line 74: Should we mention "published"? I think the backend does not prevent the like if the item is not published.

src/services/item/plugins/itemLike/itemLike.controller.test.ts
```diff
@@ -325,6 +325,37 @@ describe('Item Like', () => {
...
+        });
+        expect(res.statusCode).toBe(StatusCodes.OK);
+        // the creatorId should be the same
+        expect(res.json().creatorId).toEqual(actor.id);
+
+        // check received item like
+        // since we don't have full item, deduce from saved value
+        const savedLikes = await getItemLikesByItem(item.id);
+        expect(savedLikes).toHaveLength(1);
+        expect(savedLikes[0].creatorId).toEqual(actor.id);
```
pyphilia, line 356: Do you wanna check the created date, since you override it?

pyphilia (approved):
I added suggestion for more precise test check, but overrall LGTM 🎣

PR #1866, opened 2025-05-08
fix(files-upload): return simple error if exists

pyphilia:
Instead of a `ResultOf`, the endpoint throws an error if one exists. 

This return is a bit wrong, since some files might have still been uploaded. But it is easier for the frontend to correctly show errors.

The best solution would be to return `ACCEPTED` and deal with websockets to correctly return many status codes. But it requires some refactors (websockets, progress, refactor to use `uploadMany` instead of in serie, etc)


src/services/item/plugins/file/itemFile.controller.ts
```diff
@@ -165,6 +165,7 @@ const basePlugin: FastifyPluginAsyncTypebox<GraaspPluginFileOptions> = async (fa
           } finally {
             // force close to avoid hanging
             // necessary for errors that don't read the stream
+            log.debug('close file stream');
```
spaenleh, line 168: Keep ?

spaenleh (approved):

PR #1868, opened 2025-05-08
feat(graasp-import): export and import the item thumbnails

### LINKED ISSUE #1867: Export/Import the item thumbnails in graasp export

MatveyK:
Closes #1867


src/services/item/plugins/importExport/service.ts
```diff
@@ -411,6 +424,21 @@ export class ImportExportService {
     const exportItemId = v4();
     const itemPath = path.join(path.dirname('./'), exportItemId);
 
+    // add the thumbnail to export, if present
+    let thumbnailFilename: string | undefined = undefined;
+    try {
+      const filename = `${exportItemId}-thumbnail`;
```
pyphilia, line 430: Should we factor this out so it can be sync between import and export?

pyphilia (changes requested):
LGTM! Thanks for the simplifications!
Just take a look at my suggestion ;)

src/services/item/plugins/importExport/service.ts
```diff
@@ -411,6 +420,19 @@ export class ImportExportService {
...
+    const filename = generateThumbnailFilename(exportItemId);
+    const itemThumbnailPath = path.join(path.dirname('./'), filename);
+    const thumbnailStream = await this.itemThumbnailService.getFile(dbConnection, actor, {
+      size: ThumbnailSize.Original,
+      itemId: item.id,
+    });
+    if (thumbnailStream) {
+      archive.addReadStream(thumbnailStream, itemThumbnailPath);
+      thumbnailFilename = filename;
+    }
```
pyphilia, line 434: Should this be a private fn? So maybe you can handle the try-catch of the thumbnail service directly in this service, rather than catching it in the thumbnailService. This fn could return the `thumbnailFilename` (and stream?). 

Should we have a builder of "archive" that could get a `archiveBuilder.addThumbnail`?

Just some suggestions. The builder suggestion sounds good but requires a bigger refactor.

pyphilia (approved):

src/services/item/plugins/importExport/service.ts
MatveyK: I have refactored the try/catch into a separate function.

Indeed, the builder solution could be a nicer way to handle this. We can do this later, since it would require a bigger refactor.

PR #1869, opened 2025-05-09
fix: improve schemas for item_packed.thumbnails and search hits

spaenleh:
In this PR, I update the schema for search and item_packed to amke them work better with the generation.

### DIFF
--- a/src/services/item/item.schemas.packed.ts
+++ b/src/services/item/item.schemas.packed.ts
@@ -31,7 +31,15 @@ export const packedItemSchemaRef = registerSchemaAsRef(
       permission: customType.Nullable(customType.EnumString(Object.values(PermissionLevel))),
       hidden: Type.Optional(itemVisibilitySchemaRef),
       public: Type.Optional(itemVisibilitySchemaRef),
-      thumbnails: Type.Optional(Type.Object({}, { additionalProperties: true })),
+      thumbnails: Type.Optional(
+        customType.StrictObject(
+          {
+            small: Type.String(),
+            medium: Type.String(),
+          },
+          { additionalProperties: true },
+        ),
+      ),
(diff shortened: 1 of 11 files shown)


src/services/item/item.schemas.packed.ts
```diff
@@ -31,7 +31,15 @@ export const packedItemSchemaRef = registerSchemaAsRef(
       permission: customType.Nullable(customType.EnumString(Object.values(PermissionLevel))),
       hidden: Type.Optional(itemVisibilitySchemaRef),
       public: Type.Optional(itemVisibilitySchemaRef),
-      thumbnails: Type.Optional(Type.Object({}, { additionalProperties: true })),
+      thumbnails: Type.Optional(
+        customType.StrictObject(
+          {
+            small: Type.String(),
```
pyphilia, line 37: add format `uri`?

pyphilia: what about an item that does not have thumbnails? it doesn't have the `thumbnails` prop?

src/services/item/plugins/publication/published/plugins/search/search.schemas.ts
```diff
@@ -5,20 +5,37 @@ import { FastifySchema } from 'fastify';
...
+  'searchHit',
+  'Search Hit',
+  customType.StrictObject({
+    name: Type.String(),
+    description: Type.String(),
+    content: Type.String(),
+    creator: customType.StrictObject({
+      id: customType.UUID(),
+      name: Type.String(),
+    }),
```
pyphilia, line 25: I know it was there already, but just asking: what about a `null` creator?

pyphilia (approved):

PR #1873, opened 2025-05-14
feat: move membership endpoints

pyphilia:
Move membership endpoints inside items to have the following route: `items/:id/memberships`
This allows to remove cors options.

### DIFF
--- a/src/services/itemMembership/membership.controller.ts
+++ b/src/services/itemMembership/membership.controller.ts
@@ -1,6 +1,5 @@
 import { StatusCodes } from 'http-status-codes';
 
-import { fastifyCors } from '@fastify/cors';
 import { FastifyPluginAsyncTypebox } from '@fastify/type-provider-typebox';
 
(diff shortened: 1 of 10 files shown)


src/services/itemMembership/membership.controller.ts
```diff
@@ -14,31 +13,24 @@ import { ItemMembershipService } from './membership.service';
...
 
-const ROUTES_PREFIX = '/item-memberships';
-
 const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
   const itemMembershipService = resolveDependency(ItemMembershipService);
 
   fastify.register(MembershipRequestAPI, {
-    prefix: '/items/:itemId/memberships/requests',
+    prefix: '/:itemId/memberships/requests',
   });
```
spaenleh, line 21: Should this be registered inside the membershiups plugin so it benefits from the common path prefix ?

PR #1874, opened 2025-05-16
H5P import

### LINKED ISSUE #1872: Import H5P files in the graasp archive import

MatveyK:
Fixes #1872


src/services/item/plugins/html/h5p/h5p.service.ts
```diff
@@ -129,48 +147,37 @@ export class H5PService extends HtmlService {
...
+    try {
+      const item = await this.itemService.post(dbConnection, actor, {
+        item: metadata,
+        parentId,
+        previousItemId,
+      });
+
+      if (!isItemType(item, ItemType.H5P)) {
+        throw new Error('Expected item to be H5P but it was something else');
+      }
```
pyphilia, line 173: is that really necessary?

pyphilia (approved):
LGTM! Just a question about a part of the code 🍥

src/services/item/plugins/html/h5p/h5p.service.ts
MatveyK: This was already present, and I'm not suer of its use. I will remove it in the next PR.

PR #1881, opened 2025-05-23
feat: send email for export raw

pyphilia:
For exporting folders, the zip creation is handled in the background and a download mail is sent. Meanwhile the request responds with 202 accepted. 

The export request is logged in a corresponding table. Currently it does not do anything. 

Expiration date is 7 days. Nothing cleans up after 7 days. 

Following this PR we will:
- add a queue, for the zip generation to be handled on a worker
- this queue will also receive a schedule job to delete all archives and logs corresponding to old requests.

TODO
- [ ] write tests


src/services/item/plugins/importExport/schema.ts
```diff
@@ -25,18 +25,34 @@ export const zipImport = {
 } as const satisfies FastifySchema;
 
 export const zipExport = {
+  operationId: 'exportZip',
+  tags: ['item', 'export'],
+  summary: 'Export non-folder content',
+  description: 'Export non-folder content. Return raw file for single item.',
+
```
spaenleh, line 32: Should this still be called "zip" ? I can not see when this would be a zip if we can't export folders. It should always be a single file (image, text, html). maybe for h5p it could be a `.h5p` which is a zip inside ... but that is all I think.

I would call this "download Item" to make it evident that it gives the data directly. 
"Export" has a sort of "take it out of the platform + post-processing to it I think.

src/services/item/plugins/importExport/schema.ts
```diff
@@ -25,18 +25,34 @@ export const zipImport = {
 } as const satisfies FastifySchema;
 
 export const zipExport = {
+  operationId: 'exportZip',
```
spaenleh, line 28: You should make the operationId unique.

src/services/item/plugins/importExport/itemExportRequest.service.ts
```diff
@@ -0,0 +1,172 @@
...
+    actor: MinimalAccount,
+    item: ItemRaw,
+    archive: ZipFile,
+    type: ItemRequestExportTypeOptions,
+  ) {
+    // upload zip
+    const filepath = await this.uploadZip(dbConnection, actor, item, archive, type);
+
+    // send email
+    const memberWithEmail = (await this.memberService.get(dbConnection, actor.id)).toMemberInfo();
```
spaenleh, line 78: Split this instruction into two lines. The parenthesis make it likely to swap the order of await and `.toMemberInfo()` making it error-prone.
```suggestion
    const fullMember = await this.memberService.get(dbConnection, actor.id);
    const memberWithEmail = fullMember.toMemberInfo();
```

src/services/item/plugins/importExport/itemExportRequest.service.ts
```diff
@@ -0,0 +1,172 @@
...
+    const tmpFolder = path.join(TMP_FOLDER, 'item-export');
+    // make sure tmp folder exists
+    fs.mkdirSync(tmpFolder, { recursive: true });
+    const tmpFilepath = path.join(tmpFolder, request.id);
+
+    try {
+      // save archive content to stream
+      await pipeline(archive.outputStream, fs.createWriteStream(tmpFilepath));
+      const readFile = fs.createReadStream(tmpFilepath);
+
```
spaenleh, line 113: Is it necessary to have the temporary file on disk ? Would it be possible to pipe the stream directly to s3/local ?

src/services/item/plugins/importExport/itemExportRequest.service.ts
```diff
@@ -0,0 +1,172 @@
...
+   * @returns file path where the file is saved on the server
+   */
+  private async uploadZip(
+    dbConnection,
+    actor,
+    item,
+    archive: ZipFile,
+    type: ItemRequestExportTypeOptions,
+  ) {
+    const request = await this.itemRequestExportRepository.addOne(dbConnection, {
```
spaenleh, line 97: I think about the naming:
- `itemRequestExport` reads like an action
- `itemExportRequest` reads like what you store, so like an object. 

I would prefer the later. this is the same for "password reset" (object) and "reset password" (action).
I think tables and repositories should be called after the object not the action. 
One last example would be "File Upload" (first the noun, then the verb).

src/services/item/plugins/importExport/itemExportRequest.service.ts
```diff
@@ -0,0 +1,172 @@
...
+        },
+      },
+      lang,
+    })
+      .addText(TRANSLATIONS.EXPORT_RAW_ITEM_TEXT, {
+        itemName: item.name,
+        days: EXPORT_ITEM_EXPIRATION_DAYS.toString(),
+        exportFormat: ItemRequestExportType.Raw,
+      })
+      .addButton(TRANSLATIONS.EXPORT_RAW_ITEM_BUTTON_TEXT, link)
```
spaenleh, line 165: As highlighted in some issues, we should include the link in plain text for users that may not see the call to action button inside their mail clients.

Either the `addButton` function does it automatically, or we need to call one more method on the mailbuilder to add the link in plain text.

src/services/item/plugins/importExport/itemExportRequest.repository.ts
```diff
@@ -0,0 +1,28 @@
...
+import { itemRequestExportsTable } from '../../../../drizzle/schema';
+import { ItemRequestExportRaw } from '../../../../drizzle/types';
+
+@singleton()
+export class ItemRequestExportRepository {
+  /**
+   * Create given request export and return it.
+   * @param requestExport RequestExport to create
+   */
+  async addOne(
```
spaenleh, line 13: Should this be named `createRequest` or `logRequest`, `registerExport`.

`addOne` feels generic. But maybe this is intentional and we want our repository functions to be standard ?

spaenleh (changes requested):
Thank you for this PR. 

I have left comments related to general naming, we can decide to change them or not. I see quite a lot of the names would need to be changed.

Some of the comments are a bit nitpicking. Let me know.

src/drizzle/0006_export-item-request.sql
```diff
@@ -0,0 +1,11 @@
...
+CREATE TABLE "item_request_export" (
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"member_id" uuid,
+	"item_id" uuid,
+	"type" "item_request_export_type_enum" NOT NULL
+);
+--> statement-breakpoint 
+ALTER TABLE "item_request_export" ADD CONSTRAINT "FK_item_request_export_item_id" FOREIGN KEY ("item_id") REFERENCES "public"."item"("id") ON DELETE cascade ON UPDATE cascade;--> statement-breakpoint
+ALTER TABLE "item_request_export" ADD CONSTRAINT "FK_item_request_export_member_id" FOREIGN KEY ("member_id") REFERENCES "public"."account"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
```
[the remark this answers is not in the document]
pyphilia: I changed my opinion on the naming convention. People tend to use `FK_<target_table>_<source_table>"` https://www.navicat.com/en/company/aboutus/blog/2225-a-quick-guide-to-naming-conventions-in-sql-part-2.html
I think it also makes sense to have the category notation first. 
Also it's a good practice to name the constraints, so you don't depend on the sql server and it's easier to drop (otherwise you need to list all the constraints from the db).
Therefore I kept the references separated in drizzle as well to define the name.

pyphilia:
Here are the links within the emails, should add ellipsis on the link maybe? 

Screenshot 2025-05-28 at 13 27 06


Screenshot 2025-05-28 at 12 12 27

spaenleh:
@pyphilia I think adding an ellipsis defies the purpose of putting the link in plain text. users should be able to copy the link if needed even if it is not recognized by the email client. What does the plain text variant look like ?

src/drizzle/schema.ts
```diff
@@ -811,6 +812,14 @@ export const actionRequestExportsTable = pgTable(
   ],
 );
 
+export const itemExportRequestsTable = pgTable('item_export_request', {
+  id: uuid().primaryKey().defaultRandom().notNull(),
+  createdAt: timestamp('created_at', { withTimezone: true, mode: 'string' }).defaultNow().notNull(),
+  memberId: uuid('member_id').references(() => accountsTable.id, { onDelete: 'cascade' }),
+  itemId: uuid('item_id').references(() => itemsRawTable.id, { onDelete: 'cascade' }),
```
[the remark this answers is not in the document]
pyphilia: Yes it creates the statements automatically separated, since it adds a name.

PR #1884, opened 2025-05-28
refactor: fix lint errors

pyphilia:


src/services/item/item.controller.read.test.ts
spaenleh, line 1: remove unused imports in this file ?

src/services/itemLogin/itemLogin.controller.ts
```diff
@@ -149,6 +149,7 @@ const plugin: FastifyPluginAsyncTypebox = async (fastify) => {
           await itemLoginService.delete(tx, itemId);
           reply.status(StatusCodes.NO_CONTENT);
         } catch (e: unknown) {
+          fastify.log.error(e);
```
spaenleh, line 152: You should be able to use the log object present on the handler input arguments no ?

src/services/member/member.controller.ts
```diff
@@ -49,7 +49,7 @@ const controller: FastifyPluginAsyncTypebox = async (fastify) => {
       if (user?.account) {
         const account = await accountRepository.get(db, user?.account?.id);
         // TODO: this type is wrong. Should accountRepository.get return either MemberDTO or GuestDTO (implementing AccountDTO)?
-        return account.toCurrent() as any;
+        return account.toCurrent() as never;
```
spaenleh, line 52: Not sure about this one... I don't remember what the type issue was. Anyway between `any` and `never` 🤷

pyphilia: I think a problem between ajv and having guest/member. The todo is still there, so I think it's okay.

PR #1892, opened 2025-06-04
refactor: fix export actions test

pyphilia:
Fix test for export actions to fix lint errors.
The test is not exhaustif, but it tests it creates the right number of files in the tmp folder.


spaenleh (approved):
👍

PR #1896, opened 2025-06-06
fix: prepare changes for ESM use verbatimSyntax

spaenleh:
In this PR I make changes that should help us when moving to ESM.

I enabled and then disabled the `verbatimModuleSyntax` which forces to add `type` before importing or exporting types.

The goal is to be able to switch the package.json `type: "module"` on so we can use the project in ESM.
But for that we need to have compliant ts syntax. this is a step in that direction.


src/drizzle/schema.ts
```diff
@@ -20,15 +20,20 @@ import {
...
+} from '@graasp/sdk';
 
 import { customNumeric, ltree } from './customTypes';
 
 export const actionViewEnum = pgEnum('action_view_enum', [
   'builder',
   'player',
   'library',
-  'explorer',
+  'explorer', // TODO: remove this value from the data and migrate to library
```
pyphilia, line 36: Should we extract to an issue?

src/services/item/ws/item.events.ts
```diff
@@ -1,7 +1,7 @@
-/**
- * Item websocket events are registered under these topics
- */
```
pyphilia, line 3: The comment has been removed, was it intentional?

pyphilia (approved):

src/services/item/ws/item.events.ts
spaenleh: Yes, I did not see the interest of having that comment on the import. Do you think we should keep it ?

PR #1911, opened 2025-06-16
fix: use yazl for member export data

pyphilia:
Remove archiver to use yazl and efficiently use streams.

Simplify code.


spaenleh (approved):
👍

PR #1913, opened 2025-06-18
fix: sanitize original order before copy

pyphilia:
- add `getChildrenWithCreator` opposed to `getChildren` so we don't leftjoin the creator when it is not necessary. This simplified `rescaleOrder`
- add `fixOrderForTree` which is a sql query that rescale and correctly set orders in a tree.
- apply `fixOrderForTree` before copying, ensuring the order for the copied.

### DIFF
--- a/src/services/item/item.repository.ts
+++ b/src/services/item/item.repository.ts
@@ -306,8 +306,8 @@ export class ItemRepository {
       .select()
       .from(items)
       .where(isDirectChild(items.path, parent.path))
-      // use order for ordering
-      // backup order by in case two items has same ordering
+      // use `order` column for sorting
+      // use `createdAt` column as a backup in case two items have the same `order` value
       .orderBy(() => [asc(items.order), asc(items.createdAt)]);
   }
 
@@ -319,7 +319,7 @@ export class ItemRepository {
    * @param params
    * @returns
    */
-  async getChildrenWithCreator(
+  async getFilteredChildren(
     dbConnection: DBConnection,
     actor: MaybeUser,
     parent: ItemRaw,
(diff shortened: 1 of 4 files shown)


src/services/item/item.repository.ts
```diff
@@ -292,12 +294,32 @@ export class ItemRepository {
...
+    if (parent.type !== ItemType.FOLDER) {
+      throw new ItemNotFolder({ id: parent.id });
+    }
+
+    return await dbConnection
+      .select()
+      .from(items)
+      .where(isDirectChild(items.path, parent.path))
+      // use order for ordering
+      // backup order by in case two items has same ordering
```
spaenleh, line 310:
```suggestion
      // use `createdAt` column as a backup in case two items have the same `order` value
```

src/services/item/item.repository.ts
```diff
@@ -292,12 +294,32 @@ export class ItemRepository {
...
+  /**
+   * return children of given parent, ordered by order, with creator
+   * @param db
    * @param actor
    * @param parent
    * @param params
    * @returns
    */
-  async getChildren(
+  async getChildrenWithCreator(
```
spaenleh, line 322: I think it was smart to split this function into the "simple" get all children and have this one where you can provide the search parameters. My question now is whether the search parameters should be optional or not. Maybe they should be required to indicate that this function does more than just return the children with the creator ? Also not sure it should be named this way. Maybe we can name it something like "getFilteredChildren" ? Or relate it to the fact that it is used to display a view in the front, so the shape is tailored to this ? These are just ideas, maybe the name is fine, it is just that on the principle I do not like to put the return type in the function name.

PR #1920, opened 2025-06-24
fix: action.view migration to remove `explorer` and add `home` and `analytics`

### LINKED ISSUE #1900: `actionViewEnum` references deprecated view `explorer`
We should ensure that the data in the actions table does not reference that view anymore by migrating all instances to the new value: `library`.

_Originally posted by @pyphilia in https://github.com/graasp/graasp/pull/1896#discussion_r2132410477_

spaenleh:
In this PR:

- remove the "explorer" value from the `action.view` enum ⚠️ REQUIRES A MIGRATION ⚠️ 
- update sdk (code changes make the return type of getContextFromLink have a better type and it clashes with the current implementation)

fix #1900

### DIFF
--- a/src/services/action/action.service.ts
+++ b/src/services/action/action.service.ts
@@ -9,7 +9,7 @@ import { type DBConnection } from '../../drizzle/db';
 import type { ItemRaw } from '../../drizzle/types';
 import { BaseLogger } from '../../logger';
 import { AccountType, type MaybeUser } from '../../types';
-import { View } from '../item/plugins/action/itemAction.schemas';
+import { View, ViewOptions } from '../item/plugins/action/itemAction.schemas';
 import { MemberRepository } from '../member/member.repository';
 import { ActionRepository } from './action.repository';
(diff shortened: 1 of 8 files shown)


src/services/action/action.service.ts
```diff
@@ -56,10 +56,13 @@ export class ActionService {
       return;
     }
 
-    let view = View.Unknown;
+    let view: ViewOptions = View.Unknown;
     try {
       if (headers?.origin) {
-        view = ClientManager.getInstance().getContextByLink(headers?.origin);
+        const context = ClientManager.getInstance().getContextByLink(headers?.origin);
+        if (!['analytics'].includes(context)) {
```
pyphilia, line 63: Add a comment this is a black list.

pyphilia (approved):
Thanks for the PR! I've added some minor comments.

spaenleh:
@pyphilia in light of what we discussed concerning excluding or not analytics, I chose to add it, we can always decide to not include it in the exports or in computations if required.

PR #1924, opened 2025-06-30
refactor: pass account id instead of complete account to authorization

pyphilia:
Refactor `AuthorizedItemService` to use `accountId` instead of a full `member` props.
This PR only updates one level, which means only functions from the `AuthorizedItemService` service. Some caller functions could be optimised by using "id" only, but it not in the scope of the PR.

Rename some arguments called `actor` for a more accurate name.

### DIFF
--- a/src/services/itemMembership/membership.service.ts
+++ b/src/services/itemMembership/membership.service.ts
@@ -126,17 +126,23 @@ export class ItemMembershipService {
 
   async create(
     dbConnection: DBConnection,
-    actor: AuthenticatedUser,
+    authenticatedUser: AuthenticatedUser,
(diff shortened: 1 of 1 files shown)


src/services/itemMembership/membership.service.ts
```diff
@@ -131,7 +131,7 @@ export class ItemMembershipService {
   ) {
     // check memberships
     const item = await this.authorizedItemService.getItemById(dbConnection, {
-      actor,
+      accountId: membership.memberId,
```
spaenleh, line 134: It looks like this has changed the behaviour is it intentional ?

spaenleh (changes requested):
Nice ! See my comment

spaenleh (approved):
Yay ! 🎉

PR #1928, opened 2025-06-30
feat: create build index worker

### LINKED ISSUE #1927: Create worker for building index

pyphilia:
Create a new worker to handle build index task.
Remove pre-task of fetching pdf content, as we do it on upload.

For now, it uses `MeilisearchWrapper` as dependency to index items. In the next PR, indexing item should be a job as well.

close #1927

### DIFF
--- a/src/workers/searchIndex.worker.ts
+++ b/src/workers/searchIndex.worker.ts
@@ -1,7 +1,7 @@
 import { type Job, Worker } from 'bullmq';
 
 import { REDIS_CONNECTION } from '../config/redis';
-import { BaseLogger } from '../logger';
+import { type BaseLogger } from '../logger';
 import { Queues } from './config';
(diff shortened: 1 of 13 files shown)


src/workers/searchIndex.worker.ts
```diff
@@ -0,0 +1,57 @@
...
+      },
+      {
+        connection: { url: REDIS_CONNECTION },
+        removeOnComplete: { count: 30 },
+        removeOnFail: { count: 100 },
+      },
+    );
+
+    this.worker.on('completed', (job) => {
+      this.logger.info(`${job.id} has completed!`);
```
spaenleh, line 41: Should we prefix with the Queue name, so we can more easily see them in the logs ?

spaenleh (commented):
Interesting ! I will give this a more in-depth review later, for now it looks good !

PR #1929, opened 2025-06-30
feat: add GitHub OAuth for admin

### LINKED ISSUE #1907: Implement OAuth for admin access to queue ui

spaenleh:
fix #1907

TODO: add tests


src/drizzle/0008_add-admins.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE "admins" (
```
pyphilia, line 1: should we follow the current convention of singular table name?

src/drizzle/0008_add-admins.sql
```diff
@@ -0,0 +1,6 @@
...
+	"user_name" varchar(39) PRIMARY KEY NOT NULL,
```
pyphilia, line 2: From your demo this username should match the github name. Should this be called `github_name`? Can this logic break if we use another oAuth method?

src/drizzle/meta/0007_snapshot.json
```diff
@@ -3597,4 +3598,4 @@
     "schemas": {},
     "tables": {}
   }
-}
\ No newline at end of file
+}
```
pyphilia, line 3601: Should this really change?

src/plugins/admin/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+The admins are authenticated using GitHub OAuth2 api. For how to setup an OAuth app to work with your local setup see [Local setup for OAuth App](#local-setup-for-oauth-app)
+We only store the user github ID and their username, as well as their last authentication time.
+
+To allow a user to access the administration part, they need to exist in the `admins` table.
+The `admins` table has a unique constraint on the `user_name` field.
+
+### Local setup for OAuth app
+
+For the admin feature to work in your local development environement you need to have GitHub OAuth app credentials. See the [GitHub OAuth app documentation](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/creating-an-oauth-app) for a full walkthrough of how to set one up.
```
pyphilia, line 15:
```suggestion
For the admin feature to work in your local development environment you need to have GitHub OAuth app credentials. See the [GitHub OAuth app documentation](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/creating-an-oauth-app) for a full walkthrough of how to set one up.
```

src/plugins/admin/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+1. Go to: Settings -> Developer Settings -> OAuth apps 
+1. Create an OAuth app ("New OAuth app" button)
+   1. Choose the name you want (i.e. "Admin dev" or "Local Graasp admin")
+   1. For "Homepage URL" set `http://localhost:3000` 
+   1. You can add a description (optional)
+   1. Set the "Authorization callback URL" to `http://localhost:3000/admin/auth/github/callback`
+   1. Validate
+1. Copy the "Client Id" and paste it in your `.env.development` file next to the `GITHUB_CLIENT_ID` 
+1. Generate a new client secret, copy it and paste it in your `.env.development` file next to `GITHUB_CLIENT_SECRET` (be carefull, as this value will not be shown again, you cna always re-generate it if you loose it, also DO NOT SHARE IT WITH ANYONE!)
```
pyphilia, line 27:
```suggestion
1. Generate a new client secret, copy it and paste it in your `.env.development` file next to `GITHUB_CLIENT_SECRET` (be carefull, as this value will not be shown again, you can always re-generate it if you loose it, also DO NOT SHARE IT WITH ANYONE!)
```

pyphilia (changes requested):
Thanks for this PR, let me know if we need to discuss on some points!

Is it necessary to setup the github credentials locally? Or is there an easy way to bypass this locally? If it's too hacky to avoid the setup, let's keep it like that.

src/drizzle/meta/0007_snapshot.json
spaenleh: Yes I think it was missing some values, since it generated things that were already included in the previous migration.

src/plugins/admin/admin.plugin.ts
```diff
@@ -0,0 +1,172 @@
...
+        }
+        if (!username) {
+          throw new MissingGithubUsername();
+        }
+        // only allow users that are present in the admin table by their username
+        if (await adminRepository.isAdmin(db, githubId)) {
+          console.debug('user is an allowed admin');
+          // update info stored in the table
+          await adminRepository.update(db, { githubId, githubName: username });
+          // You can add admin checks here
```
pyphilia, line 89: Remove? We did the check above?

src/plugins/admin/admin.plugin.ts
```diff
@@ -0,0 +1,172 @@
...
+        const { username, id: githubId } = profile;
+        if (!githubId) {
+          throw new MissingGithubId();
+        }
+        if (!username) {
+          throw new MissingGithubUsername();
+        }
+        // only allow users that are present in the admin table by their username
+        if (await adminRepository.isAdmin(db, githubId)) {
+          console.debug('user is an allowed admin');
```
pyphilia, line 86: Print id and username? Or only username?

src/plugins/admin/admin.plugin.ts
```diff
@@ -0,0 +1,172 @@
...
+      return admin;
+    },
+  );
+
+  fastify.get(
+    '/admin/auth/github',
+    {
+      preValidation: adminPassport.authenticate(GITHUB_OAUTH_STRATEGY, { scope: ['user:email'] }),
+    },
+    async (_request: AdminRequest, _reply) => {},
```
pyphilia, line 118: Should it do something? Show some success message?

src/plugins/admin/admin.plugin.ts
```diff
@@ -0,0 +1,172 @@
...
+  );
+
+  // login page when user is not authenticated
+  fastify.get('/admin/login', async (_request, reply) => {
+    reply.type('text/html').send('<a href="/admin/auth/github">Login with GitHub</a>');
+  });
+
+  // create a scope where if the user is not authenticated they get redirected to the login page
+  fastify.register(
+    fp(async (authenticatedAdmin) => {
```
pyphilia, line 140: Do you need `fp`?

pyphilia (changes requested):

src/services/auth/plugins/passport/preHandlers.test.ts
```diff
@@ -1,17 +1,29 @@
...
 import { assertIsMember } from '../../../authentication';
 import { validatedMemberAccountRole } from '../../../member/strategies/validatedMemberAccountRole';
 import { isAuthenticated, matchOne } from './preHandlers';
 
-// move this test closer to matchone
-// other prehandlers are tested in plugin.test.ts
+// TODO: this tests fails now because passport registration has been moved to inside the core registration and we can not register a route there,
+// all top level routes do not use passport.
+// move this test closer to matchOne
+// other preHandlers are tested in plugin.test.ts
```
pyphilia, line 26: Does it work? Remove comment?

pyphilia (changes requested):

src/plugins/admin/admin.plugin.ts
spaenleh: From what I understood, the handler here is empty because the strategy does the work. It is invoked when you click the link to the login with github. After that github will call the callback. (There is actually another ping-pong with a code that is entirely handled by the strategy).

src/plugins/admin/admin.plugin.ts
spaenleh: I do not need `fp` in this instance, because the upper scope does not need access to what will be defined in this plugin. We always get access to what is defined in the parent which is enough for this case.

PR #1938, opened 2025-07-07
docs: improve openapi data for password endpoints

pyphilia:


src/services/auth/plugins/password/password.schemas.ts
```diff
@@ -7,6 +7,7 @@ import { customType } from '../../../../plugins/typebox';
 import { errorSchemaRef } from '../../../../schemas/global';
 
 export const passwordLogin = {
+  operationId: 'signInWithPassword',
```
spaenleh, line 10: I think we want to move out of `signIn` `signUp` and instead use `login` and `register` what do you think ?

src/services/auth/plugins/password/password.schemas.ts
```diff
@@ -20,9 +21,17 @@ export const passwordLogin = {
   querystring: customType.StrictObject({
     lang: Type.Optional(Type.String()),
   }),
+  response: {
+    [StatusCodes.OK]: customType.StrictObject({
+      resource: Type.String({ description: 'Redirection link' }),
+    }),
```
spaenleh, line 27: We provide the link as text ? not with a redirect ?

src/services/auth/plugins/password/password.schemas.ts
```diff
@@ -56,6 +66,7 @@ export const updatePassword = {
 } as const satisfies FastifySchema;
 
 export const postResetPasswordRequest = {
+  operationId: 'requestResetPasswordLink',
```
spaenleh, line 69:
```suggestion
  operationId: 'requestPasswordResetLink',
```

spaenleh (approved):
Should we try to match the operation ids with the schema var names ? I think the operationIds are currently better than the schema names.

src/services/auth/plugins/password/password.schemas.ts
pyphilia: With axios, you cannot redirect with a post request. 
Perhaps we should get rid of axios, as redirect also returns as errors... but it would be a much bigger reshape.

PR #1944, opened 2025-07-17
feat(password): set login session in password login

pyphilia:
On password login, the `/auth` redirection link was returned as a string so the frontend could redirect manually.

Instead, we directly sign in the user in the `/login-password`. The redirection to some interface will happen in the frontend.

### DIFF
--- a/src/services/auth/plugins/password/password.controller.test.ts
+++ b/src/services/auth/plugins/password/password.controller.test.ts
@@ -1,5 +1,6 @@
 import { faker } from '@faker-js/faker';
 import { compare } from 'bcrypt';
+import { compareAsc } from 'date-fns';
 import { eq } from 'drizzle-orm';
 import { ReasonPhrases, StatusCodes } from 'http-status-codes';
 import { Redis } from 'ioredis';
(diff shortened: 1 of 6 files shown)


src/services/auth/plugins/password/password.controller.test.ts
```diff
@@ -94,8 +94,14 @@ describe('Password', () => {
...
-      expect(response.statusCode).toEqual(StatusCodes.SEE_OTHER);
-      expect(response.json()).toHaveProperty('resource');
+      expect(response.statusCode).toEqual(StatusCodes.NO_CONTENT);
+      expect(response.headers['set-cookie']).toBeDefined();
+
+      // last authenticated at should be updated
+      const m = await db.query.accountsTable.findFirst({
+        where: eq(accountsTable.email, actor.email!),
+      });
+      expect(m?.lastAuthenticatedAt).not.toEqual(actor.lastAuthenticatedAt);
```
spaenleh, line 104: Should we ensure the new `last_authenticated_at` is `after` the old one ? I think we can use functions from `date-fns` for this.

spaenleh:
@pyphilia I guess the tests failing are inside another file: src/services/auth/plugins/password/password.controller.test.ts
● Password › Reset Password › PATCH Reset Password Request Route › member has previously sent a request › Reset password
● Password › Flow tests › Sign in with password and use the resource to receive the token

They expect the response to be a redirect.

PR #1949, opened 2025-07-18
fix: check admin permission on recycled items for deletion

### LINKED ISSUE #1948: Permission check on recycled items throw for deletion
The `/delete` operation on recycled items does not work because the permission check only checks non-recycled items.

pyphilia:
Create a dedicated permission check for deletion. Optimize by checking for all items at the same time.

close #1948


spaenleh (approved):
Nice optimization ! I like the idea of making functions that are dediated to a purpose 👍

PR #1971, opened 2025-08-28
fix: update to support admin schemas

spaenleh:


src/drizzle/0011_phoenix-schemas.sql
```diff
@@ -0,0 +1,49 @@
...
+	"created_at" timestamp (0) NOT NULL,
+	CONSTRAINT "users_tokens_context_token_index" UNIQUE("context","token")
+);
+--> statement-breakpoint
+ALTER TABLE "removal_notices" ADD CONSTRAINT "removal_notices_user_id_users_id_fk" FOREIGN KEY ("user_id") REFERENCES "public"."users"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "removal_notices" ADD CONSTRAINT "removal_notices_creator_id_users_id_fk" FOREIGN KEY ("creator_id") REFERENCES "public"."users"("id") ON DELETE set null ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "users_tokens" ADD CONSTRAINT "users_tokens_user_id_users_id_fk" FOREIGN KEY ("user_id") REFERENCES "public"."users"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
+CREATE INDEX "removal_notices_user_id_index" ON "removal_notices" USING btree ("user_id" uuid_ops);--> statement-breakpoint
+CREATE INDEX "users_email_index" ON "users" USING btree ("email");--> statement-breakpoint
+CREATE INDEX "users_tokens_user_id_index" ON "users_tokens" USING btree ("user_id" uuid_ops);
```
pyphilia, line 43: Move those next to the tables.

src/drizzle/0012_table_name_update.sql
```diff
@@ -0,0 +1,28 @@
+ALTER TABLE "app" RENAME TO "apps";--> statement-breakpoint
+ALTER TABLE "item_published" RENAME TO "published_items";--> statement-breakpoint
+ALTER TABLE "publisher" RENAME TO "publishers";--> statement-breakpoint
+ALTER TABLE "apps" DROP CONSTRAINT "FK_37eb7baab82e11150157ec0b5a6";
+--> statement-breakpoint
+ALTER TABLE "published_items" DROP CONSTRAINT "item_published_creator_id_account_id_fk";
+--> statement-breakpoint
+ALTER TABLE "published_items" DROP CONSTRAINT "item_published_item_path_item_path_fk";
+--> statement-breakpoint
+ALTER TABLE "apps" ADD CONSTRAINT "FK_37eb7baab82e11150157ec0b5a6" FOREIGN KEY ("publisher_id") REFERENCES "public"."publishers"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
```
pyphilia, line 10: Let's take this opportunity to rename this correctly?

src/drizzle/0011_phoenix-schemas.sql
```diff
@@ -0,0 +1,49 @@
...
+	"creator_id" uuid,
+	"created_at" timestamp (0) NOT NULL
+);
+--> statement-breakpoint
+CREATE TABLE "schema_migrations" (
+	"version" bigint PRIMARY KEY NOT NULL,
+	"inserted_at" timestamp (0)
+);
+--> statement-breakpoint
+CREATE TABLE "users" (
```
pyphilia, line 17: Change to `admins`, as long as this table will only contain admins.

src/drizzle/0011_phoenix-schemas.sql
```diff
@@ -0,0 +1,49 @@
+CREATE EXTENSION IF NOT EXISTS citext;--> statement-breakpoint
+
+CREATE TABLE "removal_notices" (
+	"id" uuid PRIMARY KEY NOT NULL,
+	"publication_name" varchar(255),
+	"reason" text,
+	"user_id" uuid,
```
pyphilia, line 7: If the `users` table becomes `admins`, I think it would be still okay to keep `user_id`.

However I think it would be clearer to have `account_id` since it links to the `account` table. Having `user` and `account` might get really confusing quickly.

pyphilia (changes requested):
Thanks for the PR, let me know if you have questions! 🫏

src/drizzle/0011_phoenix-schemas.sql
```diff
@@ -0,0 +1,49 @@
...
+--> statement-breakpoint
+CREATE TABLE "schema_migrations" (
+	"version" bigint PRIMARY KEY NOT NULL,
+	"inserted_at" timestamp (0)
+);
+--> statement-breakpoint
+CREATE TABLE "users" (
+	"id" uuid PRIMARY KEY NOT NULL,
+	"email" "citext" NOT NULL,
+	"hashed_password" varchar(255),
```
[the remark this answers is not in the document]
spaenleh: This is the default when generating the auth. It is especially called `hashed_password` to remove any confusion on whether it is in clear or hashed. I have seen this used in other database systems and I think it is considered a good practice to call it "hashed_password". Usually at some point in your app you might have the password in clear and you want to ensure that you do not confuse the two.

src/drizzle/0011_phoenix-schemas.sql
spaenleh: I did not choose the order this was auto-generated but I will move them logically next to the table definitions.

src/drizzle/0011_phoenix-schemas.sql
```diff
@@ -0,0 +1,49 @@
+CREATE EXTENSION IF NOT EXISTS citext;--> statement-breakpoint
+
+CREATE TABLE "removal_notices" (
```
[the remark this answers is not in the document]
spaenleh: I have renamed the table to `publication_removal_notices`

PR #1980, opened 2025-09-08
fix: use arm build

### LINKED ISSUE #1972: [BUILD] Run images on ARM arch in the cloud
To take advantage of better pricing for ARM cloud resources we want to migrate our images to use ARM-based runners.

We can use ARM runners in github (available on public projects) and run them on Fargate ARM machines.

spaenleh:
In this PR:

- fix #1972 Use ARM runners to build arm images
- remove the staging and production deploys since this is not how we currently deploy the stuff
- upgrade node version used in tests to be 22 to match what is in `package.json`


pyphilia (approved):
LGTM!

PR #1987, opened 2025-09-10
fix: update nudenet to use latest release and control the classes that are inferred

spaenleh:


.devcontainer/docker-compose.yml
```diff
@@ -176,10 +176,10 @@ services:
   # necessary for validation
   nudenet:
     hostname: nudenet
-    image: notaitech/nudenet:classifier
+    image: public.ecr.aws/graasp/graasp:nudenet-latest
     # exposing these ports is not necessary
-    # ports:
-    #   - 8080:8080
+    ports:
+      - 8080:8080
```
spaenleh, line 182: No need to expose these ports

src/services/item/plugins/publication/validation/processes/imageClassification.ts
```diff
@@ -1,44 +1,71 @@
...
+  class: string;
+  score: number;
+  box: [number, number, number, number];
 };
+type ClassificationResponse = {
+  prediction?: ClassPrediction[][];
+  success: boolean;
+};
+
+const CLASSIFIER_LABELS = [
```
spaenleh, line 17: We can select the labels that trigger the filter.

PR #2004, opened 2025-10-14
feat: return minimal item parent

pyphilia:
Return minimal parents with only used properties.

ref #2003

### DIFF
--- a/src/services/item/item.schemas.ts
+++ b/src/services/item/item.schemas.ts
@@ -117,7 +117,7 @@ export const itemUpdateSchema = Type.Partial(
 export const getParentItems = {
   operationId: 'getParentItems',
   tags: ['item'],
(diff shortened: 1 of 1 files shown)


src/services/item/item.schemas.ts
```diff
@@ -114,6 +114,23 @@ export const itemUpdateSchema = Type.Partial(
   ),
 );
 
+export const getParentItems = {
+  operationId: 'getParentItems',
+  tags: ['item'],
+  summary: 'Get parent items of item',
```
spaenleh:
```suggestion
  summary: 'Get parents of item',
```

spaenleh (approved):
Changes look good ! So if I understand correctly the issue is resolved by not feeding so much data in the response ? If this is enough to make the load less then this is good news.

PR #2005, opened 2025-10-14
fix: optimize get item parents

### LINKED ISSUE #2003: Optimize GET /parents

pyphilia:
Optimize sql query.

TO REBASE ON https://github.com/graasp/graasp/pull/2004

close #2003

### DIFF
--- a/src/services/item/item.controller.read.test.ts
+++ b/src/services/item/item.controller.read.test.ts
@@ -1488,9 +1488,9 @@ describe('Item routes tests', () => {
       });
       it('Cannot get parents if item is public and hidden', async () => {
         const {
-          actor,
           items: [_parent, child],
         } = await seedFromJson({
+          actor: null,
           items: [
             {
               isPublic: true,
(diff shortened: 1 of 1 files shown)


src/services/item/item.repository.ts
```diff
@@ -277,6 +279,82 @@ export class ItemRepository {
...
+
+    const result = dbConnection
+      .select()
+      .from(itemTree)
+      .leftJoin(imTree, isAncestorOrSelf(imTree.itemPath, itemTree.path))
+      .leftJoin(ivTree, isAncestorOrSelf(ivTree.itemPath, itemTree.path))
+      .where(conditions)
+      .orderBy(asc(sql`nlevel(path)`));
+
+    return (await result).map(({ item_tree }) => ({
```
spaenleh: I am not a big fan of doing the await in parenthesis like this. I think for legibility it would be better to directly await the result. Maybe add a comment that tells why we await here and not in other queries in this function.

spaenleh (approved):
Very interesting ! I let you do the proper rebasing once you have merged the depending PR.

PR #2021, opened 2025-10-23
fix: update Sentry instrumentation

spaenleh:


pyphilia (changes requested):

spaenleh:
@pyphilia I made the requested changes and also made it so that we push the sourcemaps to sentry in the build process (inside docker) this is similar to how we do it for the library.

I have removed the `enableLogs` config (we will not send log data to Sentry. Not sure if this was what you meant by the comment.

pyphilia (approved):
LGTM! We finally made it :D

spaenleh:
Well this was in great part thanks to you ! Bloody filters ...

PR #2026, opened 2025-10-24
fix: make single client changes

spaenleh:
In this PR I propose changes that help with the migration to the single client:

- expose the api routes on the `/api` path prefix in the swagger documentation -> this is so that the generation can pick them up
- add the `operationId` to the member actions call, to make it stable when we change the paths exposed


spaenleh:
Closing this since we would like to really move the api to use a global prefix. I will make the changes in another PR.

PR #2037, opened 2025-11-11
fix: expose with the /api prefix

spaenleh:
In this PR:
- expose routes under the `/api` prefix
  - only the `/health` route stays at the root
- add a nginx container that appends `/api` prefix before forwarding requests to the core container

I made the following changes:

The "legacy" server is accessible at `localhost:3000` where the nginx forwards requests. Use paths without /api if using this. This mimicks the api.graasp.org behaviour. 

A new `localhost:3001` that is the core container directly. Here you need to use paths **with** the `/api` prefix. This is designed to be used as the target for single origin and as a target for the vite server proxy. 

Example usage: 

`http://localhost:3000/items/:id` -> via nginx proxy
`http://localhost:3001/api/items/:id` -> direct to backend

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,6 +8,7 @@ test
 tmp
 **/test
 **/*.test.ts
+**/*.spec.ts
 types
 docker
 API Collections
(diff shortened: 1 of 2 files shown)


pyphilia (approved):
Huge PR! A step towards ESM ;)

PR #2048, opened 2026-01-05
feat: return only deleted items up to 3 months

### LINKED ISSUE #2047: Do not show items older than 3 months for recycled items

pyphilia:
close #2047

### DIFF
--- a/src/services/item/plugins/recycled/recycled.repository.ts
+++ b/src/services/item/plugins/recycled/recycled.repository.ts
@@ -67,21 +67,24 @@ export class RecycledItemDataRepository {
       .as('ownMemberships');
 
     // get recycled items not older than 3 months
-    const threeMonths = subMonths(new Date(), 3);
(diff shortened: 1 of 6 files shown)


src/services/item/plugins/recycled/recycled.repository.ts
```diff
@@ -65,6 +66,8 @@ export class RecycledItemDataRepository {
       )
       .as('ownMemberships');
 
+    // get recycled items not older than 3 months
+    const threeMonths = subMonths(new Date(), 3);
```
spaenleh: Should we use a more descriptive variable name that does not include the value ?

```suggestion
    const autoDeletionDeadline = subMonths(new Date(), 3);
```

spaenleh (changes requested):
Nice PR ! 

Does it require an index to work properly ?

pyphilia:
Nice catches! I added an index. It doesn't change much the speed tho, probably because the recycled item table isn't so big. I still added it in case it grows in the future.

PR #2052, opened 2026-01-13
feat: allow switching from folder to capsule

pyphilia:
Add new endpoints to allow switching from capsule to folder and vice-versa:
- PATCH /api/items/folders/:id/to-capsule
- PATCH /api/items/capsules/:id/to-folder

And an endpoint to create a capsule
- POST /api/items/capsules

Capsules being basically a folder, a lot of the logic is similar or reused. I've used some proxy variables in case we want it to be a different item type one day.

### DIFF
--- a/src/services/item/plugins/capsule/capsule.controller.test.ts
+++ b/src/services/item/plugins/capsule/capsule.controller.test.ts
@@ -542,7 +542,7 @@ describe('Capsule routes tests', () => {
     });
   });
 
-  describe('PATCH /api/items/capsules/:id/convert', () => {
+  describe('POST /api/items/capsules/:id/convert', () => {
     it('Throws if signed out', async () => {
       const {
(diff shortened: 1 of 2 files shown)


src/services/item/plugins/capsule/capsule.service.ts
```diff
@@ -0,0 +1,87 @@
...
+      ...args,
+      item: { ...args.item, type: ItemType.FOLDER, extra: { folder: { isCapsule: true } } },
+    })) as CapsuleItem;
+  }
+
+  async convertToFolder(
+    dbConnection: DBConnection,
+    member: MinimalMember,
+    itemId: UUID,
+  ): Promise<CapsuleItem> {
```
spaenleh: Shouldn't this return a `FolderItem` ?

PR #2065, opened 2026-01-29
refactor: use item type from schema

pyphilia:
Export item type schema and use this definition as type in the codebase (instead of sdk).


spaenleh (approved):
LGTM ! This was a lot of files to review 😁

PR #2071, opened 2026-02-03
fix: fix copy for h5p with custom filename on upload

### LINKED ISSUE #2070: Cannot copy h5p with invalid name
For example "é" in the name.

pyphilia:
Generates a stable and short name for saving the h5p upload. This allows to handle special characters. 

fix #2070 

TODO
- [x] add more tests


spaenleh (approved):
Looks good ! thank you for investigating this and releasing the fix !

PR #2143, opened 2026-05-29
fix: add thumbnail URL access to bookmarks

Santirv17:
Adds thumbnail URLs to bookmarked items returned by the bookmarks endpoint.
- Inject `ItemThumbnailService` into `ItemBookmarkService`.
- Pass thumbnail URLs into 'filterOutPackedItems' for 'ItemBookmarkSchemas' to have access to it.

Adds test file for thumbnail functionality
- Creates a mock bookmark repository to verify item.thumbnail exists on bookmarks with appropiate features.

### DIFF
--- a/src/services/item/plugins/itemBookmark/itemBookmark.service.spec.ts
+++ b/src/services/item/plugins/itemBookmark/itemBookmark.service.spec.ts
@@ -97,7 +97,7 @@ describe('BookmarkService', () => {
           id: itemWithoutThumbnail.id,
         },
       });
-      expect(result[1].item).not.toHaveProperty('thumbnails');
+      expect(result[1].item.thumbnails).toBeUndefined();
     });
   });
 });
--- a/src/services/item/plugins/itemBookmark/itemBookmark.service.spec.ts
+++ b/src/services/item/plugins/itemBookmark/itemBookmark.service.spec.ts
@@ -1,0 +1,103 @@
+import { v4 } from 'uuid';
+import { describe, expect, it, vi } from 'vitest';
+
+import { ItemFactory } from '../../../../../test/factories/item.factory';
+import { MemberFactory } from '../../../../../test/factories/member.factory';
+import type { DBConnection } from '../../../../drizzle/db';
+import { AuthorizedItemService } from '../../../authorizedItem.service';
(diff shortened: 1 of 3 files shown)


src/utils/config.ts
```diff
@@ -71,11 +71,7 @@ export const CORS_ORIGIN_REGEX = process.env.CORS_ORIGIN_REGEX;
 export const PUBLIC_URL = new URL(process.env.PUBLIC_URL || HOST);
 export const LIBRARY_HOST = process.env.LIBRARY_CLIENT_HOST || HOST;
 
-export const ALLOWED_ORIGINS = [
-  new URL(PUBLIC_URL).origin,
-  new URL(HOST).origin,
-  new URL(LIBRARY_HOST).origin,
-];
+export const ALLOWED_ORIGINS = [new URL(PUBLIC_URL).origin, new URL(LIBRARY_HOST).origin];
```
spaenleh: As discussed this now creates a conflict. You can safely take the version from main for this change.

src/services/item/plugins/itemBookmark/itemBookmark.service.spec.ts
spaenleh, line 1: Test looks good. 
This one mocks a lot of stuff, not necessarily a bad thing, but it means it can break easily if we change how the function works, or how it calls other functions / dependencies. 
We have jest tests (*.test.ts) that do end-to-end tests (including the DB).
They seed some data using `seedJSON` and then make a request on the server using `app.inject` (this is a HTTP request). Then they assert the JSON result. This has the benefit of involving the full path of code required to produce an API response, down to the DB level. This however is slower than mocking most of the dependencies. Both are good they are just different trade-offs.

spaenleh (commented):
Thank you for your first PR @Santirv17 !

Changes look good. On closer inspection I see that the schema did not need to be modified, as it was already supposed to include the thumbnails but they were not there ... 😅 

I left a comment about the type of test you implemented, you do not need to change the type, I just wanted to highlight the difference, and maybe take a look at the jest tests (you can ask AI to explain how they work, as they are a bit lengthy...)

Once you have resolved the conflict and the CI is green I think we will be ready to merge !

Great work !

spaenleh (approved):
Looks good ! Thank you !

PR #2147, opened 2026-06-17
feat: add optional file URL field to packed file responses

### LINKED ISSUE #2082: Add file url in `ExtendedItem`
Allows linear view to be more optimized by giving the url of the file directly (instead of n+1 calls).

Especially useful for `/children`

Santirv17:
Add file URLs to packed file item responses so clients can open file children without an additional URL request.

- file items can now expose `extra.file.url` as optional response data
- `/children` returns the URL for file children while non-file children remain unchanged

close #2082

### DIFF
--- a/src/services/item/item.service.ts
+++ b/src/services/item/item.service.ts
@@ -18,6 +18,7 @@ import {
   getParentFromPath,
 } from '@graasp/sdk';
 
+import { resolveDependency } from '../../di/utils';
 import { type DBConnection } from '../../drizzle/db';
 import {
   type ItemGeolocationRaw,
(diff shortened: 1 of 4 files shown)


src/services/item/item.service.ts
```diff
@@ -540,7 +542,7 @@ export class ItemService {
       children,
       thumbnails,
     );
-    return filteredChildren.map((children) => this.transformItemByType(children));
+    return Promise.all(filteredChildren.map((child) => this.transformItemByType(child)));
```
spaenleh: I am always wary of using `Promise.all` because it will reject at the first promise that rejects, continuing the program flow. This means that if a second promise rejects later, it can crash the server since the execution moved on. Instead I would rather use `Promise.allSettled` which ensures that all promises either resolve or reject to continue.

spaenleh (commented):
Overall changes look good, but I would take a second look at my comment just to ensure that we do not end up with a potential crash on our hand.

